PR #327, opened 2026-06-17
docs(steering) + scripts(sbir-phase3): ML methodology checklist and USAspending join scripts

hollomancer:
## Summary

This PR bundles two related artifacts surfaced in the same audit session. Title updated per Copilot's first comment to reflect actual scope.

### 1. ML methodology review checklist (the original artifact)

- Adds `docs/steering/ml-methodology-review.md` (~380 lines) — durable checklist derived from a scicode-lint audit of `packages/sbir-ml/sbir_ml/ml/`.
- Documents 30 of 66 relevant patterns (sklearn + HF inference; PyTorch patterns marked not-yet-load-bearing).
- Captures **17 manual findings + 3 net-new findings the LLM-judge surfaced**, with severity, file:line, and confirmed false positives.
- Documents a **validated local run path** using `omlx serve` + Qwen3-Coder-Next-MLX-4bit via `OPENAI_BASE_URL=http://localhost:8080/v1` (no GPU rental needed; one local patch required to scicode-lint's `DetectionResult.reasoning` `max_length`).
- Scopes the **next audit** — the rule-based transition scorer in `packages/sbir-ml/sbir_ml/transition/` where CLAUDE.md's ≥85% precision target actually lives.

**Net-new findings from the tool (not in any prior review):**
- `trainer.py:115` — `MultiLabelBinarizer.fit_transform()` before `train_test_split` (label leakage)
- `trainer.py train()` — computes test metrics but doesn't return them
- `patent_classifier.py:268-275` — feature builder `fit_transform()` runs on full dataset before any split (real leakage risk)

### 2. USAspending phase-III join scripts (bundled, not yet shipped elsewhere)

The same audit session also produced an SBIR.gov ↔ USAspending join pipeline that hadn't shipped via another PR. Rather than split into a separate PR mid-flight (which would also require carrying the per-firm digest / commercialization benchmark together since they depend on each other), the bundle is kept here:

- `scripts/data/fetch_usaspending_sbir_fy25.py` — cursor-paginated `spending_by_award` slices for FY25.
- `scripts/data/lookup_sbir_contracts_in_usaspending.py` — FPDS PIID join (NAICS + PSC enrichment).
- `scripts/data/lookup_sbir_grants_in_usaspending.py` — FABS join (CFDA enrichment; per-agency ID normalization for HHS/NIH and DoE).
- `scripts/data/build_phase3_prospect_digest.py` — per-firm Phase III prospect digest.
- `scripts/data/run_commercialization_benchmark.py` — §638(qq) / 15 USC §638(mm) commercialization benchmark using public-data proxies (FPDS + FABS for sales, Form D for investment).
- `packages/sbir-ml/sbir_ml/transition/analysis/benchmark_evaluator.py` — cast `company_id` to `str` in sort keys (3 lines).

## Copilot review — addressed in `70fc9b9f`

- ✅ **Scope/title mismatch** — PR title and description updated to reflect the bundled artifacts.
- ✅ **User-Agent strings** (4 scripts) — standardized to `SBIR-Analytics/0.1.0` matching `sbir_etl/enrichers/base_client.py`.
- ✅ **Award Amount type coercion** (2 scripts) — wrapped in `CAST(REPLACE(REPLACE(CAST("Award Amount" AS VARCHAR), '$', ''), ',', '') AS DOUBLE)` so commas/$ in SBIR.gov CSV values don't silently break SUMs.
- ✅ **Misleading dedupe comment** (`lookup_sbir_grants_in_usaspending.py`) — clarified that agency context is *dropped*, not preserved, past the dedupe step.

## Follow-ups (separate branches)

- ML test infrastructure (synthetic multi-label fixture + baseline benchmark snapshot script).
- Phase 1 PRs from the fix plan (split methodology, seeding, hand-rolled metrics, feature-builder contract).
- Transition-scorer audit per the scope in this artifact.

### DIFF
--- a/docs/steering/ml-methodology-review.md
+++ b/docs/steering/ml-methodology-review.md
@@ -1,0 +1,379 @@
+# ML Methodology Review Checklist
+
+## Overview
+
+This is the standing review checklist for ML code in `packages/sbir-ml/`. It exists to catch methodology bugs — the kind where code runs, tests pass, and reported metrics are still wrong (data leakage, missing seeds, hand-rolled metrics with magic epsilons, accuracy as a headline for imbalanced multi-label problems).
+
+The patterns are drawn from [`scicode-lint`](https://github.com/authentic-research-partners/scicode-lint)'s published 66-pattern catalog (v1.0 registry, 2026-03-15), filtered to the subset relevant to our sklearn + HuggingFace-Inference stack. PyTorch-specific patterns are documented but currently not load-bearing.
+
+Methodology bugs in the CET classifier directly distort its own reported precision/recall/F1 (computed in `trainer.py` and stored in the metrics dict). Indirectly, degraded CET classifications feed `transition/features/cet_analyzer.py` and become one signal in the rule-based **transition scorer** — which is what CLAUDE.md's "≥85% precision benchmark" actually references. The cascade is dampened by `cet_alignment`'s weight (1 of 6 signals), but real: a data-leakage bug here doesn't just produce a worse classifier, it can quietly degrade downstream transition-scoring precision in ways the CET-classifier-internal metrics don't show.
+
+### What this checklist does and does not cover
+
+**Covers:** ML methodology in `packages/sbir-ml/sbir_ml/ml/` — TF-IDF + LogisticRegression CET classifier (`cet_classifier.py`), patent classifier (`patent_classifier.py`), trainer (`trainer.py`), vectorizers (`multi_source_vectorizer.py`), HuggingFace embeddings inference (`huggingface_inference.py`).
+
+**Does NOT cover:** the rule-based **transition scorer** at `packages/sbir-ml/sbir_ml/transition/`. That's a separate system with separate methodology risks (signal weight calibration, threshold drift, ground-truth labeling protocol). CLAUDE.md's ≥85% precision target lives there, not here. See "Next audit scope" below.
+
+## When to use this checklist
+
+- **Reviewing any PR that touches `packages/sbir-ml/sbir_ml/ml/`** — required pass before merge
+- **Before re-running training that produces an artifact downstream consumers depend on** — verify the methodology hasn't drifted from this checklist
+- **Periodic baseline refresh** — re-run the checklist annually (or whenever scicode-lint releases a new pattern version) against the full ML surface area and update the "Initial Findings" appendix
+
+## How to apply
+
+For each section below, read the code being reviewed and answer the verification question. Findings are recorded against the relevant pattern ID so future reviewers can trace decisions. Pattern IDs match scicode-lint's registry (e.g., `ml-008`, `rep-001`) — useful for cross-referencing when a real tool run is later performed.
+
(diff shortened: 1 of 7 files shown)


hollomancer:
## Follow-up finding: `CETModelTrainer` is dead code (and broken)

While implementing the test infrastructure that should mechanize the Phase 1 fix plan's before/after comparison, I discovered two things that materially change how this artifact's findings should be interpreted:

**1. `CETModelTrainer` doesn't actually work end-to-end.**
`trainer.prepare_data()` returns numpy arrays via `train_test_split`, but `ApplicabilityModel.train()` at `cet_classifier.py:313` expects a pandas DataFrame (`y_train.columns`). Any attempt to invoke `CETModelTrainer.train(dataset)` raises `AttributeError: 'numpy.ndarray' object has no attribute 'columns'`. The two halves of the "training workflow" have never been wired together.

**2. Nothing in production uses it.**
Total references to `CETModelTrainer` across the repo:
- `trainer.py` itself (450 LOC)
- 2 tests (`tests/unit/ml/test_pipeline_smoke.py`, `tests/slow/ml/test_trainer.py`)
- The snapshot script I just drafted (on a sister branch — to be reworked)
- `docs/ml/cet-award-training-data.md` (docs)

The production CET asset at `packages/sbir-analytics/sbir_analytics/assets/cet/classifications.py:325` only **loads** a pre-trained model from `artifacts/models/cet_classifier_v1.pkl`. It never trains one. `artifacts/` is gitignored and no model file exists in the local working copy.

## What this means for this PR's findings

The Phase 1 fix plan that derived from this audit (split methodology, seeding, hand-rolled metrics, feature-builder contract) was correctly identifying methodology bugs — but in a file that **isn't invoked by any production code path**. Fixing them as written would not change any deployed system's behavior, because no deployed system runs `CETModelTrainer`.

The deployed `cet_classifier_v1.pkl` (if it exists in any deployed environment) was trained by an **unknown path that isn't audited here** — possibly a notebook, possibly an external script, possibly a different branch. The methodology questions are still valid; they just apply to whoever produced that artifact, not to this code.

## How to interpret this PR going forward

The methodology checklist itself (the durable artifact this PR adds) is still correct and useful — it's a pattern catalog, not a fix plan, and the patterns apply to any future ML code regardless of whether `trainer.py` lives or dies.

The "Initial Findings" section needs a banner noting that **the findings against `trainer.py` are against currently-dead code**, and the Phase 1 PRs should be re-scoped before being attempted:

- **Option A:** Delete `trainer.py`, document that training happens externally, mark `tests/slow/ml/test_trainer.py` as legacy
- **Option B:** Find the real training path (talk to the team) and audit *that* instead
- **Option C:** Fix `trainer.py` to make it actually work end-to-end, document it as the canonical training entry point, then apply the Phase 1 fixes meaningfully

This will be its own follow-up work after I check in with the team on where models actually come from. Flagging now so reviewers of this artifact don't act on findings that target dead code without that context.

🤖 Updated by Claude Code

PR #354, opened 2026-06-22
style: ruff format test_dod_fpds_substitution_test.py

hollomancer:
## Summary

`tests/unit/scripts/test_dod_fpds_substitution_test.py` landed on `main` without being `ruff format`-clean, failing the **required** Code Quality check repo-wide (`1 file would be reformatted, 598 files already formatted`).

**Formatting-only change — no logic change.** This is the same class of breakage as #346 (`test_bootstrap_form_d_leverage_ci.py`): unformatted Python reaching `main`, which the `detect-changes` config itself documents happening via PRs #340/#342/#343 through a docs-only filter gap.

Surfaced while finishing the tornado-bump PR #326 (its branch update pulled in newer `main`).

## Verification
- `ruff format --check tests/unit/scripts/test_dod_fpds_substitution_test.py` — clean after the change (was "Would reformat")
- Only this one file changed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01RMakj3aBoK2yR5bth7qeza)_

PR #355, opened 2026-06-22
fix(ci): compute docs-only correctly so mixed PRs can't skip CI

hollomancer:
## The root cause behind the recurring "unformatted Python on main"

This fixes the systemic bug that produced #346, #354, and the breakage that was masked behind the entire #326 chase.

`detect-changes` derived `docs-only` from dorny/paths-filter with `predicate-quantifier: 'every'` over an allow-list (`['**/*.md', 'docs/**']`). That quantifier **does not** mean "every changed file is a doc." Per [dorny's docs](https://github.com/dorny/paths-filter), `every` makes a *file* match only if it matches *all* patterns — but the filter's boolean output is still "**some** file matched." So `docs-only` was true whenever *any* Markdown-under-`docs/` file changed, **regardless of how many `.py` files changed alongside it.**

### Proof (PR #350's detect-changes log)
```
Received 3 items
[added] docs/research/dod-fpds-substitution-test-findings.md
[added] scripts/data/dod_fpds_substitution_test.py
[added] tests/unit/scripts/test_dod_fpds_substitution_test.py
...
Filter docs-only = true
Matching files:
  docs/research/dod-fpds-substitution-test-findings.md [added]
```
→ `docs-only=true` ⇒ every code job (Code Quality, Tests, MyPy, …) `skipped` via `docs-only != 'true'` ⇒ an unformatted test file landed on `main` ⇒ it then failed Code Quality on the next unrelated PR (#354). The prior allow-list was itself a failed attempt to fix the same class of bug (the maintainer's earlier negation-under-`every` attempt, noted in the old config comment re: #340/#342/#343).

## The fix
Compute `docs-only` **deterministically** from the full changed-file list (`dorny/paths-filter` with `list-files: json` + a small bash step), instead of relying on dorny's confusing quantifier:
- `docs-only=true` **only** when *every* changed file is Markdown (anywhere) or under `docs/`
- empty changed-file list ⇒ `false` (never skip CI vacuously)
- removes the now-unused `every-file-filters` action input and its ci.yml block

The `docs-only` output name is unchanged, so all 9 `docs-only != 'true'` gates keep working.

## Verification
Logic tested locally against representative inputs (all pass):

| changed files | docs-only |
|---|---|
| `#350`: 1 `.md` + 2 `.py` | **false** ✅ (was wrongly true) |
| pure docs (`docs/*.md`, `README.md`) | true ✅ |
| Markdown anywhere (`packages/foo/CHANGELOG.md`) | true ✅ |
| code only (`sbir_etl/foo.py`) | false ✅ |
| `docs/a.md` + `.github/workflows/ci.yml` | false ✅ |
| empty list | false ✅ |

Both files are valid YAML; the `Validate GitHub Actions Workflows` CI job will lint further.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RMakj3aBoK2yR5bth7qeza)_

### DIFF
--- a/.github/actions/detect-changes/action.yml
+++ b/.github/actions/detect-changes/action.yml
@@ -8,22 +8,14 @@ inputs:
       Evaluated with default "some" semantics: a filter is true if at least
       one changed file matches the filter.
     required: true
-  every-file-filters:
-    description: |
-      YAML string of path filters evaluated with "every file matches" semantics
-      (dorny/paths-filter `predicate-quantifier: 'every'`). Currently only the
-      `docs-only` filter is wired through to a named output; add more outputs
-      below if you need to expose additional every-file filters.
-    required: false
-    default: ''
 
 outputs:
-  # Common outputs. `docs-only` is sourced from the every_filter step (run
-  # only when `every-file-filters` is non-empty) and falls back to 'false'
-  # so callers can always compare against a stable boolean string.
+  # `docs-only` is computed explicitly from the full changed-file list (see the
+  # "Compute docs-only" step) and falls back to 'false' so callers can always
+  # compare against a stable boolean string.
   docs-only:
-    description: Whether only documentation files changed (every-file semantics)
-    value: ${{ steps.every_filter.outputs.docs-only || 'false' }}
+    description: Whether only documentation files changed (every changed file is Markdown or under docs/)
+    value: ${{ steps.docs_only.outputs.docs-only || 'false' }}
   # CI-specific outputs (some-file semantics)
   performance:
     description: Whether performance-related files changed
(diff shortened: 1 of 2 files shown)


hollomancer:
## Retrospective review

Good root-cause fix — clearly diagnosed (the `every`-quantifier semantics were genuinely confusing) and the deterministic bash reimplementation is easy to reason about. A few notes:

- Verified the `case "$f" in docs/*|*.md) ;; *) docs_only=false ;;` logic actually handles nested paths correctly — bash `case` pattern matching is plain string globbing (not filesystem pathname expansion), so `*` does match across `/`. E.g. `docs/research/data.json` matches `docs/*` as expected. So the "every file is markdown or under docs/" semantics hold for arbitrarily nested paths, not just one level. Good.
- The empty-list fallback (`docs-only=false` when zero changed files) is a sensible fail-safe against vacuous skips.
- Confirmed `every-file-filters` isn't referenced anywhere else in `.github/` after its removal from `detect-changes/action.yml` — clean removal, no dangling callers.

No issues found. The verification table in the description is a nice touch for a CI-logic change like this.

---

PR #371, opened 2026-06-26
draft(ma): discovery + verification toolkit (paused 2026-05-31)

hollomancer:
## Status: PAUSED — preserved for findability

Opened as a **draft** to surface this work in the PR queue rather than leave it stranded on a local branch. Not ready to merge.

## What this is

A four-stage M&A discovery pipeline added on `2026-05-31` (single commit `cb42b03a`):

1. `scripts/generate_ma_search_queries.py` — reads `sbir_ma_events.jsonl`, emits web-search queries for M&A candidates lacking Form D backing. Normalizes company names (strips INC/LLC/CORP/etc.) for wider recall.
2. `scripts/ma_discovery_orchestrator.py` — pluggable `SearchTool` interface; ships a `MockSearchTool` for E2E testing. Production swap targets noted: Bing / Brave / Tavily.
3. `scripts/ma_verifier.py` — keyword heuristic confirming `(company, acquirer)` pairs in snippets. Source comment notes it should be replaced by an LLM call that also extracts deal date and value.
4. `scripts/enrich_ma_with_press.py` — press-mention enrichment layer.
5. Also extends `MAEvent` with a categorical `confidence` tier (high/medium/low) alongside the existing numeric `confidence_score`.

## Why it was paused

Subsequent capital-events work landed on `main` (#286, #307, #313, #342, #343, #350) using the Form D high-confidence cohort path. That path doesn't replace this toolkit's purpose — discovery exists specifically to close the gap **where Form D doesn't help** (acquisitions of firms that never raised priced rounds and so never filed Form D). That gap remains open.

## What's needed before this can land

- Swap `MockSearchTool` for a real client (decide on Bing / Brave / Tavily and add API-key handling per repo SAM.gov / SEC EDGAR patterns)
- Replace the keyword verifier with an LLM call (extract acquirer, deal date, value)
- Integration with the existing capital-events ETL — does this feed `MAEvent` records into `capital_events.parquet`, or stand alongside as a separate detection layer?
- Test coverage (currently zero unit tests on these scripts)
- Decide where the scripts live long-term — `scripts/` root is a temporary home; a `sbir_etl/enrichers/ma_discovery/` module would match repo conventions

## Companion cleanup

Companion PR #370 ignores Claude Code / Playwright harness local state. Four stale snapshot copies of the scripts above (left at repo root after a working-tree checkout) were also deleted locally — the canonical copies live on this branch.

PR #380, opened 2026-06-27
docs(specs): M&A discovery integration design (1-pager)

hollomancer:
## Summary

Single-page design for how the web-search-based M&A discovery toolkit (scaffolded on `chore/ma-discovery-toolkit`, preserved in draft PR #371) integrates with the existing capital-events pipeline.

## Key finding

The capital-events builder at `scripts/data/capital_events/sources/ma_events.py:13-55` already reads `data/enriched_sbir_ma_events.jsonl` and filters on `confidence in {high, medium}`. The integration shape is mostly fixed — discovery just needs to flow into the existing file format. This collapses the "where does it go" question to a single live decision: how to handle collisions between discovered rows and existing Form-D-detected rows.

## Adopted

**Option C with collision rule C3**:
- Discovery fires only on Form-D-missing rows from `detect_sbir_ma_events.py`.
- Discovered rows join `sbir_ma_events.jsonl` *before* press enrichment, so both detection paths share the same downstream treatment.
- Collisions **promote** existing rows' confidence (`medium → high`, `low → medium`) rather than overwriting Form D-derived acquirer names.

## Deferred to the implementation PR

- Search backend choice (Tavily / Brave / Bing / Serper) — design lead is Tavily.
- LLM verifier model (Haiku vs Sonnet, possibly two-stage).
- Output module path (lead: `sbir_etl/enrichers/ma_discovery/`).
- Confidence threshold tuning (start at 0.75 / 0.45; revisit empirically).

## Success criteria for go/no-go

After a sample run on the actual Form-D-missing population:
- ≥ 10 net-new confirmed `medium`-or-`high` events not produced by detect_sbir_ma_events.
- ≤ 25% false-positive rate on a 20-row stratified review.
- ≤ $0.10 per row in search + LLM cost.

Below those, close draft PR #371 with lessons documented.

## Sequencing

7 follow-on PRs, ordered. Steps 1–2 (`@computed_field` confidence fix, module path move) are no-cost cleanups that can land independently of the rest.

## Reviewer prompts

- Is the C3 promotion logic the right call? Anywhere it gives the wrong answer?
- Are the per-row confidence thresholds (≥2 distinct sources for high; rumor/talks → low) calibrated correctly for SBIR-firm coverage?
- Are the success-criteria numbers (10 events, 25% FP, \$0.10/row) reasonable as a kill-switch?

### DIFF
--- a/specs/ma-discovery-integration/design.md
+++ b/specs/ma-discovery-integration/design.md
@@ -74,7 +74,9 @@ class MAEvent(BaseModel):
         return "low"
 ```
 
-**File-contract requirement.** The capital-events builder filters on the string-typed `confidence` tier (`_KEEP_CONFIDENCES = {"high", "medium"}`); it does not read a numeric score. Discovery must emit the string `confidence` tier in each row of `enriched_sbir_ma_events.jsonl` — internally derived from `confidence_score` via the `@computed_field` above, but the on-disk contract is the tier, not the score. The `≥ 0.45` threshold above is therefore an *internal* mapping; the builder-level inclusion bar is `confidence in {"high", "medium"}`. Calibrate the score → tier thresholds against existing detector output on first run.
+**File-contract requirement.** The capital-events builder filters on the string-typed `confidence` tier (`_KEEP_CONFIDENCES = {"high", "medium"}`); it does not read a numeric score. Discovery must emit the string `confidence` tier in each row of `enriched_sbir_ma_events.jsonl` — internally derived from `confidence_score` via the `@computed_field` above, but the on-disk contract is the tier, not the score. The `≥ 0.45` threshold above is therefore an *internal* starting point; the builder-level inclusion bar is `confidence in {"high", "medium"}`.
+
+**Threshold alignment requirement.** The preserved toolkit branch in PR #371 currently maps `confidence_score` to tiers at different cutoffs (`high >= 0.8`, `medium >= 0.5`) via a model validator. Do not land the confidence-field cleanup until the implementation branch either adopts the provisional `0.75 / 0.45` cutoffs in this design or explicitly replaces them with empirically calibrated thresholds. The important invariant is that every producer writes the same string tier for the same numeric score before rows reach `capital_events.parquet`.
 
 ## Triggering & cost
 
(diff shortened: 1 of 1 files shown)


hollomancer:
Addressed all 7 Copilot review comments in 9fdad9f0.

- **Line 9 (intro)** — corrected signal sources to Form D + EFTS (not 8-K).
- **Line 22 (pipeline diagram)** — marked `enrich_ma_with_press.py` as `[TODO]` / to-be-added; called out that `SyncPressWireClient` exists as a library only and no CLI produces `enriched_sbir_ma_events.jsonl` on this branch.
- **Line 32 (collision key)** — replaced `acquisition_date` with `event_date` (the actual JSONL field) and specified empty-date fallback to company-only matching.
- **Line 35 (`discovery_confirmed` placement)** — moved into `signals.discovery_confirmed`, since `scripts/data/capital_events/sources/ma_events.py` only carries `signals` / `press_wire_signals` / `signal_count` / `enriched` from the input row's metadata.
- **Line 40 (`discovered_acquirer_disagrees` placement)** — same fix; also documented the `dict[str, bool]` vs `str` trade-off and a strict-typed alternative split.
- **Line 73 (confidence_score ≥ 0.45 threshold)** — reframed as a file-contract requirement: the builder filters on the string `confidence` tier; the numeric score is an internal mapping. The score → tier thresholds calibrate against existing detector output.
- **Line 81 (idempotency key)** — replaced `(company_name, acquirer)` with `(company_name, event_date)` plus the `signals.discovery_confirmed` flag. Acquirer is null for Form D-only events so a `(name, null)` key collides across all of them.

Also tightened two adjacent consistency points: the "Why" 2nd-paragraph evidence reference, and the per-row confidence table now explains that the LLM extractor's `acquisition_date` field gets mapped to `event_date` when writing JSONL.

hollomancer:
Review note: this design does the most important thing: it collapses the integration question onto the existing `sbir_ma_events.jsonl` / capital-events path instead of inventing a parallel output. Option C plus C3 confidence promotion is a reasonable default because it preserves Form-D-derived acquirer fields while allowing independent discovery to strengthen confidence.

No blocking findings from this pass. The corrected file-contract notes (`signals.discovery_confirmed`, `event_date` collision key, numeric score internal/string tier external) are the details I would carry into the implementation PR and test explicitly.

hollomancer:
Codex review finding: the design’s confidence thresholds don’t match the preserved toolkit branch. This doc recommends `@computed_field` with high/medium cutoffs at `0.75` / `0.45`, while PR #371’s `MAEvent` helper maps high/medium at `0.8` / `0.5` via a model validator.

Since the downstream builder only sees the string tier, this threshold is load-bearing. I’d make the implementation sequence explicitly include aligning #371 to the thresholds chosen here, or state that the final thresholds remain unsettled and must be calibrated before the confidence field lands.

hollomancer:
Fixed in `7bed4be3`: clarified that the `0.75 / 0.45` score-to-tier cutoffs are provisional, called out the current #371 mismatch (`0.8 / 0.5`), and made threshold alignment an explicit requirement before the confidence-field cleanup lands.

PR #400, opened 2026-06-28
feat(bea-nipa): parquet cache + scrub remaining hardcoded tax rates

hollomancer:
## Summary

> **What changed since this PR was opened:** A 2026-06-29 audit found that the bulk of the BEA NIPA data-acquisition work the original spec proposed (BEA API client wiring, NIPA-derived federal + state/local rate computation, `FiscalTaxEstimator` integration) had **already shipped** in `sbir_etl/transformers/fiscal/nipa_rates.py`. This PR is rewritten to cover only the two genuine gaps the audit surfaced.

### 1. Parquet on-disk cache for `NIPARateProvider`

`get_rates(year)` resolution order is now: in-memory cache → on-disk parquet (API-sourced rows only) → BEA API → `_BASELINE_RATES` fallback. API fetches upsert into `data/reference/bea/nipa_tax_rates.parquet` via atomic temp-file + rename so a fresh process reuses cached rates instead of re-hitting BEA. Baseline rows aren't cached (deterministic from `_BASELINE_RATES`). Constructor renamed `cache_dir` → `cache_path` (prior `cache_dir` was unused).

### 2. Scrub three remaining hardcoded `0.22` runtime sites

| File | Before | After |
|---|---|---|
| `transformers/fiscal/sensitivity.py:110-122` | `base_rate = 0.22`, `0.18 *` for corporate | Center bands on `NIPARateProvider.get_rates(base_year).federal_income_rate` / `.federal_corporate_rate` |
| `tools/mission_c/tax_estimation.py:52` | `INDIVIDUAL_INCOME_TAX_RATE = 0.22` constant | Removed. Per-row resolution via `NIPARateProvider.get_rates(fy_int).federal_income_rate` (year-stamped per award) |
| `config/schemas/domain.py:196` | `"effective_rate": 0.22` audit-trail default | `0.194` — the 2022 NIPA baseline already in `_BASELINE_RATES[2022]`. Comment explains it's a static audit-trail snapshot; runtime resolves via `NIPARateProvider` |

## Explicitly out of scope (per the rewritten spec)

- Adding NIPA Tables 3.4 / 3.6 / 1.12 / 6.2D. The three tables `NIPARateProvider` already fetches produce all 8 derived rates downstream consumers use.
- `mission_c`'s `DEFAULT_EFFECTIVE_TAX_RATES` industry dict (SOI-sourced, not NIPA — different spec).
- The statutory `PAYROLL_TAX_RATE = 0.153` constant (FICA, no NIPA equivalent).
- Audit-trail rework. `fiscal_audit_trail.log_configuration()` continues to log the updated `TaxParameterConfig` defaults; per-analysis NIPA snapshotting is a follow-up.

## Test plan

- [x] `pytest tests/unit/transformers/ tests/unit/tools/test_mission_c.py tests/unit/config/test_schemas.py` → **444 passed**.
- [x] `pytest tests/unit/transformers/fiscal/test_nipa_rates.py -v` → **20 passed** (16 pre-existing + 4 new covering API-fetch persistence, second-provider cache hit, baseline-not-cached invariant, upsert vs. distinct-key writes).
- [x] `pytest tests/unit/transformers/test_fiscal_parameter_sweep.py tests/unit/transformers/fiscal/test_sensitivity.py` → 44 passed (sensitivity bands unchanged in width; only center shifted).
- [x] `pytest tests/unit/tools/test_mission_c.py` → 13 passed (asserts NIPA-baseline range instead of removed constant).
- [x] `ruff check` + `ruff format --check` clean on all 8 changed files.

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/tools/mission_c/tax_estimation.py
+++ b/packages/sbir-analytics/sbir_analytics/tools/mission_c/tax_estimation.py
@@ -172,7 +172,9 @@ class TaxEstimationTool(BaseTool):
 
                 # Tax estimates — federal income rate is year-stamped via NIPA;
                 # FICA payroll rate is statutory and stays a module constant.
-                fy_int = int(fy) if fy is not None else assessment_year
+                # `fy` comes from a pandas row, so missing values are typically
+                # NaN (not None); `int(np.nan)` would raise ValueError.
+                fy_int = int(fy) if fy is not None and pd.notna(fy) else assessment_year
                 individual_income_rate = rate_provider.get_rates(fy_int).federal_income_rate
                 individual_income_tax = wages * individual_income_rate
                 payroll_tax = wages * PAYROLL_TAX_RATE
--- a/packages/sbir-analytics/sbir_analytics/tools/mission_c/tax_estimation.py
+++ b/packages/sbir-analytics/sbir_analytics/tools/mission_c/tax_estimation.py
@@ -24,6 +24,8 @@ from typing import Any
 
 import pandas as pd
 
+from sbir_etl.transformers.fiscal.nipa_rates import NIPARateProvider
+
 from ..base import BaseTool, DataSourceRef, ToolMetadata, ToolResult
 
 
@@ -46,10 +48,10 @@ DEFAULT_EFFECTIVE_TAX_RATES = {
     "Other": 0.18,
 }
 
-# Payroll tax rates (FICA)
+# Payroll tax rates (FICA): statutory rate, no NIPA equivalent.
 PAYROLL_TAX_RATE = 0.153  # Combined employer + employee (Social Security + Medicare)
-# Approximate individual income tax effective rate for SBIR-relevant income levels
-INDIVIDUAL_INCOME_TAX_RATE = 0.22
+# Individual income effective tax rate is resolved per-row via NIPARateProvider
+# (BEA NIPA Tables 3.2/3.3/1.5); see TaxEstimationTool.execute.
 
 
 class TaxEstimationTool(BaseTool):
@@ -71,6 +73,7 @@ class TaxEstimationTool(BaseTool):
         fpds_revenue: pd.DataFrame | None = None,
         commercialization_scores: pd.DataFrame | None = None,
         effective_tax_rates: dict[str, float] | None = None,
+        nipa_rate_provider: NIPARateProvider | None = None,
         discount_rate: float = 0.03,
         assessment_year: int = 2026,
         **kwargs: Any,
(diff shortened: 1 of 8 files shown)


hollomancer:
Addressed all five Copilot review comments in a491cb1:

- **requirements.md (line 80)**: Updated acceptance criterion 1 to clarify that only API-sourced rates are persisted; baselines are deterministic from `_BASELINE_RATES` and intentionally not written. This matches the implemented provider and existing tests.
- **nipa_rates.py `_read_parquet_cache` (line 250)**: Added schema validation — after `pd.read_parquet`, verify `required_cols = {"year", "source", *_PARQUET_RATE_COLUMNS}` is a subset of `df.columns`; on mismatch, log a warning and return `None` so the provider falls back to fetch/baseline.
- **nipa_rates.py `_write_parquet_cache` (line 305)**: Replaced the fixed `.tmp` suffix with `tempfile.mkstemp(dir=cache_path.parent, suffix=".parquet.tmp")` + `os.replace`, so concurrent refresh processes no longer race on the same temp path. The temp file is `unlink(missing_ok=True)`'d on exception to avoid leaking partial writes.
- **tax_estimation.py (line 176)**: `fy` is a pandas row value where missing reads as `NaN`, not `None`. Guarded with `pd.notna(fy)` so `int(np.nan)` can't raise `ValueError`; matches the existing `assessment_year` fallback semantics in this block.
- **test_mission_c.py (line 189)**: `test_default_tax_rates` now takes `tmp_path` and passes an explicit `cache_path=tmp_path / "nipa_cache.parquet"` to `NIPARateProvider`, making the test hermetic against an existing local `data/reference/...` cache.

Verification:
- `uv run pytest tests/unit/transformers/fiscal/ tests/unit/tools/` — 143 passed
- `uv run ruff check sbir_etl tests packages/sbir-analytics/sbir_analytics/tools/mission_c` — All checks passed
- `uv run ruff format --check ...` — 474 files already formatted
- `uv run python -m mypy sbir_etl/transformers/fiscal/nipa_rates.py packages/sbir-analytics/sbir_analytics/tools/mission_c/tax_estimation.py` — Success: no issues found in 2 source files

hollomancer:
Review note: the narrowed BEA NIPA scope looks sound. The persistent parquet cache is defensive enough for normal use after the follow-up fix: schema validation on read, unique temp file plus `os.replace`, and no baseline pollution. I also checked for current `NIPARateProvider(cache_dir=...)` call sites and did not find breakage from the `cache_path` rename.

No blocking findings from this pass. CI is green. Residual risk is mostly operational: the default cache path writes under `data/reference/bea/`, so any scheduled/materialized environment should either provision that path or pass an explicit cache path if repo-local writes are undesirable.

hollomancer:
Codex review: I didn’t find a blocking issue in this diff. The `NIPARateProvider` cache order and the Mission C/sensitivity rate plumbing look internally consistent, and I don’t see existing repo callers passing the old `cache_dir=` argument into `NIPARateProvider`.

One small compatibility note before merge: since this PR renames the constructor argument to `cache_path`, it is worth either documenting that as a deliberate break or accepting `cache_dir` as a deprecated alias for one release if any notebooks/ad hoc scripts outside the repo may instantiate it directly.

PR #324, opened 2026-06-12
feat(analytics): add leverage-ratio validation, sensitivity, and gating

hollomancer:
### Motivation

- Supply an auditable, reproducible leverage-ratio computation separate from its checks so a single defect cannot make both result and validation agree.  
- Make methodological choices (match threshold, window, dollars, negatives, STTR) explicit and report sensitivity ranges rather than a single preferred configuration.  
- Protect publication and future refreshes with automated quality gates and a reproducible manual-review queue that captures evidence and reviewer fields.

### Description

- Add a compact primary calculator `LeverageConfig`, `prepare_obligations`, and `compute_leverage_ratio` in `packages/sbir-analytics/sbir_analytics/tools/mission_b/leverage_ratio.py` that produces `company`, `agency`, `cohort`, and `headline` aggregates and a stable `scenario_id`.  
- Add an independent validation/sensitivity/gating module `packages/sbir-analytics/sbir_analytics/tools/mission_b/leverage_ratio_validation.py` implementing invariants and gates including `source_reconciliation`, `no_sbir_in_non_sbir_numerator`, `no_duplicate_obligations_after_matching`, `deobligations_handled_as_configured`, `stable_aggregation_across_dimensions`, `match_quality_coverage`, `required_output_dimensions`, and `no_unexplained_invalid_ratios`, plus `enforce_report_gate()` to fail refreshes when thresholds are missed.  
- Implement full-factorial sensitivity runs (`run_sensitivity_analysis`) across entity-match thresholds, analysis windows, nominal vs inflation-adjusted dollars, inclusion/exclusion of negative obligations, and STTR treatment, and add `sensitivity_summary` to expose headline ranges.  
- Add deterministic manual-review sampling (`build_stratified_review_sample`) with reviewer/evidence fields (`review_status`, `reviewer`, `review_notes`, `entity_match_confirmed`, `sbir_classification_confirmed`, `review_evidence_reference`) that reference source obligation IDs for reproducible human inspection.  
- Document gates, sensitivity expectations, manual-review workflow, and known limitations in `docs/leverage-ratio-validation.md` and update the leverage-ratio spec status in `specs/leverage-ratio-analysis/requirements.md`.  
- Add unit tests `tests/unit/tools/mission_b/test_leverage_ratio_validation.py` that cover reconciliation, de-obligation handling, duplicate/missing-dimension failures, sensitivity matrix behavior, and review-sample evidence generation.

### Testing

[...]

PR #323, opened 2026-06-12
Add follow-on funding multiplier analysis asset, pure calculator, fixtures, tests, and docs

hollomancer:
### Motivation

- Implement the Leverage Ratio specification to compute non-SBIR-to-SBIR follow-on funding ratios (DOD primary, DOE civilian comparison) and produce a reproducible reconciliation vs. the NASEM 4:1 benchmark.  
- Provide an explicit, auditable calculation surface (pure function) and deterministic fixture so results are traceable and testable across cohort, agency, company, and fiscal-year dimensions.  
- Surface match-quality and methodological decisions (entity-match thresholds, STTR handling, negative obligations, nominal vs adjusted dollars) so differences from published benchmarks are classified as methodological rather than implementation errors.

### Description

- Added a pure, policy-driven calculator at `packages/sbir-analytics/sbir_analytics/assets/leverage_ratio/analysis.py` exposing `calculate_leverage_ratios`, `LeverageRatioPolicy`, and `LeverageRatioResult` and returning traceable `company`, `agency`, `cohort`, `fiscal_year`, and `quality` tables.  
- Added production-shaped adapters in `integration.py` (`build_canonical_obligations`) to join SBIR awards, entity-resolution outputs, and USAspending transactions into the canonical transaction-level input, including match-score normalization and cohort/profile derivation.  
- Added reconciliation utilities in `reconcile.py` to compare pipeline DOD results to the NASEM benchmark and generate a human-readable markdown summary; added a Dagster asset `leverage_ratio_analysis` in `asset.py` that writes Parquet tables and JSON/Markdown reconciliation outputs to `data/processed/leverage_ratio/`.  
- Added deterministic fixtures and tests: CSV fixtures in `tests/fixtures/leverage_ratio/`, unit tests in `tests/unit/leverage_ratio/test_analysis.py`, and an integration test in `tests/integration/leverage_ratio/test_pipeline.py`; and added user-facing docs `docs/leverage-ratio-analysis.md` and `docs/leverage-ratio-reproducibility.md` documenting semantics, execution, outputs, and unresolved product decisions.

### Testing

- Ran lint/format checks with `ruff` against the new modules and they passed.  
- Ran unit tests: `PYTHONPATH=packages/sbir-analytics uv run pytest -o addopts='' tests/unit/leverage_ratio -q` — all unit tests passed (`5 passed`).  
[...]

PR #414, opened 2026-06-30
docs(specs): follow-on funding multiplier validation/sensitivity/review-sampling design (supersedes #324)

hollomancer:
## Summary

Spec doc capturing the design intent from PR #324 (closed as parallel-implementation fork) so the validation, sensitivity, and review-sampling features can be cleanly re-implemented on PR #323's foundation when prioritized.

## Why a spec instead of code

PR #324 was a parallel implementation of the leverage-ratio analysis at a different module path (\`tools/mission_b/\`) with a different data schema (\`canonical_company_id\` vs PR #323's \`company_id\`, \`analysis_amount\` + \`is_sbir\` flag vs \`sbir_amount\`/\`non_sbir_amount\` columns). Schema-aligning #324 onto #323's foundation would have been a 2-3 hour rewrite — for code that the Copilot review had identified as having substantive bugs (silent NaN coercion, tautological validation, IndexError on empty scenarios, type-mixing in evidence sorting).

Closing #324 and capturing the design intent here preserves the genuinely-valuable parts (the validation invariants list, the sensitivity factorial structure, the review-sample evidence fields) without locking in the buggy implementation.

## What this spec covers

Three new modules to add under \`packages/sbir-analytics/sbir_analytics/assets/leverage_ratio/\`:

- **validation.py** — 8 independent invariants with tolerance handling and a report gate. Lessons from #324's Copilot review baked in (e.g., source_reconciliation computes totals independently from raw is_sbir/is_sttr columns, not from the eligibility flags the analysis uses — so it's not tautological).

- **sensitivity.py** — Full-factorial sweep across 5 methodology parameters (72 scenarios). Handles empty-scenario case explicitly so an over-restrictive filter combo doesn't crash the run.

- **review_sampling.py** — Stratified deterministic sample with the evidence fields a human reviewer needs. Robust to mixed-type obligation_ids.

Plus a report-gate function (fails the Dagster asset run when invariants fail or sensitivity range exceeds 2× threshold).

## Effort

~16-20 hours / 2-3 days for thorough implementation including test coverage.

## Sequencing

Land PR #323 first (the core analysis asset). Then implement this spec on top of #323's foundation as a fresh PR.

### DIFF
--- a/specs/follow-on-multiplier-validation/design.md
+++ b/specs/follow-on-multiplier-validation/design.md
@@ -1,34 +1,36 @@
-# Leverage-Ratio Validation, Sensitivity & Review-Sampling — Design
+# Follow-on Funding Multiplier Validation, Sensitivity & Review-Sampling — Design
 
 **Status:** Spec.
 **Date:** 2026-06-30.
-**Builds on:** [PR #323](https://github.com/hollomancer/sbir-analytics/pull/323) (`packages/sbir-analytics/sbir_analytics/assets/leverage_ratio/`).
+**Builds on:** [PR #323](https://github.com/hollomancer/sbir-analytics/pull/323) (`packages/sbir-analytics/sbir_analytics/assets/follow_on_multiplier/`).
 **Supersedes:** [PR #324](https://github.com/hollomancer/sbir-analytics/pull/324), closed as a parallel-implementation fork (different schema, different module path).
+
+> The follow-on funding multiplier is the non-SBIR federal obligations per dollar of SBIR/STTR investment for SBIR-recipient firms. NASEM's reviews of DoD SBIR call this quantity the *leverage ratio*; this codebase uses *follow-on funding multiplier* for the same calculation.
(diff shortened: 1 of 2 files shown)


hollomancer:
Pushed commit `cb2212da` addressing all six Copilot points:

- `ValidationResult` field count: three → six.
- `deobligations_handled_as_configured`: dropped #324-specific `analysis_amount` reference; now uses #323's canonical `obligation_amount` semantics.
- `required_output_dimensions`: softened from "and no others" to "core columns present" since per-aggregate dimension columns are intentional.
- Review-sampling + testing-approach: aligned to #323's canonical schema (`company_id`, `match_confidence`) — dropped #324's `canonical_company_id`/`entity_match_confidence`.
- Report-gate sensitivity-span: explicit NaN/inf/non-positive filtering before `max / min`, plus `SensitivityGateInsufficientData` error when too few scenarios survive.

PR #429, opened 2026-07-12
chore: remove orphaned files and dead config entries

hollomancer:
- MANIFEST.in: inert under hatchling build backend, references
  nonexistent src/ layout
- lifecycle.json: orphaned S3 policy; real rules live in
  infrastructure/cdk/stacks/foundation.py
- config/transition/presets.yaml: never loaded by any code path
- notebooks/getting_started_executed.ipynb: committed CI-generated
  artifact, now gitignored
- definitions_core.py: superseded, zero references anywhere
- pyproject.toml testpaths: drop nonexistent sbir-analytics/tests path


hollomancer (commented):
The runtime import repairs and dead-code removals look consistent, but this cleanup leaves two live documentation references pointing at files deleted here:

- `docs/research-plan-alignment.md:38,151` still describes `packages/sbir-ml/sbir_ml/transition/performance/contract_analytics.py` as a built/current analytics component.
- `docs/steering/ml-methodology-review.md:176` still includes the deleted `packages/sbir-ml/sbir_ml/ml/huggingface_inference.py` in a command.

Please update or annotate these references so the cleanup does not leave broken/current-looking guidance behind.

Issue #441, opened 2026-07-15
Epic: modular analysis and report platform


hollomancer:
## Objective

Create one reusable, versioned analysis-run platform for technology censuses, technology-area transition reports, and recurring analytical reports.

## Proven inputs

- PR #439 (merged): config-driven technology census, evidence, provenance, immutable snapshots, and comparison safety.
- PR #418: fail-loud input validation and `not_evaluable` semantics.
- PR #425: award-key uniqueness, transaction-vs-award grain, and required-column findings.
- Existing weekly-report refactor: thin CLI over importable orchestration and rendering modules.

## Foundation sequence

1. Canonical `AwardCorpus`, `ReportingWindow`, stable award identity, and source manifest.
2. Versioned `AnalysisSpec` / `AnalysisRun` contracts and profile registry.
3. Importable technology cohort service with distinct classifier strategies for census and transition analyses.
4. Typed evidence-channel stages with explicit `computed`, `unavailable`, and `not_applicable` states.
5. Versioned metric/claim manifest consumed by snapshots, verification, and renderers.
6. Thin CLI, Dagster, tool, snapshot, and HTTP adapters.

## Guardrails

- Preserve PR #439 drone-manufacturing and UAS-relevance calibration.
- Preserve nanotechnology, quantum, and hypersonics parity fixtures.
- Do not create one universal matching methodology.
- Do not add one API enum entry per profile.
- Keep expensive execution asynchronous; keep snapshot reads deterministic and auditable.
- Every profile declares taxonomy and methodology versions.

## Exit criteria

Adding a configured technology profile does not require bespoke CLI, Dagster, snapshot-directory, or API-route code, and existing calibrated outputs remain unchanged unless an explicitly versioned methodology change is made.

hollomancer:
## Consolidation checkpoint

The prerequisite stack is now merged into `main`:

- #439 → `7ca0eab3`
- #440 → `bb735176`
- #404 → `1a852176`
- #431 → `2b64eac6`
- #418 → `68888202`

Baseline `main` is `68888202125b5ed6d4234be2b041f00ab50bacb0`. The consolidated drone-census + tech-area/nanotechnology regression set passed **95 tests**, and `agent/modular-analysis-platform` was created at that baseline.

Stale donors #423 and #425 are closed without merge; their branches remain for provenance. Their useful work is now split into:

1. #449 — award identity and dataset-grain contract (draft, based on `main`, full CI green)
[...]

PR #425, opened 2026-07-09
Product 1 join-key audit: PIID uniqueness, grain, IDV inheritance (DO NOT MERGE — review)

hollomancer:
## Product 1 PIID join-key audit — fixes behind review, **do not merge**

Audits Product 1 (`phase_iii_retrospective_candidates`) for the bug class the benchmark surfaced: **silent joins/identities on non-unique FPDS keys**. Full report: `specs/product-1-status-denial-flags/audit-piid-grain.md`.

### Go/no-go
**Do not freeze the validation sampling frame yet — but reassuringly, Product 1 currently emits nothing** (contract source `contracts_ingestion.parquet` unmaterialized; no `phase_iii_candidates.parquet`). No annual-report number is at risk today; this audit is **preventive**. Freeze only after M0a lands a keyed, `research`-carrying contract source and the fixed pipeline is re-run.

### Bugs found + fixed
1. **Bare-PIID `target_id` → `_candidate_id` collisions.** FPDS order PIIDs recur — the 600-record SR3 pull is **all PIID "0001" across 286 firms**. Distinct awards collapsed to one candidate identity. Fixed: `award_key_series` (unique key, else compound; raises on bare-PIID-only).
2. **Wrong-grain coded status.** `_is_phase_iii_already_coded` ran per row; coded awards' non-coded mods survived as false flags. Fixed: award-grain, **coded if ANY transaction carries SR3/ST3**; targets collapsed to award grain.
3. **Silent no-op exclusion** when the `research` column is absent (USAspending data). Fixed: raise loudly.

### Guards
- Loud guards in `_prepare_contracts`; centralized `award_key_series`.
- Regression test `tests/unit/phase_iii_candidates/test_award_key_grain.py` (two "0001" orders under different parent IDVs; fails on bare PIID, passes on compound key).
- Gotcha recorded in `docs/steering/data-quality.md`.

### Empirical (Phase B)
- `award_data.csv`: 540,343 lines vs **219,501 records** — fully explained by embedded newlines (award-grain, not mods).
- SBIR.gov `Contract` (cross-system join key): 9,204 values shared by >1 award ("NAS 96-1" on 348); ATN shared on 47k values. SBIR↔FPDS bare-field joins are unsafe.

### Repo-wide (report-only, NOT fixed here)
[...]

PR #449, opened 2026-07-16
fix(data): enforce award identity and dataset-grain contracts

### LINKED ISSUE #447: Foundation: award identity and dataset-grain contracts
Parent: #441
Donor PR: #425

Extract the durable correctness findings from the Product 1 PIID audit into shared dataset contracts.

Acceptance criteria:
- stable award identity never falls back silently to a bare, non-unique PIID
- compound-key requirements are explicit and validated
- null/partial key columns cannot masquerade as valid identities
- transaction inputs collapse to award grain under a documented policy
- coded status is aggregated at award grain
- required coding columns fail loudly
- regression fixtures cover repeated order PIIDs across parent IDVs
- data-quality guidance documents permitted keys and grains

This issue owns reusable correctness primitives, not Phase III research scripts.

hollomancer:
Closes #447
Supersedes #425
Parent epic: #441

## What this isolates

- adds one shared, strict contract-award identity utility
- rejects partial or conflicting award keys instead of silently using bare PIIDs
- requires agency + actual parent IDV + PIID when no complete precomputed key exists
- collapses FPDS/USAspending modifications explicitly to award grain
- evaluates Phase III coded status across every transaction before candidate filtering
- documents the transaction-grain and financial-aggregation boundary
- updates the Phase III integration fixture to exercise an explicit award-grade key

This salvages the useful identity work from orphan commit `4c000729` on the retained #425 donor branch, then tightens it so every row must carry a complete identity and parent-agency fields cannot stand in for parent-IDV identifiers. It is rebased directly on the consolidated `main` and deliberately excludes the Phase III research scripts and product-spec work.

## Verification

- focused award identity, Phase III grain, and affected integration regressions: **13 passed**
- Ruff: **all checks passed**
- mypy on the changed utility and pairing module: **success**
- GitHub Actions: **fully green**, including integration, fast/slow unit suites, package mypy, static analysis, performance, and S3 integration

The donor branch is retained for provenance until this successor is reviewed.

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/assets/phase_iii_candidates/pairing.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/phase_iii_candidates/pairing.py
@@ -4,12 +4,20 @@ from __future__ import annotations
 
 import pandas as pd
 
+from sbir_etl.utils.award_identity import (
+    award_key_series,
+    collapse_transactions_to_award_grain,
+)
+
 # FPDS Element 10Q codes that mark a contract as already-coded Phase III.
 # Duplicated intentionally — avoids cross-package import for a five-element set.
 _PHASE_III_RESEARCH_CODES = frozenset({"SR3", "ST3"})
 
 # Tokens we accept as evidence that ``sbir_phase`` already says "Phase III".
 _PHASE_III_LABELS = frozenset({"PHASE III", "III", "3", "PHASE 3"})
+
+# If neither column exists, the already-coded exclusion cannot run safely.
+_CODED_STATUS_COLUMNS: tuple[str, ...] = ("research", "sbir_phase")
 
 
 PAIR_S1_COLUMNS: list[str] = [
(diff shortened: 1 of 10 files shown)


packages/sbir-analytics/sbir_analytics/assets/phase_iii_candidates/pairing.py
```diff
@@ -124,15 +132,27 @@ def _prepare_contracts(contracts: pd.DataFrame) -> pd.DataFrame:
 
     df = contracts.copy()
 
-    # Compute "already coded" mask before column projection so we can read
-    # whichever raw column the input frame carries.
-    coded_mask = (
+    if not any(column in df.columns for column in _CODED_STATUS_COLUMNS):
```
hollomancer, line 135: **Blocking — this makes the default asset incompatible with its canonical producer.** `_default_retrospective_loader` reads `contracts_ingestion.parquet`, which `ContractExtractor._collect_and_write` creates from `FederalContract.model_dump()`. That real schema has `contract_id`, `agency`, `parent_contract_id`, etc., but no top-level `research` or `sbir_phase`, so every non-empty default run now raises here. If a coding field were added, the next call still cannot construct an identity: the producer exposes a bare `contract_id`, does not expose a supported agency-code alias, and leaves column 1's canonical generated award ID nested as `metadata["award_id"]`. The integration fixture masks both failures by injecting `research` and `contract_award_unique_key`. Please propagate the Element 10Q field and canonical generated ID through the producer (or add an explicit validated adapter at the loader boundary), then add an integration regression using actual `FederalContract.model_dump()` output.

sbir_etl/utils/award_identity.py
```diff
@@ -0,0 +1,165 @@
...
+    if df.empty:
+        return pd.Series(index=df.index, dtype="object", name="_award_key")
+
+    for column in UNIQUE_AWARD_KEY_COLUMNS:
+        if column not in df.columns:
+            continue
+        values = _normalized_series(df, column)
+        nonempty = values.ne("")
+        if nonempty.all():
+            return values.rename("_award_key")
```
hollomancer: **Blocking identity ambiguity:** this returns as soon as the first complete precomputed alias is found, so later aliases are never compared. For example, `contract_award_unique_key="A"` plus `generated_unique_award_id="B"` silently groups the row under `A`, even though the PR body and steering guidance say conflicting aliases fail loudly. Please normalize/coalesce all present precomputed aliases row-wise, reject conflicts (while retaining the intended partial-key policy), and add a regression with two complete but different aliases.

hollomancer (commented):
The award-grain direction is sound, but these two issues are blocking for the stated identity contract: the default retrospective asset can no longer consume its canonical producer output, and conflicting precomputed aliases are still accepted silently. CI is green because the integration fixture now injects fields the real producer does not emit.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_candidates/pairing.py
hollomancer: Fixed in `1ef965ca`. The contract model now carries the extractor's generated unique award ID plus optional research metadata, the default retrospective loader adapts legacy metadata into the production schema, and an extractor→parquet→default-loader regression test covers the real path. Focused and broader model/extractor suites are green.

sbir_etl/utils/award_identity.py
hollomancer: Fixed in `1ef965ca`. Award-key normalization now compares every complete precomputed alias and raises on disagreement instead of silently choosing the first alias. Added a conflicting-alias regression test.

PR #450, opened 2026-07-16
Add monthly procurement transition reporting

hollomancer:
## Summary

Adds a monthly public-data procurement transition reporting pipeline for SBIR awards and SAM.gov opportunities.

- adds a SAM.gov Opportunities model, extractor, and download script
- adds directed and competitive follow-on opportunity pairing for Phase III candidate surfacing
- adds monthly cohort enrichment, candidate generation, and report rendering scripts
- adds a scheduled GitHub Actions workflow for monthly artifact generation
- documents public-data-only assumptions, monthly cadence limits, and the hand-audit gate before representative distribution

## Impact

Procurement center representatives can receive monthly office-routed packets showing recently awarded, completed, changed, or end-date-near SBIR records and related public opportunity candidates. HIGH candidate sections remain gated by the audit sample and the 85 percent observed precision requirement before production distribution.

## Validation

- Ruff passed on staged Python files
- Workflow YAML parsed successfully
- Staged whitespace check passed
- Focused pytest suite passed: 68 passed

## Notes

Live monthly execution requires SAM_GOV_API_KEY. Optional AI summaries require OPENAI_API_KEY; without it, the report falls back to deterministic public-field text.


hollomancer:
Consolidation/stacking note: this concurrent draft overlaps `phase_iii_candidates/pairing.py` with the award-identity foundation in #449.

Please rebase (or temporarily stack) this product-layer PR on #449 before review, and consume the shared award key / transaction-to-award-grain contract rather than restoring bare `contract_id` identity. Draft #451 is the separate research benchmark and should remain research-only; #450 can stay focused on the monthly SAM.gov/reporting product.

I have not rewritten this branch because it appeared during the consolidation and may have active work. Current CI also reports code-quality and `sbir-analytics` mypy failures that should be addressed after the rebase.

.github/workflows/monthly-procurement-transition.yml
```diff
@@ -0,0 +1,122 @@
...
+        id: period
+        env:
+          REQUESTED_MONTH: ${{ inputs.month }}
+        run: |
+          MONTH=$(python -c "import datetime,os; requested=os.environ.get('REQUESTED_MONTH','').strip(); today=datetime.date.today(); first=today.replace(day=1); previous=(first-datetime.timedelta(days=1)).strftime('%Y-%m'); print(requested or previous)")
+          echo "month=$MONTH" >> "$GITHUB_OUTPUT"
+
+      - name: Restore prior public-award snapshot
+        uses: actions/cache@v5
+        with:
```
hollomancer, line 45: Minor: the rest of the repo pins `actions/cache@v6` (see the other two workflows using it), but this uses `@v5`. Suggest bumping to `@v6` for consistency. `checkout@v7` / `upload-artifact@v7` here already match the repo convention.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,298 @@
...
+        if master.empty:
+            for column in ("signal_class", "confidence_bucket", "center_name", "center_code"):
+                master[column] = pd.Series(dtype="object")
+            return master
+        if "prior_award_id" in master and "award_id" in awards:
+            master = master.merge(awards, left_on="prior_award_id", right_on="award_id", how="left")
+        if "target_id" in master and "notice_id" in opportunities:
+            master = master.merge(
+                opportunities, left_on="target_id", right_on="notice_id", how="left", suffixes=("", "_opp")
+            )
```
hollomancer, line 317: Merge-suffix shadowing worth a look: `normalize_awards` already emits a `source_url` column, so after this left-merge with `suffixes=("", "_opp")` the **opportunity's** `source_url` becomes `source_url_opp`, while the award's (usually null for the public SBIR CSV) keeps the plain `source_url` name. Both `_packet()` and `ai.py`'s evidence dict then read `row.get("source_url")` expecting the SAM link and only reach the correct URL via the `ui_url` fallback — the explicitly-built `Opportunity.source_url` is never surfaced. Same collision applies to `agency`/`office` (handled via `agency_opp`/`target_*` in `_pick`, so those are fine). Consider dropping/renaming the award `source_url` before the join, or reading `source_url_opp` explicitly.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,298 @@
...
+        master["confidence_bucket"] = master.get("is_high_confidence", False).map(
+            lambda value: "HIGH" if bool(value) else "WATCHLIST"
```
hollomancer: `master.get("is_high_confidence", False)` returns the scalar `False` (not a Series) when the column is absent, so `.map(...)` would raise `AttributeError: 'bool' object has no attribute 'map'`. In practice `score_candidate_pairs` always writes the column so this only bites if a hand-supplied/older `phase_iii_candidates.parquet` is passed to `--candidates`. Cheap to harden: `master.get("is_high_confidence", pd.Series(False, index=master.index))`.

---

scripts/data/enrich_procurement_awards.py
```diff
@@ -0,0 +1,81 @@
...
+    for _, row in companies.iterrows():
+        company = str(row.get("company") or "").strip()
+        if not company:
+            continue
+        summaries[company] = lookup_company_federal_awards(
+            company, str(row.get("uei") or "").strip() or None
+        )
+    cohort["naics_code"] = cohort.apply(
+        lambda row: (
+            (summaries.get(str(row.get("company"))) or object()).naics_codes[0]
```
hollomancer: This `apply` looks up `summaries.get(str(row.get("company")))` three times per row and uses `or object()` as a dead-branch guard (the ternary already short-circuits, so `object().naics_codes` is never evaluated). Per the repo's "Simplicity First" standard, this reads as overcomplicated. A precomputed `company -> naics0` dict + `cohort["company"].map(...)` (matching the `federal_award_count` / `possible_followon_count` style just below) would be clearer and drop the `object()` sentinel entirely.

---

hollomancer (commented):
## Review — Monthly procurement transition reporting

Reviewed the full diff (20 files, +1,594 / −49). Overall this is well-structured, clearly-scoped work with genuinely good instincts around public-data provenance, no-silent-truncation, and keeping the AI layer non-authoritative. Comments below are mostly polish + a couple of latent robustness issues; nothing blocking. Leaving as **Comment** since the PR is still a draft and CI hasn't reported yet.

### What it does
- New `Opportunity`/`OpportunityContact` Pydantic models + `SamGovOpportunitiesExtractor` (bounded pagination, retry, explicit "refuse partial output" caps).
- `pair_filter_s2` (directed: notice-type gate + UEI match, agency+lineage fallback) and `pair_filter_s3` (competitive follow-on: NAICS/PSC overlap + Jaccard ≥ 0.10) with new `WEIGHTS_DIRECTED`/`WEIGHTS_FOLLOWON` and thresholds; refactors the per-row scoring loop out of the Dagster asset into a reusable `score_candidate_pairs`.
- Deterministic `MonthlyReportBuilder` producing per-center markdown/HTML packets, `master_candidates.csv`, `audit_sample.csv`, `evidence.ndjson`, and a `manifest.json`; optional evidence-cited AI summaries that are discarded unless every sentence carries a `[SAM]/[SBIR]/[USASPENDING]` citation.
- Four CLI scripts + a scheduled monthly GitHub Actions workflow, plus docs stating the public-data-only assumptions and the ≥85% hand-audit gate before distribution.

### Strengths
- The `score_candidate_pairs` extraction is a clean refactor that de-duplicates the scoring loop and makes the scripts testable without Dagster.
- Good defensive posture in the extractor: record/page caps raise rather than silently truncating, and this is covered by a test (`test_page_cap_raises_instead_of_returning_partial_data`).
- The AI boundary is well-designed — `validate_cited_summary` enforces citations, `temperature=0.0`, and the summarizer can't influence scores. Nicely tested.
- Consistent with project conventions: `datetime.UTC`, no `from __future__ import annotations` added to the Dagster asset file, module-level `_validate_weights` fail-fast guards extended to the new weight sets.

### Issues & suggestions
See inline comments for specifics. Summary:
1. **`source_url` merge-suffix shadowing** (`core.py::_master`) — award `source_url` shadows the opportunity's, so packets/AI reach the SAM link only via the `ui_url` fallback. Worth fixing so the intended `Opportunity.source_url` surfaces. *(inline)*
2. **`master.get("is_high_confidence", False).map(...)`** raises if the column is ever absent (scalar `False` has no `.map`). Cheap to harden. *(inline)*
3. **`enrich_procurement_awards.py` naics `apply`** is overcomplicated (`or object()` dead branch + triple dict lookup); a precomputed map would match the surrounding style. *(inline)*
4. **`actions/cache@v5`** vs the repo's `@v6`. *(inline)*
5. **`_coerce_competition_type` overload** (assets.py): the new single-letter branches (`U/S/P/O/K/R`) are SAM.gov *notice-type* codes, but this function is shared with the RETROSPECTIVE/FPDS path where the same field carries *extent-competed* codes. It happens to be safe today (FPDS extent codes are A–G/CDO, no overlap, and `target_competition_type` is populated with the notice code for opportunities), but it silently conflates two code systems in one function. A short comment or a signal-class-aware split would prevent a future collision.

### Test coverage
Solid unit coverage for the extractor, both pair filters (happy paths), cohorting, packet/manifest emission, and AI validation. Gaps to note (some already tracked as open spec tasks 3.7/3.8/4.3):
- No test exercises the **no-UEI lineage fallback** in `pair_filter_s2` or the **agency fallback** in `pair_filter_s3` — these are the trickiest branches and are currently unverified.
- No test wires the `summarizer` callback through `MonthlyReportBuilder` (only `validate_cited_summary` in isolation).
- The ≥85% precision gate is a documented manual hand-audit, not enforced in this workflow — consistent with the docs, just flagging that HIGH sections ship ungated by CI.

### Risk
Low. New public-data pipeline, additive, no changes to existing loaders/graph. Live runs need `SAM_GOV_API_KEY`; AI degrades gracefully without `OPENAI_API_KEY`. Recommend confirming CI is green (status is currently pending) before undrafting.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,715 @@
...
+            lines += [f"**Evidence-bounded comparison:** {_markdown_text(summary)}", ""]
+        if rationale:
+            lines += [f"**Analyst screening note:** {_markdown_text(rationale)}", ""]
+        if not summary and not rationale:
+            lines += [
+                "The screening process surfaced this pair, but no written technical comparison "
+                "is available. Compare the award abstract with the solicitation's performance "
+                "requirements before routing.",
+                "",
+            ]
```
hollomancer, line 449: **In a real run, this fallback is every card's "technical connection" — the section that's supposed to connect the dots is empty.**

The connective fields this card depends on (`alignment_rationale`, `interest_alignment`, `technology_ecosystem`, `potential_transition_lane`) exist only in the hand-authored example CSV — `score_candidate_pairs()` never emits them. And the AI summary requires `--ai` + `OPENAI_API_KEY`, HIGH confidence, both texts present, and falls under the 10-summary cap. I verified this by regenerating the Army example with a production-shaped candidates file (only the columns the scorer emits): every card's "Technical connection to validate" collapses to this same boilerplate sentence, and "Why this was surfaced" loses mission interest and technology tags entirely.

Since this section is the PCR's main bridge between the SBIR firm/abstract and the solicitation requirement, consider building a deterministic connection paragraph from fields already present in `master`:

1. **Same-firm signal** — when the notice's `awardee_uei` matches the award `uei` (which is the primary join in `pair_filter_s2`, so it holds by construction for exact directed matches), say so: "This notice names the SBIR awardee (UEI …)." That's the single strongest statement you can make on a directed card, and it's currently never rendered.
2. **Quote the matched lineage phrase** when `lineage_language_score > 0` — "The notice text contains 'sole source' and 'Phase III'" — instead of the abstract label "lineage language in the notice". The terms are already enumerated in `_LINEAGE_TERMS`.
3. **List the shared technical terms** driving the Jaccard component — the token intersection between award title+abstract and solicitation description ("Both describe: autonomous, obstacle, reconnaissance, lidar…"). The renderer already has both texts; the tokenizer would need to live somewhere `sbir_etl` can import (lifting `_tokenize` out of `sbir_analytics.assets…similarity` avoids a reverse package dependency).
4. **Name the actual shared codes** — "Both are coded NAICS 541715" rather than "matching NAICS codes".

All four are derivable with no new inputs, keep the packet deterministic, and turn the AI path into an enhancement instead of the only connective tissue.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,715 @@
...
+        if _same_text(row.get("award_psc_code"), row.get("opportunity_psc_code")):
+            ranking_signals.append("matching product/service codes")
+        if ranking_signals:
+            lines.append(f"- Ranking signals: {', '.join(ranking_signals)}")
+        score = _number(row.get("candidate_score"))
+        rank = "Priority lead" if confidence == "HIGH" else "Evidence watchlist"
+        score_text = f"; composite {score:.2f}" if score is not None else ""
+        lines.append(
+            f"- Screening rank: {rank}{score_text}. This is a triage rank, not a probability."
+        )
```
hollomancer, line 483: **Plain-language: these bullets render scoring-internal vocabulary, not facts a PCR can act on.**

"agency continuity", "topical similarity across codes and text", "critical-technology alignment", and "composite 0.87" are the model's signal names. A representative can't verify or explain any of them upward. Two changes would help:

- Render the underlying **values**, not the signal labels: "The same agency funded both (Department of the Army)", "Award and notice share NAICS code 541715", "The notice text contains 'Phase III'". Everything needed is already in the merged row.
- For the score, either keep it out of the reader-facing card (it survives in `master_candidates.csv` / `audit_sample.csv` for the audit) or translate it — "matched 4 of 6 screening signals" is meaningful; "composite 0.87" is false precision that the very next sentence has to disclaim.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,715 @@
...
+            for _, row in watch.iterrows():
+                lines += self._candidate_card(row)
+
+        if not awards.empty:
+            lines += [
+                "## Award pipeline",
+                "",
+                "| Award work | Company | Phase | Awarded | Recorded end | Amount |",
+                "|---|---|---|---|---|---:|",
+            ]
```
hollomancer, line 599: **The award table never says why an award is in the packet.**

`build_award_cohorts()` computes five inclusion flags (`newly_observed`, `changed_since_prior_report`, `awarded_in_period`, `recent_recorded_end`, `approaching_recorded_end`) but none of them is rendered — they're used only to filter. So the table can't answer the PCR's first question: "why is this award on my desk this month, and what does its timing suggest?" The PR description promises "recently awarded, completed, changed, or end-date-near SBIR records", but the packet doesn't distinguish them.

Suggest a plain-language "Why listed" column — "Newly awarded this month", "Recorded end date 2026-07-31 (within 6 months)", "Record changed since last report" — and sorting by nearest recorded end so time-sensitive rows surface first. Relatedly, "Award cohort" in the Snapshot is analyst vocabulary; "Awards in this packet: 5 (new, changed, or ending soon)" says the same thing in the reader's terms.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,715 @@
...
+        lines = [
+            f"# Monthly Procurement Transition Packet — {_markdown_text(center, limit=160)}",
+            "",
+            f"**Reporting month:** {self.report_month}",
+            "",
+            "> Use this packet to route leads for human review. Recorded performance-period "
+            "end dates do not verify technical completion, and screening results do not prove "
+            "Phase III eligibility.",
+            "",
+            "## Snapshot",
```
hollomancer, line 541: **The packet never defines its own vocabulary — add a short "How to read this packet" section.**

"Directed" vs. "competitive", "priority lead" vs. "needs more evidence before routing", and what the screen can and cannot claim are defined in `docs/procurement-transition-report.md` — but a PCR receives only the packet. Five plain sentences up front would let it stand alone, e.g.: what a directed candidate is (the government signaled it may award directly, and the evidence points back to SBIR/STTR-funded work), what a competitive follow-on is (an open competition whose subject overlaps a recent award — not a Phase III claim), what "priority lead" means (passed more screening checks; still requires human validation), and the one-sentence statutory Phase III test.

A standing definitions section would also let you slim the per-card repetition — the identical "Review question" and near-identical "Representative check" paragraphs on every card add length without adding information once the framing is stated once.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,715 @@
...
+            f"- Screening rank: {rank}{score_text}. This is a triage rank, not a probability."
+        )
+
+        lines += ["", "#### Representative check", ""]
+        if signal == "directed":
+            lines.append(
+                "Confirm that the solicitation derives from, extends, or uses the cited "
+                "SBIR/STTR-funded work, and that the funded capability covers the stated need. "
+                "The screening score does not establish statutory Phase III lineage."
+            )
```
hollomancer, line 491: **Use the SBA Policy Directive's exact phrasing: "derives from, extends, or completes."**

The statutory/directive test for Phase III work is that it *derives from, extends, or completes* prior SBIR/STTR-funded work; this renders "derives from, extends, or uses". PCRs apply the directive's wording daily — matching it exactly both reads as authoritative and avoids the impression the tool applies a looser (or different) test than the one the representative must document.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -0,0 +1,715 @@
...
+                        (award_cohorts["agency"].fillna("") == str(center))
+                        | (award_cohorts["branch"].fillna("") == str(center))
+                    ]
+                markdown = self._packet(str(center), rows, related_awards)
+                slug = _slug(rows.iloc[0].get("center_code") if not rows.empty else center)
+                (centers_dir / f"{slug}.md").write_text(markdown, encoding="utf-8")
+                (centers_dir / f"{slug}.html").write_text(
+                    "<!doctype html><meta charset='utf-8'><pre>" + html.escape(markdown) + "</pre>",
+                    encoding="utf-8",
+                )
```
hollomancer, line 675: **The `.html` artifact is not readable — it's raw markdown in a `<pre>` block.**

A PCR who opens it sees literal `**Disposition:**`, `##` headings, `&gt;` blockquotes, and the backslash escapes `_markdown_text` inserts (an abstract mentioning "C#" renders as `C\#`). Since the HTML file is the version most likely to be opened from an email attachment, this is the worst-rendered artifact in the packet. Either drop the `.html` until it renders properly, or convert the markdown for real — the output grammar is small and fixed (headings, bold, blockquote, bullets, one table), so even a minimal deterministic converter would do; alternatively a `markdown` library dependency.

---

hollomancer (commented):
Review focused on PCR-readability, firm/abstract ↔ solicitation dot-connecting, and plain language, per request. Six line comments on `sbir_etl/reporting/procurement_transition/core.py`; summary below.

## The central finding: the example packet oversells what a production packet says

The Army example reads well because its connective tissue (`alignment_rationale`, `interest_alignment`, `technology_ecosystem`, `potential_transition_lane`) is hand-written into the example candidates CSV. The production scorer (`score_candidate_pairs()`) emits none of those columns, and the AI summary path is gated on `OPENAI_API_KEY` + HIGH confidence + both texts present + a 10-card cap. I regenerated the example using a candidates file with only the columns the scorer actually emits: **every card's "Technical connection to validate" degrades to the same boilerplate sentence**, and mission interest / technology tags / suggested routing disappear. In other words, the section that connects the SBIR abstract to the solicitation requirement is empty by default in real runs — which is exactly the concern this review was asked to check.

The fix doesn't require AI: the merged row already contains everything needed for a deterministic connection paragraph — same-firm UEI match (the primary directed join, never rendered as a fact today), the actual matched lineage phrase ("sole source", "Phase III") rather than the label "lineage language", the shared technical terms behind the Jaccard score, and the actual shared NAICS/PSC codes. Details in the line comment on `_candidate_card`.

## Plain-language and readability

- **Ranking signals render model vocabulary, not facts** — "agency continuity, topical similarity across codes and text… composite 0.87" means nothing to a PCR. Render the values ("The same agency funded both"; "Both coded NAICS 541715") and translate or drop the composite number.
- **No self-contained definitions** — "directed", "competitive follow-on", "priority lead" are defined in docs the PCR will never see. A 5-sentence "How to read this packet" section would let the packet stand alone and permit trimming the repeated per-card boilerplate.
- **Award table lacks "why listed"** — the five cohort flags (new / changed / awarded this month / ended / ending within 180 days) are computed but never rendered, so the pipeline table can't answer the reader's first question or convey urgency. "Cohort" is also analyst jargon.
- **Statutory phrasing** — "derives from, extends, or **uses**" should be the directive's "derives from, extends, or **completes**"; PCRs apply that exact test.
- **The `.html` artifact is unreadable** — escaped markdown inside `<pre>`, including literal `**`, `##`, and backslash escapes. That's the file most likely to be opened from an email.

## What's solid

The report architecture is genuinely good for this audience: evidence-first card layout (award scope beside solicitation need), honest fallbacks when text is missing, watchlist kept separate from priority leads, scores framed as triage rather than probability, and the audit gate before distribution. Code quality is also careful where it matters — untrusted public text is escaped and truncated, the SAM.gov description fetch is host-allowlisted, pagination is bounded with explicit refusal on truncation, hydration is capped with recorded coverage stats, and the AI path is validation-gated (uncited output discarded) and can't touch scores. Unit tests cover the renderer, pairing, hydration, and extractor well.

Not re-litigated here: the existing owner comment about rebasing onto #449's award-identity foundation and the CI mypy failures — both still stand.

---

hollomancer:
Consolidated into #466 (single PR against `main` carrying all of #450 + #464 + the awardee-first reshape). Closing to avoid a lingering stack — please review #466.

PR #456, opened 2026-07-17
NASA cross-agency transition detection — TechPort source (puller + link core)

hollomancer:
NASA-focused, stacked on #451. The DoD archive ranker (#455) **does not generalize to NASA** — every firm-link key in the GSA `falextracts` archive fails for NASA (`AwardNumber` ~6%, `Awardee` ~5%, FPDS `Sol#` ~2.5%, **0 matched** our NASA Phase III firms). NASA doesn't post firm-linkable award docs to sam.gov; it runs SBIR via NSPIRES with sole-source Phase III.

**NASA's own source is TechPort** (`techport.nasa.gov/api`, public): **~20,036 SBIR/STTR projects**, ~95% of a sample with a performing firm org + rich description.

## What this adds
- `scripts/phase3_benchmark/pull_techport_nasa.py` — pure `parse_project` / `link_firm` + a **paced, cached, retried, manifested** puller (the API rate-limits under rapid calls).
- `tests/unit/scripts/test_pull_techport_nasa.py` — canned-fetcher fixture tests (**3 passed**).
- `specs/phase3-match-benchmark/nasa-techport.md` — findings, build design, and the **first-run gate**.

## First-run gate (before any NASA AUC)
Confirm TechPort SBIR projects are **transition/Phase III** vs original Phase I/II (program/phase/date fields). If transition-bearing → description is the ranker *target* (NASA analogue of the DoD J&A); if original SBIR → it's query-side. **No NASA AUC is reported until this is resolved.**

## Scope / honest limits
- Full ~20k pull **deferred** — API was throttling at authoring; needs a paced background run.
- NASA **undercount 16/202 (7.9%) still stands** (USAspending); only the *ranker's* NASA recovery was blocked.
- Org→UEI is fuzzy name matching; link rate to be measured on the real pull.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #455, opened 2026-07-17
Transition ranker + eval-validity: retrieval core, 2x2 ablation, survival analysis, blind-adjudication package

hollomancer:
## What this contains (evolved well beyond the original ranker)

**Ranker core** (`transition_ranker.py`): firm-clustered retrieval (GroupKFold-by-firm), temporal/id_xref/notice-type/TF-IDF features, tests.

**Eval-validity arc** (`specs/phase3-match-benchmark/eval-validity.md` + scripts) — external-review-driven; read this file first:
- **2×2 text-richness ablation** (`text_richness_2x2.py`): under TF-IDF, target-richness main effect +0.134–0.139, robust to metadata-hard negatives (paired CI on Δ [−0.008,+0.016]); text-reuse/plagiarism refuted.
- **Model axis** (`dense_vs_sparse_2x2.py`): richness gradient is TF-IDF-specific; **thin ≠ thin** (dense wins thin NASA 0.759 vs 0.658, loses thin DoD 0.659 vs 0.714 — information content, not length, is the variable).
- **Within-DoD number** (`dod_within_retrieval.py`): AUC **0.714** (968 firms, thin targets, metadata-hard negs). Provenance caveat stated up front: positives are FPDS-coded = the complement of the uncoded claim → upper bound.
- **Survival** (`transition_survival.py`): only 13% of DoD SBIR firms ever show coded Phase III; KM median time-to-Phase-III **not reached**; the "9y median lag" is a right-censored, conditional lower bound.
- **DoD inventory** (`dod_transition_inventory.py`): 5,091 Phase II→III positives / 974 firms; **87.5% of Phase III descriptions <150 chars** — the standalone oversight fact.
- **Retractions on the record**: the 0.945 rich-desc counterfactual (selection artifact: 81% vs 67.5% R&D-PSC); the ~900-char §638 floor (Goodhart) → corrected ask is a **structured parent-SBIR-award-ID linkage field**.

**Blind-adjudication package** (`specs/phase3-match-benchmark/adjudication/`): 90-case blind sheet (75 detector candidates + 15 dark-cell existence-scan rows), KEY (do not open until labeled), frozen protocol. This is the validation gate — precision/recall + label-noise floor land here.

14 unit tests on the pure estimator cores (KM, bootstrap, retrieval AUC, negatives) pass.

**Honest status:** detector demoted to triage + existence-proof (base-rate wall ~1% needs ~0.95 AUC); no empirical error bars until the adjudication lands.

[...]

PR #454, opened 2026-07-17
Phase III undercount at award grain (141/16) + dark-count bounds (capture-recapture, frame size)

hollomancer:
## What this contains

**Undercount reproduction** (`undercount_award_grain.py` + tests): the 141 DoD / 16 NASA described-not-coded undercount reproduced at award grain through PR #449's compound-key identity contract (coded 6,351/1,038; overlap 821/186). Manifested described-puller (`pull_described_phase3.py`) re-verified 962→141.

**Dark-count bounds** (`specs/phase3-match-benchmark/mse-dark-phase3.md`):
- 2-source capture-recapture (code × description): total ≈7,441, **dark ≈949 [767–1,130] — a LOWER bound** (positive list dependence ⇒ underestimate; direction stated).
- **Feasibility gate run: no clean structured 3rd FPDS signal exists** (set-aside is generic "SBA", not SBIR-specific; no SBIR non-competed reason) → MSE stays 2-source; J&A only as sensitivity.
- **Frame size measured**: un-flagged post-SBIR DoD contracts to the 8,064 SBIR firms ≈ **131k [71k–202k]** (uniform n=300; median 0/firm, 53% zero-firms). So dark ∈ [949, 131k]; implied Y-rate 0.7% ⇒ small-n direct validation is base-rate-blocked (15 rows expect 0.1 hits).
- Headline defensible sentence: **the code misses ≥ ~15% of DoD Phase III** (141 proven + ≥949 estimated).

**Scopes**: `cross-agency-scope.md` (DoD→NASA retrieval 0.828, 306 firms), `dod-nonsbir-contract-pool.md` (recipient-scoped puller spec, gated).

**Known API gotcha** documented: `spending_by_award` returns a misleading HTTP 500 for `start_date` before 2007-10-01.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #451, opened 2026-07-16
feat(phase3): add reproducible research benchmark

hollomancer:
Closes #448
Supersedes #423
Parent epic: #441
Related production-adapter work: #442

## Stack order

This draft is intentionally based on #449 (`agent/award-identity-dataset-contracts`) because the donor benchmark evaluated FPDS transaction rows and missed nested parent-IDV identity. After #449 merges, retarget this PR to `main`; its review diff will remain the research benchmark only.

## What this isolates

- bounded, parameterized FPDS Element 10Q ATOM puller
- namespace-safe parser that preserves the order PIID and extracts nested parent-IDV PIID
- run manifest with query, source vintage, retrieval time, parameters, counts, raw-page hash, field completeness, and feed-exhaustion status
- award-grain collapse through the shared #449 identity contract
- conservative as-of Phase II pairing that cannot select a future award
- explicit same-firm **proxy-positive** semantics and deterministic same-office hard negatives
- deterministic lexical Jaccard baseline with ROC-AUC and bootstrapped interval
- corrected requirements, design, findings, product decisions, and research-question link

## Deliberately not carried forward

The donor's `plan.md`, hard-coded local paths, import-time I/O, ad hoc embedding dependency, dark/NASA/freeze/sample scripts, missing-link report, and unsupported 191/$365M/~1,264 claims are excluded. The bounded 141/962 (14.7%, approximately $244.2M) result is retained only as an explicitly provisional lower bound requiring a rerun at award grain.

This remains research-only under `scripts/phase3_benchmark`. Production retries, checkpoints, refresh policy, and a reusable external-source API belong in #442.

## Verification

- new parser/pairing tests: **5 passed**
- stacked identity/grain and affected integration regression suite: **18 passed**
- Ruff check and format: **passed**
- mypy on both research modules: **passed**
- repository hygiene and live-document link audit: **passed**
- GitHub Actions: **fully green**, including fast/slow unit suites, integration, S3 integration, package mypy, and CI summary

The #423 donor branch is retained for provenance.

PR #458, opened 2026-07-18
Harden Phase III benchmark methodology

hollomancer:
## What changed

This is the single replacement for PRs #454, #455, and #456, stacked on the canonical benchmark foundation in #451.

- centralizes tie-corrected retrieval AUC and corrects as-of, award-row, and hard-negative cohort construction
- makes native award keys authoritative; separates contracts/IDVs and SBIR/STTR strata
- adds Chapman OR=1, list-dependence, and text-list false-positive sensitivity
- tracks per-query pagination and refuses to label bounded pulls complete
- preserves TechPort organization roles/IDs and exposes ambiguous firm linkage
- narrows survival to first observed DoD Phase II → first later observed coded Phase III action
- rewrites policy-facing claims so legacy 141/16, ~949, AUCs, and 9-year figures remain provisional rather than reproduced or identified bounds

## Why

The prior PRs mixed selected proxy cohorts with claims about the population lacking the selection property. They also contained award-identity, chronology, tie-handling, pagination, list-precision, and estimand problems that could change headline quantities or their interpretation.

## Validation

- 44 focused Phase III unit tests pass
- Ruff passes for the benchmark scripts and focused tests
- mypy passes for 8 changed estimator files
- `git diff --check` passes
- stale-claim searches pass

## Reproduction status

The gitignored coded/described frames and TechPort cache were not available. Corrected empirical commands therefore emit `blocked_missing_inputs`; historical values are retained in `provisional-results.json` as legacy/provisional entries, not silently accepted as reproduced.

## Before policy use

Regenerate the manifested inputs, rerun all estimators, and complete the pending blind adjudication before treating any direct flag count, capture scenario, retrieval score, or survival summary as policy evidence.


specs/phase3-match-benchmark/mse-dark-phase3.md
```diff
@@ -0,0 +1,397 @@
...
+"SBIR PHASE III" description) cannot be reached by a classifier — at ~1% prevalence that needs ~0.95 AUC,
+unreachable by text, and DoD's descriptions are empty anyway (`eval-validity.md`: median 43 chars; a dense
+model does *worse* on that boilerplate, 0.659 vs 0.714). **Capture-recapture sidesteps the wall entirely**:
+it never classifies a contract; it infers the total from the overlap of incomplete high-precision *lists* —
+the method a census uses to estimate its own undercount and epidemiology uses for hidden populations.
+
+**Already computed (2-source, Lincoln-Petersen/Chapman):** code (6,351) × description (962), overlap 821 →
+total ≈ 7,441, **dark ≈ 949 [95% CI 767–1,130]**. But two-source *assumes the signals are independent*, and
+they are positively correlated (a CO who codes SR3 also writes "SBIR PHASE III"), which inflates the overlap
+and **underestimates** the total — so **949 is a lower bound, not an estimate.** A third list breaks this: it
```
hollomancer: **Blocker:** this reinstates the claim that 949 is a lower bound, contradicting `findings.md` and `provisional-results.json`, which correctly call it the OR=1 independence scenario. Positive dependence does not establish a bound when list precision, heterogeneous capture, and reachability are unknown. Please remove all lower-bound/floor language and report 949 only with the list-OR and false-positive sensitivity.

specs/phase3-match-benchmark/mse-dark-phase3.md
```diff
@@ -0,0 +1,397 @@
...
+
+Log-linear Poisson, predicted 000 cell by dependence structure:
+
+| model | dark n000 | total N |
+|---|--:|--:|
+| independence | 1,154 | 7,866 |
+| [AB] | 1,860 | 8,572 |
+| [AC] | 468 | 7,180 |
+| [AB+AC] | 838 | 7,550 |
+| **[AB+BC] (best AIC, df≥1)** | **2,078 [1,794–2,372]** | 8,790 |
```
hollomancer: **Blocker:** list C is only reachable for task orders—standalone contracts have a structural zero because they have no parent vehicle. Pooling standalone and task-order units in one homogeneous ABC model makes this estimate unidentified for the stated population. C is also assigned at vehicle level and expanded to correlated child orders, so the contract-level CI is too narrow. Cheapest resolution: fit AB for standalone contracts, ABC for task orders, sum the dark cells, and block-bootstrap by vehicle and firm.

specs/phase3-match-benchmark/mse-dark-phase3.md
```diff
@@ -0,0 +1,397 @@
...
+| [AB+AC+BC] (all-pairwise, df=0) | **1,543 [915–2,745]** | 8,255 |
+
+**Reading:** the AB term is strongly AIC-favored — direct confirmation that code×description dependence is
+real and the 2-source 949 was biased low, exactly as the stratified floors (1,007/1,205) indicated. Across
+credible structures **dark ≈ 500–2,400**, headline **all-pairwise 1,543 [915–2,745]**, implying the code
+misses **~19–28%** of DoD Phase III (total N ≈ 7,200–8,800).
+
+**Sensitivities that adjudication will settle:** list-C precision < 1 (admin orders on declared vehicles
+inflate the 271 C-only cell → inflates N; the sibling stratum C091–C105 measures this); list-B false
```
hollomancer: The C091–C105 adjudication stratum does **not** estimate list-C precision. List C is `child of a self-declared Phase III vehicle`; C091–C105 instead sample `unflagged sibling on a high coded-order-purity vehicle`. Those are different capture rules. Draw a random, vehicle-clustered sample directly from C-only orders on self-declared vehicles, stratified by vehicle purity/size.

specs/phase3-match-benchmark/mse-dark-phase3.md
```diff
@@ -0,0 +1,397 @@
...
+| [AB] | 1,860 | 8,572 |
+| [AC] | 468 | 7,180 |
+| [AB+AC] | 838 | 7,550 |
+| **[AB+BC] (best AIC, df≥1)** | **2,078 [1,794–2,372]** | 8,790 |
+| [AB+AC+BC] (all-pairwise, df=0) | **1,543 [915–2,745]** | 8,255 |
+
+**Reading:** the AB term is strongly AIC-favored — direct confirmation that code×description dependence is
+real and the 2-source 949 was biased low, exactly as the stratified floors (1,007/1,205) indicated. Across
+credible structures **dark ≈ 500–2,400**, headline **all-pairwise 1,543 [915–2,745]**, implying the code
+misses **~19–28%** of DoD Phase III (total N ≈ 7,200–8,800).
```
hollomancer: The stated 19–28% range does not follow from the table. There are 361 directly observed code-negative units, so the AC row gives `(361+468)/7,180 = 11.5%`; all-pairwise gives 23.1%; AB+BC gives 27.7%. The displayed dark range therefore implies roughly **12–28%**, not 19–28%. Please add model-specific miss rates rather than combining one model's lower endpoint with another model's dark-count range.

specs/phase3-match-benchmark/adjudication/phase3_adjudication_protocol.md
```diff
@@ -0,0 +1,87 @@
...
+  "sibling-of-coded" as a capture rule; each confirmed Y is an observed uncoded Phase III. **Blinding note:**
+  appended mid-marking, so this block is knowably all-sibling (a reshuffle would have lost your in-progress
+  labels). No model scores exist for these; judge tech continuity exactly as elsewhere. Note the firm shown
+  is the VEHICLE holder's dominant coded vendor — verify same-firm as usual.
+
+- **GSA-channel stratum (C106–C115, appended 2026-07-19)** — described "SBIR PHASE III" contracts
+  **awarded by GSA AAS, DoD-funded** (the assisted-acquisition channel: $1.77B, 0/189 carry the 10Q code).
+  These are text-flagged (like the miscoded 141) but from a different awarding pipeline with rich FEDSIM
+  descriptions — expected to adjudicate easily. Their Y-rate = precision of the described list in the GSA
+  channel, which underwrites the $1.77B figure. Same append-only blinding note as the sibling stratum.
```
hollomancer: These ten firm-linked, one-per-firm cases cannot underwrite the full $1.77B across all 189 GSA awards. They exclude records without the required firm/Phase-II linkage and are not dollar-weighted. Sample directly from all 189 awards, adjudicate Phase III status separately from specific SBIR lineage, and report both award- and dollar-weighted precision with CIs.

specs/phase3-match-benchmark/mse-dark-phase3.md
```diff
@@ -0,0 +1,397 @@
...
+
+**Sensitivities that adjudication will settle:** list-C precision < 1 (admin orders on declared vehicles
+inflate the 271 C-only cell → inflates N; the sibling stratum C091–C105 measures this); list-B false
+positives symmetric. The all-pairwise model has 0 df (an estimator, not a testable fit); [AB+BC] is the best
+*testable* model. All values provisional pending re-run through #458's estimator discipline.
```
hollomancer: All new vehicle/NASA/GSA results currently exist only as prose and CSV output—there is no committed cell-building/MSE script, frozen input manifest/hash, or update to `provisional-results.json`. Green CI does not reproduce these figures. Before merge, add the exact extraction/cell/model code plus machine-readable expected results and provenance.

hollomancer (commented):
Methodology review of the post-rebase additions. The published seven-cell arithmetic reproduces, but the population model, precision calibration, and headline translation do not yet support the new policy-facing estimates. The pooled three-list estimate and 19–28% range should remain provisional and non-headline until the structural-zero/cluster issues below are resolved.

specs/phase3-match-benchmark/mse-dark-phase3.md
```diff
@@ -0,0 +1,397 @@
...
+  self-declared Phase III vehicles (vs DoD 149), 3 with zero coded orders; list C = 40 orders, C-only = 16
+  (small direct-observation set: Barron, Sunpower, Fiber Materials, Bally Ribbon vehicles).
+- **3-source fit is underpowered for NASA** — the within-C window has a zero cell (B∩C̄∩A̅... B-only-C = 0),
+  so the AB odds ratio is degenerate (n=40). Do not fit; report the 2-source ≈73 with the dependence caveat.
+
+**Why this strengthens the DoD story:** NASA is the natural control. Where the award/administration
+separation and Phase-III-vehicle structure are absent (no DCMA analogue, few dedicated vehicles, centralized
+contracting), the order-level miss rate halves and the vehicle-mediated loss goes to ~zero. The DoD failure
+is not "government can't code Phase III" — it is specifically the **order-under-vehicle handoff**, present
+where that structure is used and absent where it isn't. This is the mechanism claim surviving an
```
hollomancer: NASA is a descriptive comparator, not a causal control: agency, contract mix, year, firm composition, vehicle use, and coding process all differ without matching or standardization. This supports a negative-control comparison, but not an out-of-sample confirmation of the DoD handoff mechanism. Please narrow the claim or standardize by contract type, year, and vehicle status.

specs/phase3-match-benchmark/adjudication/phase3_adjudication_protocol.md
```diff
@@ -0,0 +1,87 @@
...
+- `specs/phase3-match-benchmark/adjudication/phase3_transition_leads_audit_v2.csv` — the score-visible source (your original 71 labels live here; treat as *provisional/anchored*).
+
+## The unit and the question
+One row = one **(firm, candidate-event)** pair. Candidate = a DoD notice, a NASA TechPort project, or a
+**DoD contract** (see the dark-cell stratum). **Question:** does this candidate event represent a Phase III /
+transition of *this firm's SBIR-funded technology*? Apply the same rule to every row regardless of type.
+
+## The strata (hidden in the KEY — you can't tell which is which while labeling)
+- **detector candidates** (60 DoD notices + 15 cross-agency NASA projects) — top-ranked by the model. Labeling
+  these gives the detector's **precision/recall**.
```
hollomancer: This design cannot estimate population recall: detector cases are selected top-ranked leads, while the 15 random dark-cell cases expect about 0.1 positives and are explicitly demoted to an existence scan. It can estimate selected-lead precision and Unknown rate. Please remove `recall` and update the stale 90/75-case references now that the files contain 115 cases.

hollomancer:
## Methodology review — all seven blockers addressed in 8ba770c9

1. **Lower-bound language (blocker):** removed throughout `mse-dark-phase3.md`. 949 is reported as the OR=1 independence scenario consistent with `findings.md`/`provisional-results.json`; all stratified figures are labeled model scenarios, never bounds.
2. **Structural zero / clustering (blocker):** re-fit exactly as prescribed — standalone contracts as a 2-list A×B stratum (Chapman), task orders as the 3-list ABC stratum, dark cells summed; block bootstrap resamples task orders by parent vehicle. Committed as `scripts/phase3_benchmark/vehicle_mse.py` with 6 unit tests (incl. structural-zero and known-N recovery), frozen inputs + SHA-256 manifest under `specs/phase3-match-benchmark/inputs/`, machine-readable output. Corrected results: dark 380–1,380 across models; [AB+BC] CI [850, 2,485]; the saturated model is unstable under cluster resampling (CI to ~19k) — disclosed.
3. **C091–C105 ≠ list-C precision:** conceded — different capture rule. Added a new `list_c_only` stratum (C116–C127): random, vehicle-clustered sample from C-only orders on self-declared vehicles. Protocol documents the sibling stratum as measuring the sibling rule only.
4. **19–28% arithmetic:** retracted. Per-model miss rates now computed within each model: **10.5–21.5%** (table in the spec).
5. **GSA $1.77B underwriting:** claim narrowed — C106–C115 estimate firm-linked-lead precision only; the spec records the proper design (dollar-weighted random sample of all 189, Phase-III status adjudicated separately from lineage) as future work.
6. **Reproduction:** `vehicle_mse.py` + tests + frozen input caches + `provisional-results.json` entries (`dark_structure_stratified_models`, `gsa_awarded_described_phase3`, `self_declared_phase3_vehicles_dod`) committed. The GSA/NASA probes remain prose-recorded and are marked `arithmetic_reproduced_inputs_unverified` accordingly.
7. **NASA causal claim:** narrowed to descriptive negative-control comparator; the only standardized cut (standalone-vs-standalone, 9.3% vs 8.4%) is noted as such.
8. **Protocol recall/stale counts:** "recall" removed (design identifies selected-lead precision + Unknown rate only); counts updated to 127 cases across 5 strata.

The pooled 1,543/2,078 figures and the 19–28% range are superseded and should not be quoted; the corrected headline posture is "141 description-only flags + dark model-scenarios 380–1,380 (structure-stratified, list-C precision unadjudicated)."

hollomancer:
## Re-review — CI unblocked, last open thread closed (`02f7452`)

**CI was red for one reason only:** `Code Quality` failed on `Would reformat: tests/unit/scripts/test_vehicle_mse.py` — a single unformatted file, nothing methodological. Ran `ruff format` on it; the CI scope (`sbir_etl packages/*/… tests`) is now clean on both `ruff check` and `ruff format --check`. Every other check on the branch was already green.

**The seven blocker threads are all genuinely resolved by `8ba770c`,** verified against the file contents rather than the commit message:

| Thread | Status |
|---|---|
| 949 as a lower bound | Resolved — all lower-bound/floor language gone; reported as the OR=1 independence scenario |
| List C structural zero for standalone contracts | Resolved — AB for standalone, ABC for task orders, dark cells summed, vehicle-block bootstrap (`vehicle_mse.py`, frozen inputs + SHA-256 manifest) |
| C091–C105 doesn't measure list-C precision | Resolved — new `list_c_only` stratum C116–C127, random and vehicle-clustered from C-only orders |
| 19–28% doesn't follow from the table | Resolved — retracted; per-model rates 10.5–21.5% |
| Ten GSA cases can't underwrite $1.77B | Resolved — scoped to firm-linked leads, and the doc now says it does *not* underwrite the $1.77B |
| No committed cell-building/MSE code or machine-readable results | Resolved — `vehicle_mse.py` + 6 unit tests + `inputs/*.json` + `provisional-results.json` extended |
| NASA as causal control | Resolved — narrowed to descriptive negative-control comparator |

I've resolved those seven threads.

**The one still-open thread was partly unaddressed.** The `recall` removal landed in the "What this produces" section (line 97) but two earlier passages still contradicted it: the detector-candidate stratum said labeling "gives the detector's **precision/recall**", and the freeze section still said "This **75-case** set" after the appends took it to 127. Both now say selected-lead precision over 127 cases, with a one-line explanation of why recall isn't identified — the design never observes positives the model failed to rank. That thread is now closed.

**Mergeability: not yet, and not because of CI.** This is a 39-file, ~4,280-line draft carrying the entire Phase III methodology corpus. Two things stand between it and merge:

1. It adds `scripts/phase3_benchmark/transition_ranker.py`, which **#467 also adds** with a small diff. #467 is the merge-ready one; land it first and rebase this branch onto it, or these conflict.
2. It's still marked draft, and the volume of methodology here deserves a deliberate "these numbers are the ones we stand behind" pass rather than inheriting green CI as approval. The figures are now internally consistent and reproducible, which is what the blockers asked for — but consistency isn't the same as sign-off.

Nothing new surfaced in this pass beyond the protocol wording above.

---

scripts/phase3_benchmark/vehicle_mse.py
```diff
@@ -0,0 +1,182 @@
...
+        cells = np.array([counts.get((bool(p[0]), bool(p[1]), bool(p[2])), 0) for p in CELL_ORDER], float)
+        fitted = loglinear_dark(cells, terms)
+        if fitted is not None:
+            estimates.append(dark_stand + fitted["n000"])
+    lo, hi = np.percentile(estimates, [2.5, 97.5])
+    return float(lo), float(hi)
+
+
+def _sha(path: Path) -> str:
+    return hashlib.sha256(path.read_bytes()).hexdigest()
```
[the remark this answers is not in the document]
hollomancer: **Agree, and the repo already has the pattern twice. Leaving open — it's a one-line change.**

`_sha` is called on `args.coded` and `args.described`, the two m0a parquet frames, and it runs inside the `result = {...}` literal at the top of `main()` — so a memory spike there kills the run before any estimation happens, and the failure looks like a crash rather than "your input is too big."

Existing streaming implementations to match: `_file_sha256` in `packages/sbir-analytics/sbir_analytics/assets/transition/contracts.py` and `_sha256` in `sbir_etl/extractors/usaspending_award_archive.py`. Both use the same 1 MiB `iter(lambda: ..., b"")` idiom:

```python
def _sha(path: Path) -> str:
    digest = hashlib.sha256()
    with path.open("rb") as stream:
        for chunk in iter(lambda: stream.read(1024 * 1024), b""):
            digest.update(chunk)
    return digest.hexdigest()
```

Byte-identical output, so the hashes already committed in `vehicle_mse_result.json` stay valid.

Worth doing precisely because this hash is load-bearing: it is the mechanism that closed the "green CI does not reproduce these figures" thread from the July round. A provenance hash that can OOM on its own input is the wrong thing to have be fragile.

---

specs/phase3-match-benchmark/inputs/vehicle_mse_result.json
```diff
@@ -0,0 +1,92 @@
+{
+  "inputs": {
+    "/Users/hollomancer/projects/sbir-analytics/data/derived/m0a_coded_dod.parquet": "42207a95a341e4b3256c3223514b56a5071341a0e2050d4078000564f8ce3120",
+    "/Users/hollomancer/projects/sbir-analytics/data/derived/m0a_desc_phase3_dod.parquet": "e803b5fd083072f489f65ead8bcc22e5744117b1f6ac05989d26674f3ea26175",
+    "specs/phase3-match-benchmark/inputs/vehicle149_children.json": "7b8cfa0a128612dc2d86bd796c4ef9831a276bc7a332656c124a192f98094ffc"
```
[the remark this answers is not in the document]
hollomancer: **Agree, and it's worse than non-portable — the absolute paths are dict *keys*, so the provenance record can never be matched by anyone else. Leaving open.**

```json
"inputs": {
  "/Users/hollomancer/projects/sbir-analytics/data/derived/m0a_coded_dod.parquet": "42207a95...",
  "/Users/hollomancer/projects/sbir-analytics/data/derived/m0a_desc_phase3_dod.parquet": "e803b5fd...",
  "specs/phase3-match-benchmark/inputs/vehicle149_children.json": "7b8cfa0a..."
}
```

The third entry is repo-relative because it's a committed cache and the default path is relative; the first two are absolute because whoever ran it passed `--coded`/`--described` explicitly. So the file already demonstrates both conventions side by side.

The consequence is specific. `main()` builds this dict as `{str(p): _sha(p) for p in (...)}`, keyed on whatever `Path` came off the CLI. A second person regenerating the frames at `data/derived/...` produces a dict with *different keys* — and identical hashes tell them nothing, because there's no shared key to compare on. The record can't be diffed, which is the one thing a provenance record has to support.

That matters more than usual here because of what this file is for. The July review blocked on "no committed cell-building/MSE script, frozen input manifest/hash, or update to `provisional-results.json` — green CI does not reproduce these figures," and `vehicle_mse_result.json` is the artifact that answers it. Right now it certifies reproducibility against two files on one laptop.

Fix is small and matches the existing third entry: normalize to repo-relative before keying, e.g.

```python
def _key(path: Path) -> str:
    resolved = path.resolve()
    try:
        return resolved.relative_to(REPOSITORY_ROOT).as_posix()
    except ValueError:
        return path.name  # or fail loudly — an input outside the repo isn't reproducible
```

The `except` branch is the real decision: an input that lives outside the repo is arguably not a reproducible input at all, and refusing to record it may be more honest than recording a bare filename. Either way the committed JSON needs regenerating, since the keys change.

---

specs/phase3-match-benchmark/inputs/vehicle_mse_result.json
```diff
@@ -0,0 +1,92 @@
...
+    "AB+AC+BC": {
+      "dark_standalone": 223.0,
+      "dark_task_order": 848.4,
+      "dark_total": 1071.4,
+      "aic_task_order": -49605.8,
+      "total_N": 7783.0,
+      "code_miss_rate": 18.4
+    }
+  },
+  "ab_bc_block_bootstrap_ci": [
```
hollomancer, line 88: **A single-model CI is placed as a sibling of `models`, where it reads as the interval for the whole analysis — and the across-model spread falls partly outside it.**

`ab_bc_block_bootstrap_ci: [850, 2485]` is computed from one hardcoded call, `block_bootstrap_ci(units, ["AB", "BC"])`. Its point estimate is the AB+BC row's `dark_total = 1380.1`. But in the JSON it sits at the top level next to `n_units_observed` and `models`, not inside `models["AB+BC"]`, so nothing in the structure ties it to the model it came from except the key name.

Lay the eight `dark_total` values against it:

| model | dark_total | inside [850, 2485]? |
|---|---|---|
| AC | 382.4 | no |
| AB+AC | 395.4 | no |
| AC+BC | 499.3 | no |
| independence | 996.7 | yes |
| AB+AC+BC | 1071.4 | yes |
| BC | 1117.4 | yes |
| AB | 1174.6 | yes |
| AB+BC | 1380.1 | yes |

Three of eight models land below the interval's lower endpoint. So model uncertainty here is not merely comparable to sampling uncertainty — it extends outside it, in the direction that matters for a "how much is missed" claim. A reader quoting `dark_total` from the AC row and the interval from the top level would report `382 [850, 2485]`, which is incoherent.

This is the same composition error that blocked in the July round — "please add model-specific miss rates rather than combining one model's lower endpoint with another model's dark-count range" — and that thread's fix landed correctly in the code: `code_miss_rate` is now computed per model. The JSON layout then reintroduces it one level up, for the interval instead of the rate.

Two changes, both mechanical:

1. Move the CI inside its model entry (`models["AB+BC"]["dark_total_ci"]`), so it cannot be read as analysis-wide. Better still, compute it for every model in the `MODELS` loop — it's the same `block_bootstrap_ci(units, terms)` call and the loop already has `terms` in hand.
2. If only one model gets an interval, record which and why. Right now `["AB", "BC"]` is a bare literal at the bottom of `main()` with nothing stating it is the AIC-preferred model — and by AIC it *is* the best at -49607.1, but only by 1.3 over the saturated model, which is well inside noise.

The prose in `mse-dark-phase3.md` may already handle this carefully; the concern is the machine-readable artifact, since that is what a later script or a reader in a hurry will pull numbers from.

---

scripts/phase3_benchmark/vehicle_mse.py
```diff
@@ -0,0 +1,182 @@
...
+import hashlib
+import json
+import re
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+
+CELL_ORDER = [(1, 1, 1), (1, 1, 0), (1, 0, 1), (0, 1, 1), (1, 0, 0), (0, 1, 0), (0, 0, 1)]
+MODELS = [[], ["AB"], ["AC"], ["BC"], ["AB", "AC"], ["AB", "BC"], ["AC", "BC"], ["AB", "AC", "BC"]]
```
hollomancer, line 28: **`["AB", "AC", "BC"]` is the saturated model — 7 parameters on 7 cells — so its dark cell is pure functional-form extrapolation with zero residual degrees of freedom, yet it's reported as a peer row.**

`_design` returns `["1", "A", "B", "C"] + terms`, so the three-way-interaction entry gives a 7×7 design against `CELL_ORDER`'s 7 observed cells. The fit is exact by construction: every `mu` equals its `cell`, the residual df is 0, and `n000 = exp(beta[0])` is determined entirely by the choice to omit the ABC term — not by anything the data could have contradicted. There is no goodness-of-fit signal left to distinguish it from a different extrapolation.

That would be a footnote if AIC separated it, but it doesn't. From the committed result:

- `AB+BC`: aic -49607.1, dark_total 1380.1
- `AB+AC+BC`: aic -49605.8, dark_total 1071.4

A gap of 1.3 AIC units. Conventionally that's indistinguishable, so a reader applying "pick the lowest AIC" has no principled way to reject a model whose estimate rests on no residual information at all. And its `dark_total` sits in the middle of the range, so it looks reassuringly moderate rather than degenerate.

This is the class of problem the July round was about — the resolved threads pushed hard on unidentified models (the pooled ABC structural-zero thread) and on "no lower-bound semantics." The structural zero was fixed properly by splitting the strata. The saturated-model case is the remaining instance of the same underlying issue: a row whose number is not identified by the data appearing next to rows that are.

Three defensible options:

1. **Drop it** from `MODELS`. With 7 cells the identifiable hierarchical set is the seven models with at least one residual df; the saturated model adds no information.
2. **Keep it, mark it.** Emit `residual_df` per model (`7 - design.shape[1]`) so `AB+AC+BC` is visibly `0`, and say in the docstring that a zero-df row is reported for completeness and is not a candidate for selection.
3. **Report the AIC spread explicitly** — if the top models are within ~2 units, the honest summary is "AIC does not discriminate among these," which is itself a finding worth stating given the estimates range 382–1380.

(2) is the smallest and probably the most useful — one derived integer makes the distinction self-evident in the JSON, and it costs nothing.

Separately, and much smaller: `MODELS` omits `["A"]`-style main-effect-only variants because main effects are always included, which is correct — worth a one-line comment saying so, since the list otherwise looks like an arbitrary subset of interaction combinations rather than the complete hierarchical set.

---

scripts/phase3_benchmark/vehicle_mse.py
```diff
@@ -0,0 +1,182 @@
...
+        n1, n2, m = int(s[:, 0].sum()), int(s[:, 1].sum()), int((s[:, 0] & s[:, 1]).sum())
+        dark_stand = chapman(n1, n2, m) - (n1 + n2 - m)
+        picked = np.vstack([clusters[i] for i in rng.randint(0, len(clusters), len(clusters))])
+        counts: dict = {}
+        for row in picked:
+            counts[tuple(row)] = counts.get(tuple(row), 0) + 1
+        cells = np.array([counts.get((bool(p[0]), bool(p[1]), bool(p[2])), 0) for p in CELL_ORDER], float)
+        fitted = loglinear_dark(cells, terms)
+        if fitted is not None:
+            estimates.append(dark_stand + fitted["n000"])
```
hollomancer, line 128: Non-converging bootstrap replicates are dropped silently, and the count never reaches the output.

```python
fitted = loglinear_dark(cells, terms)
if fitted is not None:
    estimates.append(dark_stand + fitted["n000"])
```

`loglinear_dark` returns `None` on `LinAlgError` — a singular `weighted @ design`, which is exactly what a resampled cell pattern produces when a rare cell (`BC` has 22 units, `B` has 14) draws to zero. Those aren't random failures: they're concentrated on replicates where the sparse cells are underrepresented, which is the same tail that drives the dark-cell estimate. Dropping them isn't neutral, it's a selection on the statistic being bootstrapped.

The 95% interval is then computed over however many survived, with `n_boot=600` claimed nowhere and verified nowhere. If 590 survive, fine. If 300 do, `[850, 2485]` is a percentile of a filtered subsample and the coverage claim doesn't hold. A reader of `vehicle_mse_result.json` cannot tell which happened.

Also note the loop can exhaust its 300 IRLS iterations without converging and still return a result — the `break` only fires on the tolerance check, and there's no `else` clause or convergence flag, so a non-converged `beta` is returned as if it had converged. That's a second silent path to a bad `n000`, and it won't show up as a `None`.

Minimal fix — return the count and surface it:

```python
def block_bootstrap_ci(...) -> tuple[float, float, int]:
    ...
    return float(lo), float(hi), len(estimates)
```

then record `"n_boot_effective": n_effective` next to the interval. If it's materially below `n_boot`, that's a result about the data (the sparse cells can't support a cluster bootstrap at this size), not a nuisance to hide.

Two smaller things in the same function:

- `np.percentile(estimates, ...)` raises `IndexError` on an empty list. Given every replicate can fail for a degenerate `terms`/`units` combination, a clear error beats a numpy traceback.
- `rng.randint(0, len(clusters), len(clusters))` resamples the right number of clusters, but cluster sizes vary, so the resampled task-order total isn't held near the observed total. That's standard for a cluster bootstrap and I'd leave it — worth a one-line note in the docstring only because the standalone stratum immediately above *does* preserve its n exactly, and the asymmetry looks like an oversight rather than a choice.

`test_vehicle_mse.py:79` exercises this with `n_boot=60` and `terms=[]`, which is the most stable configuration — the independence model has 4 parameters and won't go singular. The failure mode above needs the sparse-interaction models to surface, so current coverage wouldn't catch it.

---

scripts/phase3_benchmark/vehicle_mse.py
```diff
@@ -0,0 +1,182 @@
...
+    parser.add_argument("--out", type=Path, default=None)
+    args = parser.parse_args(argv)
+
+    coded = pd.read_parquet(args.coded)
+    described = pd.read_parquet(args.described)
+    children = json.loads(args.children.read_text())
+    a_keys = {coded_key(o, i) for o, i in zip(coded["order_piid"], coded["idv_piid"])}
+    gen_col = "gen_id" if "gen_id" in described.columns else "generated_internal_id"
+    b_keys = {k for k in (described_key(g) for g in described[gen_col]) if k}
+    c_keys = {(_nk(k["order"]), _nk(k["idv"])) for k in children if k.get("order")}
```
hollomancer, line 152: Three key constructors, two of which normalize a missing parent to `"NONE"` and this one doesn't — so a children record with a blank `idv` becomes a phantom task order.

```python
coded_key:    parent = _nk(idv_piid) if str(idv_piid).strip() not in ("", "None") else "NONE"
described_key: return (_nk(...), "NONE" if parent == "-NONE-" else _nk(parent))
c_keys:       {(_nk(k["order"]), _nk(k["idv"])) for k in children if k.get("order")}
```

The guard is `if k.get("order")` — it checks the order, never the idv. `_nk(None)` and `_nk("")` both return `""`, so a child with a missing or blank `idv` yields a key with parent `""`.

That propagates into the two places the parent field is load-bearing:

- `build_units` sets `standalone = key[1] == "NONE"`, so a `""` parent is **not** standalone. The unit joins the task-order stratum — the one fit with the three-list log-linear model — despite having no parent vehicle, which is precisely the structural-zero violation the module docstring says it exists to prevent.
- `block_bootstrap_ci` does `groupby("parent")`, so all such units collapse into a single `""` cluster and are resampled as one block, distorting the cluster structure the CI depends on.

There's a second-order version too: a unit reachable only through `c_keys` with a blank idv can never match the corresponding `coded_key`/`described_key` entry for the same order, because those would carry `"NONE"`. So it appears as a C-only unit that "wasn't" in lists A or B, inflating the `C` cell (271 in the committed result) with rows that are really key mismatches.

`vehicle149_children.json` is a committed frozen cache, so today's numbers are presumably fine — but that's the argument for the guard, not against it: this file is meant to be regenerated (the July round asked for exactly that), and a silent stratum misassignment is not something the output would reveal.

Cheapest fix is to route all three through one constructor rather than duplicating the convention twice and omitting it once:

```python
def _unit_key(order: object, parent: object) -> tuple[str, str]:
    normalized = _nk(parent)
    return (_nk(order), normalized or "NONE")
```

`coded_key` and the `c_keys` comprehension both become calls to it; `described_key` keeps its `-NONE-` sentinel handling but ends in the same normalization. If a blank idv in the children cache should instead be an error — plausible, since a child of a self-declared vehicle by definition has a parent — then raise, and the frozen cache proves it never happens.

---

hollomancer (commented):
## Review of #458 at `254afae`

The eight July threads are all resolved and the fixes are real, not cosmetic. `vehicle_mse.py` is the clearest evidence: the structural zero is respected by fitting standalone contracts as a separate AB stratum and summing dark cells rather than pooling into one unidentified ABC model; the CI is a block bootstrap clustered on parent vehicle; `code_miss_rate` is computed per model instead of splicing one model's endpoint onto another's range; and the docstring says plainly that these are model-scenario estimates with no lower-bound semantics and that list-C precision is unadjudicated. That's the substance of four separate blocking comments, addressed in the code rather than in prose.

**Copilot's two threads** — both agreed, both left open, both one-line-ish:

- `_sha` reading whole parquets into memory. The repo has two streaming implementations to copy, and output is byte-identical so the committed hashes stay valid.
- Absolute `/Users/...` paths in `vehicle_mse_result.json`. Worse than non-portable: the paths are dict *keys*, so nobody else's run produces a comparable record — which defeats the one thing this artifact exists to do.

**Four new findings, inline:**

1. **`ab_bc_block_bootstrap_ci` is a single-model interval sitting at the top level**, where it reads as analysis-wide. Three of the eight models' `dark_total` values (382, 395, 499) fall *below* its lower endpoint of 850. This is the July "don't combine one model's endpoint with another's range" thread reappearing one level up, in the JSON layout, after being fixed correctly in the code.
2. **`["AB","AC","BC"]` is the saturated model** — 7 parameters, 7 cells, zero residual df — so its dark cell is functional-form extrapolation only, and its AIC is within 1.3 of the best model, so AIC can't rule it out. Suggest emitting `residual_df` per model at minimum.
3. **Non-converging bootstrap replicates are dropped silently** and the surviving count never reaches the output, so the coverage claim on `[850, 2485]` is unverifiable. Failures concentrate where sparse cells (`B`=14, `BC`=22) resample to zero — the same tail that drives the estimate. Separately, IRLS can exhaust 300 iterations and return a non-converged `beta` as if it converged.
4. **`c_keys` doesn't normalize a blank parent to `"NONE"`** while `coded_key` and `described_key` both do. A children record with a missing `idv` gets parent `""`, is classified as a task order rather than standalone — violating the very structural zero this module was rewritten to respect — and forms its own bootstrap cluster.

None of the four changes what's already in `provisional-results.json` as provisional; (1) and (2) are about how the committed artifact will be read, (3) and (4) are about what happens when someone regenerates it, which the PR explicitly anticipates ("regenerate the manifested inputs, rerun all estimators").

**Mergeability:** `mergeable_state` is clean and CI is green. The blocking question is whether `vehicle_mse_result.json` ships as-is — with laptop-absolute keys and a top-level interval — since that file is the artifact answering the July "green CI does not reproduce these figures" thread. I'd want (1) and the path keys fixed before merge; (2), (3), (4) are follow-ups.

---

PR #464, opened 2026-07-25
Address procurement report review feedback

hollomancer:
## Summary

Addresses the unresolved review feedback on #450.

- renders deterministic public-field evidence from matching UEIs, organizations, lineage phrases, technical terms, NAICS codes, and PSCs
- replaces reader-facing score jargon with verifiable facts and plain-language ranking
- explains why each award appears in the packet and prioritizes time-sensitive recorded end dates
- adds self-contained packet definitions and uses the exact “derives from, extends, or completes” Phase III test
- produces semantic, safely rendered HTML instead of escaped Markdown in a `<pre>` block
- hardens missing confidence, null routing, nullable text, and placeholder identifier handling
- corrects the monthly candidate-builder description and regenerates the deterministic example packet

## Why

Production candidate rows do not contain the hand-authored annotations used by the example packet, so the award-to-solicitation comparison could collapse to boilerplate. The HTML artifact also exposed raw Markdown, while score labels described model internals rather than facts a procurement-center representative could validate.

## Impact

The representative-facing packet is now self-contained and evidence-first without requiring AI-generated connective text. Detailed numeric scores remain available in the audit CSVs, while the packet shows the underlying public evidence and preserves the human-validation gate.

## Validation

- 99 relevant unit tests passed
- Ruff lint passed across core packages and tests
- Ruff format check passed for 726 files
- mypy passed for `sbir_etl`, `sbir-analytics`, `sbir-ml`, and all changed typed modules
- `git diff --check` passed

## Stacking

This draft targets `agent/monthly-procurement-transition-report` so it can be merged into PR #450 without duplicating the feature branch diff against `main`.


hollomancer:
Consolidated into #466 (single PR against `main` carrying all of #450 + #464 + the awardee-first reshape). Closing to avoid a lingering stack — please review #466.

PR #466, opened 2026-07-29
feat(reports): monthly procurement (PCR) packet — awardee-first, evidence-first

hollomancer:
## Summary

Single consolidated PR for the monthly procurement-transition ("PCR") reporting feature. Supersedes and folds in **#450** (base pipeline) and **#464** (evidence-first review feedback) so the whole feature lands in one PR against `main` rather than a stack.

Adds an **awardee-first** reshape on top of that work: the packet a Procurement Center Representative receives now groups every lead under the awardee it belongs to, and leads with a plain-language bottom line.

## What this PR contains

**From #450 — monthly procurement transition pipeline**
- SAM.gov Opportunities model, extractor, and download script
- Directed + competitive follow-on opportunity pairing for Phase III candidate surfacing
- Monthly cohort enrichment, candidate generation, report rendering, and a scheduled GitHub Actions workflow

**From #464 — evidence-first packets**
- Deterministic public-field evidence (matching UEIs, orgs, lineage phrases, shared terms, NAICS/PSC) instead of model-internal score jargon
- Semantic, safely-rendered HTML; hardened null/placeholder handling

**New in this PR — awardee-first + BLUF**
- Each awardee is one section listing **every** open procurement it is relevant to: possible direct-award paths first, then competitive follow-ons, each ordered by soonest response deadline
- Below-threshold matches kept under the awardee ("weaker — needs more evidence"), not dropped; awardees with no match still listed
- Awardee funded scope stated once in the header instead of repeated per lead
- Plain-language **Bottom line** header (awardee count, relevant procurements, most-urgent deadline, hold count) replaces the how-to-read glossary
- New pure `group_candidates_by_awardee` helper; AI-summary budget allocated by score up front (decoupled from the new deadline render order)

## Scope / safety

Rendering-only change on top of the existing pipeline — **no** scoring, pairing, or `phase_iii_candidates.parquet` schema changes. The human-validation gate and the "derives from, extends, or completes" statutory test are preserved.

## Validation

- `pytest tests/unit/reporting/ tests/unit/phase_iii_candidates/` — 74 passed (incl. new grouping-order, watchlist-demotion, and zero-match tests, plus the regenerated golden example)
- `ruff check` + `ruff format --check` clean on changed files
- `mypy` clean on the changed renderer module

Design doc: `docs/superpowers/specs/2026-07-29-awardee-first-procurement-packet-design.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #442, opened 2026-07-15
Epic: external evidence and source adapters


hollomancer:
## Objective

Standardize lifecycle mechanics shared by public evidence and reference-data sources without collapsing their domain-specific semantics.

## Donor work

- PR #401 — NIH RePORTER enrichment specification.
- PR #403 — restricted-entity list ingestion and screening specification.
- PR #405 — national UCC-1 extractor/registry specification.
- PR #431 — concrete state-rate refresh implementation and validation behavior.
- PR #371 — paused M&A discovery/search/evidence concepts.

## Shared contract

A source adapter should provide:

- `fetch(request, cursor) -> raw page`
- `normalize(raw page) -> typed records`
- `validate(records) -> quality result`
- checkpoint/resume and refresh-state behavior
- retry, rate-limit, pagination, and cache policy
- source/version/hash/citation provenance

## Keep domain-specific

- NIH activity-code and project semantics.
- Restricted-entity exact/fuzzy screening and list-specific identifiers.
- State-specific UCC portal/session behavior.
- State-tax interpretation and carry-forward rules.
- M&A verification, extraction, and confidence scoring.

## Exit criteria

New sources reuse tested lifecycle/provenance mechanics, retain their own typed domain adapters, and can participate in scheduled refreshes without copying request loops or inventing new cache formats.

PR #467, opened 2026-07-31
spec(phase3): notice-corpus freeze + fusion-ranker coefficients

hollomancer:
## Summary

Spec only — no implementation. Scopes the "freeze the notice corpus, fit the fusion" work that unblocks the transition ranker's full learned fusion (AUC 0.844 [0.800, 0.886]) for monthly procurement-packet candidate selection.

- **Why:** the 0.844 result's training substrate (~273 recovered notices incl. J&A text) and its recovery pipeline were never committed; the ranker findings parked productionization under #442, which is a generic, unstarted adapter-framework epic. The coupling is soft — a one-time provenance-stamped dataset freeze (precedent: `phase3_match_benchmark_pairs.parquet`) unblocks coefficient fitting without the framework.
- **What:** recovery scripts committed + re-run (GSA `falextracts` + FPDS joins), corpus frozen with manifest + frame hash, fusion refit via the already-ported GroupKFold-by-firm harness (must land within the published CI), coefficients frozen as a hash-validated load-only artifact, DIRECTED/FOLLOWON integration with the RETROSPECTIVE ≥0.85 gate bit-identical, precision@K hand-audit sample emitted.
- **Relationship to #466:** T7 (packet integration) rebases onto the merged #466 branch, which carries the ported harness (`transition_ranker` core, `text_similarity`, `ranking_features`).
- **Relationship to #442:** narrow one-time freeze here; scheduled re-recovery remains #442's scope.

## Prerequisites (implementation-time)

AWS CLI access to the GSA `falextracts` archive; FPDS ATOM access; optional `SAM_GOV_API_KEY`.


packages/sbir-ml/sbir_ml/transition/detection/fusion_scoring.py
```diff
@@ -0,0 +1,82 @@
...
+
+    n = len(award_texts)
+    if not (len(target_texts) == len(naics_codes) == len(notice_types) == n):
+        raise ValueError("all input sequences must be the same length")
+    if n == 0:
+        return []
+
+    model = load_fusion_coefficients(coefficients_path)
+    word = paired_tfidf_cosine(award_texts, target_texts, analyzer="word")
+    char = paired_tfidf_cosine(award_texts, target_texts, analyzer="char_wb")
```
hollomancer: **Train/serve feature-distribution mismatch — the biggest correctness risk in the PR.**

The frozen `scaler_mean`/`scaler_scale` were calibrated on `refit_fusion.build_features`, which differs from this path in two ways that both shift the feature distribution:

1. **Different idf corpus.** Training fitted TF-IDF over the 828-row J&A notice corpus; here `paired_tfidf_cosine` refits idf over whatever the monthly run happens to contain (SAM opportunity descriptions, a different length and vocabulary regime). Cosine magnitudes are idf-dependent, so `tfidf_word ≈ 0.029 ± 0.039` (the frozen scaler) is not the distribution this path produces.
2. **No identity scrub.** `build_features` runs `_scrub_identity` on the notice side before computing similarity; this path scores raw target text. Any firm name or PIID present in the opportunity description inflates both cosines relative to training.

Because `tfidf_word` carries **+2.21** and `tfidf_char` **−0.71**, the fused score is not monotone in either feature alone — it depends on the *ratio* of the two standardized values. A scale shift in one and not the other genuinely reorders candidates, so this isn't a benign calibration offset.

Worth either (a) scrubbing identity here to match training, and documenting that only within-run ordering is meaningful (never the absolute probability), or (b) measuring the production feature distribution against `scaler_mean`/`scaler_scale` on one real run and recording it — the spec's own §6 "scale-shift measurement before any threshold change" seems to point at exactly this.

---

packages/sbir-ml/sbir_ml/transition/detection/fusion_scoring.py
```diff
@@ -0,0 +1,82 @@
...
+    scores: list[float] = []
+    for i in range(n):
+        # ``after_first`` and ``id_cited`` carry zero frozen weight — zero here.
+        row: dict[str, float] = {
+            "tfidf_word": float(word[i]),
+            "tfidf_char": float(char[i]),
+            "after_first": 0.0,
```
hollomancer: Two things here that are safe today but silently fragile:

- `after_first` is hardcoded **0.0** while training held it constant at **1.0** (`refit_fusion.build_features` line ~118). It's inert only because the frozen coefficient is exactly `0.0`. The moment a corpus with varying `after_first` is refit, this becomes a wrong-value bug with no test to catch it — the loader validates the *corpus hash*, not that callers supply features on the same scale the fit saw. Consider asserting `coefficient == 0.0` for any feature supplied as a placeholder, or supplying `scaler_mean[i]` (the neutral value) instead of a literal.
- `NOTICE_TYPE_ORDINAL` here is a second, lowercased copy of the table in `scripts/phase3_benchmark/transition_ranker.py`. The values agree today; nothing keeps them agreeing. One table with a normalizing lookup would remove the drift class entirely.

Related: with `after_first`, `id_cited` fixed at 0 and `naics_len` effectively constant at 6 for SAM notices, the deployed score reduces to a monotone function of `2.21·z(tfidf_word) − 0.71·z(tfidf_char)` plus a `notice_type` term. Only two of six features actually vary in production. That's fine as a ranker, but it's worth saying so in the module docstring — right now the doc reads as if the full six-feature fusion is doing the work.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -660,6 +660,53 @@ def _procurement_key(row: pd.Series) -> str | None:
...
+    notice_types = [
+        _first_value(row.get("opportunity_notice_type"), row.get("target_notice_type"))
+        for _, row in rows.iterrows()
+    ]
+    try:
+        scores = score_pairs_with_fusion(award_texts, target_texts, naics, notice_types)
+    except Exception:
+        scores = [0.0] * len(rows)
+    rows = rows.copy()
+    rows["fusion_score"] = scores
```
hollomancer: Both `except Exception` paths degrade to `fusion_score = 0.0` **silently**. Since `fusion_score` is now the *primary* sort key, an all-zero fallback doesn't just lose the ranking — it silently reverts the packet to deadline order with no signal in the output or the logs that the ranker never ran. The packet would look completely normal.

At minimum log a warning (the repo has `AlertCollector` for exactly this), and ideally surface it in the report header so a consumer can tell a ranked packet from an unranked one. Compare line 1184, where the equivalent broad catch is explicitly scoped to an *optional hook*; this one is catching a core scoring dependency.

Two smaller things in the same function:
- The two early-return paths mutate the caller's frame in place (`rows["fusion_score"] = 0.0; return rows`) while the success path does `rows = rows.copy()` first. Copy on all three paths for consistency.
- Four separate `rows.iterrows()` passes to build four aligned lists; one pass building all four (or vectorized `.get()` + `combine_first`) would be both faster and easier to keep row-aligned.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -690,9 +738,12 @@ def group_candidates_by_awardee(rows: pd.DataFrame, awards: pd.DataFrame) -> lis
 
     def _entries(mask: pd.Series, seen: set[str]) -> list[pd.Series]:
         subset = rows.loc[mask]
+        # Rank a firm's procurements by the frozen fusion ranker (its validated
+        # per-firm-lead use), most promising first; soonest deadline breaks ties.
         ordered = sorted(
             (row for _, row in subset.iterrows()),
             key=lambda row: (
+                -float(row.get("fusion_score") or 0.0),
```
hollomancer, line 778: **This is a product behavior change that no test on the branch covers.** `git grep fusion_score -- tests/` returns nothing across the whole branch — the packet ordering flip is entirely untested, in a PR that otherwise has good test density (`test_fusion_model`, `test_fusion_scoring`, `test_refit_fusion`, `test_award_grain` are all solid).

Two substantive points about the change itself:

1. **The docstring at line 717 is now wrong.** It still promises "each sorted by soonest response deadline," which was the contract this packet was designed around — a representative works the packet top-down against closing dates. Deadline is now demoted below a continuous score.
2. **The tie-break comment is effectively dead.** `fusion_score` is a float from a logistic; exact ties essentially never occur, so `_deadline_sort_key` never runs. A procurement closing in three days can now sit below one closing in six months on a 0.02 score difference. If that's intended, the docstring should say so plainly; if not, bucketing the score (e.g. quartile, then deadline) would preserve time-sensitivity while still surfacing the strongest matches.

Also worth confirming in a test: `watchlist` ordering changed too, not just `directed`/`competitive`. The PR body's "RETROSPECTIVE bit-identical" claim looks right — RETROSPECTIVE doesn't flow through `_entries` — but the assertion should exist rather than be inferred.

---

scripts/phase3_benchmark/refit_fusion.py
```diff
@@ -0,0 +1,184 @@
...
+        metrics = evaluate(matrix, labels, groups, owners)
+        results.append({"stage": name, "features": list(columns), **metrics})
+    return results
+
+
+def freeze_coefficients(corpus: pd.DataFrame, columns: tuple[str, ...], frame_hash: str) -> dict:
+    features = build_features(corpus).loc[:, list(columns)].to_numpy(dtype=float)
+    labels = corpus["label"].to_numpy()
+    scaler = StandardScaler().fit(features)
+    model = LogisticRegression(max_iter=1000).fit(scaler.transform(features), labels)
```
hollomancer, line 119: `freeze_coefficients` fits the scaler and the logistic on **100% of the corpus** — no holdout, no GroupKFold. That's a defensible choice for freezing a deployment artifact (you want all the data), but it has a consequence downstream that the findings doc doesn't flag:

`emit_audit_sample.build_audit` scores the same corpus with these coefficients, so the reported **precision@1 = 0.681 is in-sample**. The AUC 0.847 [0.792, 0.898] *is* cross-validated and honestly reported; precision@1 is not, and the two are presented side by side in `findings.md` as if they carry equal weight. Given the doc's own admirable skepticism about the raw 0.920, this deserves the same treatment — either label it in-sample, or compute it from the GroupKFold out-of-fold scores (`evaluate` already produces `top1`/`top3`; the ladder's final-stage `top1 = 0.674` is the honest held-out analogue and is right there in `refit_ladder.json`).

Minor in the same file: `final_auc = float(str(final["auc"]))` round-trips a float through `str` — `float(final["auc"])` does the same thing.

---

specs/phase3-notice-corpus-fusion/findings.md
```diff
@@ -0,0 +1,105 @@
...
+"max abstract↔notice") rather than the single longest.
+
+## Artifacts
+
+- Scripts: `scripts/phase3_benchmark/{notice_matching,make_join_seed,
+  pull_gsa_archive,build_notice_corpus,transition_ranker,refit_fusion}.py`
+- Corpus (local, `/data/*` gitignored): `data/derived/phase3_notice_corpus.parquet`
+  + manifest (frame hash `6388d446…`); regenerable from the committed scripts.
+- Committed: `corpus.manifest.json`, `refit_ladder.json`, this findings doc.
+- **No coefficients frozen** — the CI gate held.
```
hollomancer: This "## Artifacts" block sits under the `---` "Superseded — firm-grain attempt" divider but reads as the document's current artifact inventory, and it now contradicts the header:

- **"No coefficients frozen — the CI gate held."** Coefficients *are* frozen and committed (`fusion_coefficients.json`).
- Frame hash `6388d446…` is the superseded firm-grain corpus; the frozen artifact and `corpus.manifest.json` both carry `4c4064f0…`.
- The script list omits `award_grain.py`, `recover_award_grain.py`, and `emit_audit_sample.py` — i.e. exactly the pipeline that produced the result the doc leads with.

Either move this block above the `---` divider so it's unambiguously scoped to the superseded attempt, or replace it with the current artifact set. As written, someone grepping for "were coefficients frozen?" gets the wrong answer from the last line of the file.

Separately, line 29: "94/138 firms" — 138 is the positive/candidate-set count; the corpus has **101** firms (`corpus.manifest.json`). `emit_audit_sample` prints `audit["owner"].nunique()` under the label `firms`, which is where the mislabel originates. Worth fixing in both places since the firm count is the thing that governs the GroupKFold CI width.

---

specs/phase3-notice-corpus-fusion/requirements.md
```diff
@@ -0,0 +1,116 @@
+# Phase III Notice Corpus & Fusion Ranker — Requirements
+
+> **Status:** Draft spec. No implementation.
```
hollomancer: `> **Status:** Draft spec. No implementation.` — stale. T1–T7 are all marked DONE in `tasks.md`, coefficients are frozen, and the packet consumes them. The PR description carries the same stale framing ("Spec only — no implementation"), which materially undersells what a reviewer needs to look at: this branch ships a production scoring path and changes packet ordering.

Also worth reconciling here: the "Done when" block and §2 specify the corpus carries "one row per (firm award, notice) with award text, notice text, notice type, dates, identifiers, label, and **stratum**", and §3 requires **two label channels** (FPDS-coded `SR3`/`ST3` primary + description-labeled replication). What actually shipped is a single `citation` channel with no stratum column (`corpus.manifest.json`: `"label_channels": {"citation": 138}`). The award-grain pivot is well-argued in `findings.md` and looks like the right call, but it superseded a requirement rather than satisfied it — the requirements doc should record that, otherwise the independent-replication check silently disappears from the record.

---

hollomancer (commented):
## Review: notice-corpus freeze + fusion-ranker coefficients

Reviewed the full 26-file diff against the merge base. CI is green across all 21 checks.

**The research here is strong, and the intellectual honesty is the best part of it.** The firm-grain attempt failing at 0.699/0.784, the diagnosis that the study's own award-level-beats-firm-level finding was the lever, the award-grain pivot via cited PIIDs, and then *refusing to accept the too-good raw 0.920* and scrubbing identity tokens down to 0.847 — that's the sequence you want and it's rare to see it written down rather than quietly cleaned up. The CI gate being two-sided (refusing to freeze above 0.886 as well as below 0.800) is a genuinely good design decision. Grouping by owner firm in GroupKFold is right, the hard-negative tiering (same office → same agency → any) is right, and the hash-validated load-only loader is exactly the right shape for a frozen artifact.

My concerns are concentrated on the last mile — **T7, the production integration** — which is where the diff is thinnest and where the framing hasn't caught up with what shipped.

### Blocking-ish

1. **Train/serve feature mismatch** (`fusion_scoring.py`). The frozen scaler was calibrated on identity-scrubbed text with idf fitted over the 828-row J&A corpus; production scores raw text with idf refitted per monthly run. With `tfidf_word` at +2.21 and `tfidf_char` at −0.71, the score depends on the *ratio* of the two standardized values, so this reorders candidates rather than just offsetting them. Inline comment has detail.
2. **The packet ordering change is untested.** `git grep fusion_score -- tests/` is empty across the branch. `fusion_score` is now the primary sort key for `directed`/`competitive`/`watchlist`, the `group_candidates_by_awardee` docstring still promises deadline ordering, and the "soonest deadline breaks ties" comment describes a branch that essentially never executes against a continuous score.
3. **Silent scoring failure.** Both `except Exception` paths in `_attach_fusion_scores` fall back to all-zero scores, which silently reverts the packet to deadline order with no signal anywhere that the ranker didn't run.

### Worth fixing before merge

4. **precision@1 = 0.681 is in-sample** — `freeze_coefficients` fits on 100% of the corpus and `emit_audit_sample` scores the same rows. The out-of-fold analogue (`top1 = 0.674`) is already sitting in `refit_ladder.json`. Given how carefully the AUC is caveated, this number should get the same treatment.
5. **`findings.md` contradicts itself.** The trailing Artifacts block still says "No coefficients frozen — the CI gate held" and cites the superseded `6388d446…` hash. It's under the superseded divider but reads as the current inventory.
6. **Stale status framing.** `requirements.md` says "Draft spec. No implementation" and the PR description says "Spec only — no implementation." Both understate the review surface. Relatedly, the two-label-channel requirement (§3) and the `stratum` column were superseded by the award-grain pivot rather than met — a good call, but it should be recorded rather than left as an unmet SHALL.

### Minor

- `after_first` hardcoded `0.0` in production vs `1.0` in training (inert only because the coefficient is exactly zero, with no test pinning that).
- `NOTICE_TYPE_ORDINAL` duplicated across `fusion_scoring.py` and `transition_ranker.py`; `_PIID_RE` duplicated across `award_grain.py` and `refit_fusion.py`. The second pair matters more than it looks: if the scrub regex ever diverges from the extraction regex, leakage stops being fully scrubbed and the 0.847 quietly drifts back toward 0.920.
- `build_features` does `np.diagonal(award_similarity(...))`, materializing an N×N cosine matrix to read N values. Fine at 828 rows, quadratic if the corpus grows — `paired_tfidf_cosine` already solves this in `sbir-ml` and could be reused.
- `emit_audit_sample.main` divides by `firms` with no guard; empty corpus → `ZeroDivisionError`.
- `urlopen` in `pull_gsa_archive._default_opener` has no timeout — a stalled connection hangs the pull indefinitely.
- `float(str(final["auc"]))` in `refit_fusion.main`.
- `fusion_coefficients.json`, `refit_ladder.json`, and `corpus.manifest.json` all lack trailing newlines.

### One methodological note, not a blocker

A notice that is firm A's positive is reused as firm B's negative. Since GroupKFold splits on the owner firm, the same notice *text* can appear in both train and test folds under different groups. This is standard practice for hard-negative retrieval setups and I don't think it invalidates anything, but with only 101 firms it's worth a sentence in `findings.md` acknowledging it as a known optimism source alongside the leakage scrub.

---

hollomancer:
## Re-review — all seven threads addressed in `b92eb85`

Every open thread was checked against the branch head before changing anything; all seven were accurate. Dispositions:

**1. Train/serve feature-distribution mismatch** — confirmed and fixed. `build_features` scrubbed identity; the production path did not. `score_pairs_with_fusion` now takes `firm_names` and applies the same scrub, and `core.py` supplies the `company` column. The idf-refit half cannot be fixed from this side, so it is documented instead: the module docstring now states plainly that only within-run ordering is meaningful and the absolute probability is not.

**2. `after_first` placeholder / duplicated ordinal table** — confirmed. It was hardcoded `0.0` while training held it at `1.0`, inert only because the frozen coefficient happens to be exactly zero. Placeholders are now supplied at their training-mean value (`scaler_mean[i]`, which standardizes to exactly zero), and `_placeholder_values` raises if either placeholder's frozen weight is ever non-zero — so a future refit fails loudly instead of scoring on a stale assumption. The docstring now says outright that only three of six features really vary here.

**3. Silent `except Exception` → `0.0`** — confirmed, and this was the sharpest one: with `fusion_score` as the primary sort key, the fallback reverted the packet to deadline order with nothing in the output or logs to show it. Both paths now log a warning and set a `fusion_ranked` flag the report header can surface. Both early paths copy before mutating, and the four `iterrows()` passes are one.

**4. Untested ordering flip + wrong docstring** — confirmed; `git grep fusion_score -- tests/` was empty. Three tests added: score outranks a sooner deadline, `watchlist` ordering changed too (it had only been inferred), and the fallback both preserves deadline order and warns. The `group_candidates_by_awardee` docstring no longer promises deadline order and now says explicitly that ranking by a continuous score demotes time-sensitivity.

On your bucketing suggestion — I left the ordering as pure score-then-deadline rather than quartile-then-deadline. That's a product call about whether the packet is a strongest-match queue or a calendar, and it should be yours, not mine. The docstring now states the current behavior clearly enough to decide against.

**5. In-sample precision@1** — confirmed. `findings.md` now labels 0.681 as in-sample and points at the out-of-fold `top1 = 0.674` from `refit_ladder.json` as the number to quote. `float(str(...))` collapsed to `float(...)`.

**6. `findings.md` Artifacts block** — confirmed on all three counts. Added an "Artifacts as shipped" inventory above the divider (frozen coefficients, `4c4064f0…`, the full script list including `award_grain`/`recover_award_grain`/`emit_audit_sample`); the old block is retitled and scoped to the superseded attempt, keeping `6388d446…` where it belongs. The `94/138 firms` mislabel is fixed at its source too — `emit_audit_sample` was printing `owner.nunique()` (the award grain, 138) under the label `firms` (101).

**7. Stale "Draft spec. No implementation."** — confirmed. The status block now says what the branch actually ships, including that packet ordering changes. I also recorded the supersession you flagged: requirements 2 (stratum) and 3 (two label channels) were superseded by the award-grain pivot, not satisfied, and the cost is stated — **the independent-replication check on the label is genuinely gone**, so the 0.847 rests on a single labeling rule with no second-channel corroboration.

The PR body still reads "Spec only — no implementation" and undersells the review surface; worth editing, but I left the description alone.

**Verification:** `807 passed` across `tests/unit/{reporting,transition,scripts}`; ruff check + format clean on the CI scope; mypy clean on `sbir_etl` (201 files) and `packages/sbir-ml/sbir_ml` (41 files).

**Mergeability:** this is the only merge-ready PR of the four open. One thing to know before merging: it adds `scripts/phase3_benchmark/transition_ranker.py`, which #458 also adds with a small diff — whichever lands second will conflict there.

---

PR #469, opened 2026-08-01
fix(config): make environment profiles authoritative

hollomancer:
## Summary

- normalize `development`/`dev` and `production`/`prod` through one profile resolver
- normalize custom profile names case-insensitively
- keep `SBIR_ETL__PIPELINE__ENVIRONMENT` authoritative, with `SBIR_ETL_ENV` as a lower-precedence compatibility fallback
- ensure explicit selections control both the loaded profile and canonical runtime metadata
- keep local pytest and CI on the same `test` profile
- remove ignored production path keys, a dead USAspending S3 path, a stale production version, and an unexpanded Neo4j database placeholder
- reject unknown `PathsConfig` keys instead of silently discarding them
- propagate the canonical selector through GitHub Actions and Docker/Server Compose entry points

## Root cause

The loader selected `config/<environment>.yaml` literally, while callers used long names (`development`, `production`) and the repository files use short names (`dev.yaml`, `prod.yaml`). Separate production detection recognized both spellings, so environment-dependent defaults and file loading could disagree. Production also carried path names outside the typed schema; Pydantic silently ignored them.

## Impact

Development and production entry points now load the profiles they advertise. Short and long aliases produce the same canonical metadata, explicit parameters retain precedence over process environment, custom profile lookup is case-insensitive, and invalid path keys fail during validation. Activating `prod.yaml` no longer introduces a literal `${NEO4J_DATABASE:-neo4j}` database name or stale `1.0.0` pipeline metadata.

The data-refresh workflow retains its original scope: only the USPTO step explicitly selects the production profile.

## Validation

- full unit suite: **4,839 passed, 7 skipped**
- 193 focused configuration/workflow tests passed
- configuration integration suite under `SBIR_ETL__PIPELINE__ENVIRONMENT=test`: passed
- Ruff and MyPy passed
- development, CI, and server Compose validation passed
- fresh GitHub CI passed unit, integration, container, performance, S3, lint, and type checks
- `git diff --check` passed


hollomancer:
## Review

The root cause is real and the resolver is the right shape. Verified against `origin/main`:

- `usaspending_dump_s3_path` has zero readers outside `config/base.yaml` — genuinely dead, safe to drop.
- `NEO4J_DATABASE` already defaults to `"neo4j"` at `sbir_etl/config/schemas/data.py:150`, so removing the unexpanded `${NEO4J_DATABASE:-neo4j}` literal from `prod.yaml` is a straight fix with no behavior loss.
- `base.yaml` is the only config file with a `paths:` block, and every key in it is a real `PathsConfig` field — `extra="forbid"` won't fire on existing config. The keys removed from `prod.yaml` (`root`, `data_dir`, `raw_dir`, …) were already being silently discarded, so removing them changes nothing at runtime.

### Two behavior changes not called out in the description

**1. `.github/workflows/data-refresh.yml` — the scope widened from one step to the whole workflow.** Moving `SBIR_ETL__PIPELINE__ENVIRONMENT` up to workflow-level `env:` affects *every* job, not just the USPTO step that previously set it inline. Jobs that ran under the `development` default now load `prod.yaml`. That's probably the intent, but it's a production-config change to jobs this PR never mentions and it deserves a line in the description.

**2. `.github/workflows/ci.yml` — CI and local now load different profiles.** CI now resolves `test.yaml` where it previously fell through to `dev.yaml`. Local `pytest` still resolves to `dev.yaml`, so the two validate against different profiles from here on. Either note it, or set the same variable in the pytest env config so local and CI agree.

### Smaller

- **Mismatched defaults on the same source variable.** `docker-compose.server.yml` has `ENVIRONMENT: ${ENVIRONMENT:-prod}` directly above `SBIR_ETL__PIPELINE__ENVIRONMENT: ${ENVIRONMENT:-production}`; `docker-compose.yml` has the same `dev` / `development` split. Both alias to the same profile so it's harmless, but reading as a typo is a cost with no benefit — worth picking one spelling per file.
- **`docs/configuration.md` is now stale.** `_apply_env_overrides` skips `SBIR_ETL__PIPELINE__ENVIRONMENT`, which makes line 220 ("Any configuration value can be overridden at runtime with an environment variable") false. `config/README.md` was updated; this one wasn't.
- **`_normalize_environment_profile` case handling is asymmetric.** It looks up on `.lower()` but returns the original casing for unknown names, so `SBIR_ETL__PIPELINE__ENVIRONMENT=Staging` looks for `Staging.yaml` while `staging` looks for `staging.yaml`. Returning `requested.lower()` would make the fallback consistent with the alias path.

Happy to push the compose-default cleanup if you want it in this PR.

---

PR #470, opened 2026-08-01
refactor(neo4j): package migrations with sbir-graph

hollomancer:
## What

- move the existing migration framework and versions 001–007 into `sbir_graph.migrations`
- update the Neo4j client and checkout CLI to use the packaged runner
- remove Docker's separate top-level migration copy and stale tooling exclusions
- require an explicit migration rollout during normal application startup
- make migration discovery, import, and execution fail closed
- document the production backup, inspection, dry-run, and upgrade sequence
- add discovery, failure, dry-run, applied-version compatibility, and opt-in auto-migration coverage

## Why

`sbir-graph` wheels only package `sbir_graph`. The former top-level `migrations` package was therefore absent from installed distributions, so wheel-installed clients could silently skip every schema/data migration while checkout and Docker environments masked the defect.

Packaging the migrations also activates consequential data migrations that must not run silently during ordinary client construction.

## Impact and rollout

Normal `Neo4jClient` construction now uses `auto_migrate=False`. Before upgrading a production graph, operators must:

1. take and verify a backup;
2. inspect `current` and `history`;
3. run `upgrade --dry-run` (now read-only);
4. review the plan and run the explicit `upgrade` command.

Migrations **003**, **006**, and **007** are data migrations: they merge duplicate organizations, re-home legacy `:Award` nodes into `:FinancialTransaction`, and re-home legacy `:Company` nodes into `:Organization`.

Direct `sbir_graph` callers may still opt in with `Neo4jConfig(auto_migrate=True)`. Discovery, import, or statement failures then close the driver and abort client initialization. Failed migrations are not marked applied. Existing version identifiers and Cypher statements remain compatible with prior tracking.

## Checks

- 188 affected config/client/runner/migration tests passed
- full unit suite: 4,826 passed, 7 skipped
- MyPy passed across all 14 affected production sources
- source CLI passed under `python -I`
- built `sbir_graph-0.1.0` wheel contains all 12 migration package files
- isolated wheel install resolves `sbir_graph` outside the checkout and discovers exactly 001–007
- moved migration bodies were compared against `origin/main`; namespace imports and fail-closed error handling are the intentional changes


hollomancer:
## Review

The mechanics are correct. Verified on the branch:

- `MIGRATIONS_DIR` is `Path(__file__).parent / "versions"`, so discovery follows the move with no path fixup needed.
- `packages/sbir-graph/pyproject.toml` uses hatchling with `packages = ["sbir_graph"]`, which picks up the new subpackage tree — consistent with the wheel-contents check in the description.
- `Dockerfile` and `Dockerfile.full` already `COPY packages/sbir-graph/sbir_graph/ /app/sbir_graph/`, so dropping the separate `COPY migrations/` is safe.
- No stale `from migrations.` imports remain on the branch.
- Dropping the `migrations/` mypy exclude is a no-op for blocking CI, which runs `mypy sbir_etl` only.

### The rollout question

The release note understates its own consequence, and I'd want this settled before merge rather than after.

`auto_migrate` defaults to `True` in both `Neo4jClient` (`client.py:29`) and the config schema (`data.py:169`), and `Neo4jClient.__init__` calls `runner.upgrade()` during construction. So on the first client init after this ships, any wheel-installed deployment silently applies migrations **003, 006, and 007** — merging duplicate `Organization` nodes, re-homing `:Award` into `:FinancialTransaction`, and re-homing `:Company` into `:Organization` — against whatever is in that database, at client-construction time, with no dry-run and no operator prompt. "Wheel installs will now execute the already-documented migrations they previously skipped" is accurate but reads much softer than what 006 and 007 actually do.

Two changes would de-risk it without touching the fix itself:

**1. Stop swallowing migration failures.** `client.py:106` still catches every exception from `runner.upgrade()` and logs it at `warning`. That swallow is precisely why the packaging defect went unnoticed as long as it did, and it now guards a destructive path. At minimum promote it to `logger.error`; better, fail initialization when a *data* migration (003/006/007) errors, and keep the soft-fail only for schema migrations. Same shape at `runner.py:88`, where a migration that fails to import is a `logger.warning` and then just silently absent from the discovered set — a partial-apply that looks like success.

**2. Gate the first post-upgrade run.** Consider shipping with the data migrations requiring an explicit `scripts/neo4j/migrate.py upgrade`, or documenting a `--dry-run` step in the Neo4j runbook, so an operator sees the plan before 006/007 rewrite the graph.

The refactor itself is clean — this is a deployment-sequencing call, not a code objection.

---

PR #471, opened 2026-08-01
refactor(reporting): centralize analyzer lifecycle

hollomancer:
## What

- move the common analyzer sequence and missing-input report construction into `ModuleAnalyzer`
- keep each analyzer's stage, primary input, denominator, count keys, duration source, and messages explicit
- require subclasses to declare lifecycle metadata and implement both domain report hooks
- retain small domain adapters for hygiene and change calculations
- add characterization coverage for exact report composition, call order, short-circuiting, defaults, and domain-specific metric sources

## Why

The SBIR, CET, Patent, and Transition analyzers duplicated the same lifecycle and empty-report construction across roughly 300 lines. Small but important differences—especially Patent's split validation/loading sources and Transition's award denominator—were buried inside those copies.

## Impact

Production analyzer code is 147 lines smaller. Domain hygiene, changes, insight logic, thresholds, confidence bands, and output models are unchanged. Missing `None` inputs remain distinct from empty DataFrames, and log/report messages retain their existing text. Missing subclass hooks now fail at instantiation instead of mid-report, and required lifecycle metadata no longer has silent empty defaults.

Refs #441 for context only. This internal legacy-analyzer cleanup neither implements nor constrains the modular analysis/report platform and does not partially close that issue.

## Checks

- 69 analyzer tests passed
- 120 related reporting/model tests passed
- MyPy passed for all analyzer modules
- fresh GitHub CI passed unit, integration, lint, type, and S3 checks
- independent review found no contract, API, or scope blockers

### DIFF
--- a/sbir_etl/utils/reporting/analyzers/base_analyzer.py
+++ b/sbir_etl/utils/reporting/analyzers/base_analyzer.py
@@ -4,9 +4,10 @@ from abc import ABC, abstractmethod
 from datetime import datetime
 from typing import Any
 
+from loguru import logger
 from pydantic import BaseModel
 
-from sbir_etl.models.quality import ModuleReport
+from sbir_etl.models.quality import ChangesSummary, DataHygieneMetrics, ModuleReport
 
 
 class AnalysisInsight(BaseModel):
@@ -25,6 +26,16 @@ class AnalysisInsight(BaseModel):
 class ModuleAnalyzer(ABC):
     """Base class for module-specific statistical analyzers."""
 
+    stage = ""
+    primary_data_key = ""
+    total_data_key: str | None = None
+    processing_keys = ("", "", "")
+    duration_data_key: str | None = None
+    analysis_label = ""
+    start_analysis_label: str | None = None
+    missing_data_warning = ""
+    missing_data_error = ""
+
     def __init__(self, module_name: str, config: dict[str, Any] | None = None):
         """Initialize the analyzer.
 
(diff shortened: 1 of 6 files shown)


hollomancer:
## Review

Ship with nits. I diffed all four analyzers against the extracted lifecycle line by line and found no behavior change. Every mapping is faithful, including the two the description flags — Patent's split `validation_results` / `loading_results` duration source, and Transition's `awards_df` denominator via `total_data_key`. The dropped `module_data.get("taxonomy_data", {})` in the CET analyzer was a dead statement with no assignment. Missing-`None` short-circuiting still happens before `get_key_metrics`, so empty DataFrames stay distinct from absent inputs.

Three nits, none blocking:

**1. `base_analyzer.py` — the new hooks aren't abstract.** `analyze` lost `@abstractmethod`, and `_calculate_report_data_hygiene` / `_calculate_report_changes_summary` raise `NotImplementedError` instead. The class is still an `ABC`, so `@abstractmethod` is free here and moves a forgotten override from a runtime crash mid-report to an instantiation error.

**2. `base_analyzer.py` — empty-string class defaults.** `stage = ""` and `processing_keys = ("", "", "")` mean a subclass that forgets to set them produces a silently wrong report rather than failing. Given the hooks above are already required of every subclass, `stage` and `primary_data_key` may as well be too.

**3. `analysis_label` + `start_analysis_label`.** These two attributes and the `or` fallback exist solely to reproduce `"Starting patent analysis"` vs `"Patent analysis complete"` casing. It's the one place the abstraction earns less than it costs — normalizing the two log messages would drop an attribute and a branch. Understood if the characterization tests are deliberately pinning the existing strings.

---

PR #472, opened 2026-08-01
Add deterministic label-free Phase III census

hollomancer:
## What changed

- Freezes the label-free Phase III census criteria, estimand, clause order, and six-cell sensitivity grid before observing any census count.
- Adds the deterministic `phase_iii_census` Dagster asset, producing an auditable cumulative drop-off ladder and the full time-window × agency-match grid.
- Reuses the existing exact-UEI pair builder consumed by `phase_iii_candidates`; the weighted scoring path remains untouched.
- Stabilizes Phase II award grain and one-to-one SBIR.gov/federal reconciliation.
- Adds a dedicated, deterministic full-history SBIR.gov source-row materializer with exact-row fingerprints and fail-closed provenance verification.
- Adds bounded-memory, schema-verified USAspending contract extraction and checksum-bound source manifests.
- Updates the research-question answerability record.

## Why

The repository has no labeled Phase III outcomes, so hand-set scoring weights cannot be fit or validated. This adds a parallel path whose estimand is a transparent count under frozen statutory/research-backed criteria, without weights, thresholds, similarity scoring, or new ML.

## Validation

- 116 focused Phase II, source-grain, census, and shared-pairing tests pass.
- Focused Ruff formatting/lint and mypy checks pass.
- Independent rematerialization of the SBIR.gov input produced the same 219,500-row parquet SHA-256.
- Hard-constraint review found no imports or calls to `_score_pair`, `TransitionScorer`, weight constants, or transition-detection ML code.

## Draft status

The February 6, 2026 USAspending mirror extraction and final Phase II/census parquet materialization are still running. No census result has been selected or summarized; the PR will be updated with the auditable tables after source verification completes.


hollomancer:
## Review

The census core is genuinely careful work. `criteria.py` freezes the clauses cleanly, the cumulative ladder is honest about where rows die, `_summarize` dedupes by transaction before summing obligations so de-obligations don't double-count, and `validate_pair_frame` fails closed on exactly the cases where an absent column would masquerade as a real zero. On the extraction side, `_awk_prefilter_lines` does the right paranoid things: vendor values go into map files rather than the awk program, non-ASCII and embedded newlines fall back to Python, `LC_ALL=C` is pinned, awk's own counters are reconciled against the Python-side count, and `feed_errors` is actually surfaced (line 989) rather than letting a truncated feed silently shrink the input. That last one is the failure mode I went looking for and didn't find.

Three things I'd fix before this leaves draft.

### 1. The sensitivity checkpoint will predictably fail the asset — and discards the work when it does

`assets.py:283` calls `enforce_sensitivity_checkpoint(sensitivity)`, which raises when `distinct_firms` or `distinct_contracts` spans more than 3× across the six cells. The `dropoff.to_parquet(...)` write is *after* that call, so nothing lands on disk.

But the six cells are `{none, 5y, 10y} × {same_agency, same_department}`, and `none/same_department` is a strict superset of `5y/same_agency`. On real federal contracting data that ratio will almost certainly exceed 3×. As written, the expected outcome of a successful run is a failed asset with no artifacts.

If the checkpoint means "a human must look at this before anyone quotes a number," it should be a Dagster asset check, or a `MetadataValue` warning on a successful materialization — write both tables, then flag. Losing a correctly-computed drop-off ladder because the grid is wide is the wrong trade, and it makes the asset unrunnable without hand-editing on every execution.

### 2. `pair_filter_s1` got slower on the path that already exists

Previously the already-coded-Phase-III rows were dropped from `contracts` *before* the merge, inside `_prepare_contracts`. Now `build_uei_pairs` merges everything and `pair_filter_s1` applies `_is_phase_iii_already_coded` row-wise to the full pair frame (`pairing.py:281`).

The result set is identical — both paths read the same `research` / `sbir_phase` source fields, just under `target_*` names now — so this is purely a cost question. But there are now two `.apply(axis=1)` passes over a strictly larger cross-join, plus `_coalesce_fields` and `_metadata_field` doing Python-level `.map` per column over the contracts frame. On the full USAspending × Phase II universe that's a real regression on an existing path, and the draft note says materialization is still running. The coded check is two `.isin` calls — vectorizing it recovers most of the loss.

### 3. Scope

The census proper — `criteria.py`, `phase_iii_census/assets.py`, the `pairing.py` split — is roughly 900 lines and reviewable. Bundled with it: a 694→1392-line rewrite of `contract_extractor.py`, a new 570-line SBIR.gov source materializer, 328 lines of change to the existing `phase_ii.py` asset, and archive-script changes. ~7,700 additions across 34 files.

Each piece is defensible on its own — the extractor rewrite in particular fixes positional-column parsing, which is a genuine correctness issue and arguably more urgent than the census. But as one PR, nobody can give the frozen census criteria and a source-extraction rewrite equal attention. I'd land the extractor + source-materializer work as its own PR first, then rebase the census onto it.

### Smaller points

- **`assets.py:38-53` — the dagster import shim.** The `try: from dagster import ... except Exception:` fallback defines fake `asset` / `Output` / `MetadataValue` in production code. If dagster ever fails to import for an unrelated reason, this asset degrades to a plain function returning a fake `Output` instead of erroring loudly. Dagster is a hard dependency; `test_import_isolation.py` can import it directly.
- **`assets.py:57-61` — hardcoded relative output paths.** `Path("data/processed/...")` bypasses `PathsConfig` entirely. Directly at odds with #469, which is making the paths schema authoritative and now rejects unknown path keys.
- **`assets.py:66` — `FROZEN_SPEC_COMMIT` asserts something nothing checks.** The constant claims `specs/phase-iii-census/design.md` is frozen at that SHA, but no code verifies it, so it silently goes stale on the next amendment. A `sha256` of the design file, checked at materialization, would actually enforce the claim.
- **`assets.py:187` — inconsistent error type.** If `contracts_exists` is `True` in the manifest but the file has since been deleted, `sha256_file(contracts_path)` raises a bare `FileNotFoundError` rather than the `CensusInputError` every other failure in that function produces.
- **`pairing.py:47` — fragile slice.** `PAIR_COLUMNS = [*PAIR_S1_COLUMNS[:-1], ...]` silently assumes `agency_match_level` is last in `PAIR_S1_COLUMNS`. It is today; a reorder breaks it quietly.
- **Branch state.** 12 commits behind `main` and GitHub reports it unmergeable — it'll need a merge before CI results mean anything.

---

hollomancer:
Superseded by the reviewed split stack, now merged: source materialization in #474 and the Phase III census in #475. Closing this conflicting duplicate implementation in favor of those merge-tested changes.

PR #473, opened 2026-08-01
refactor(uspto): centralize identifier normalization

hollomancer:
## What

- Add one domain-specific `normalize_uspto_identifier` utility for identifiers carried by USPTO records.
- Replace the byte-for-byte-equivalent model and transformer implementations with that shared function.
- Preserve the transformer's existing private static callable and the current permissive ingestion behavior, including Unicode word characters, underscores, hyphens, numeric coercion, and punctuation-only input.
- Add contract tests for the legacy normalization edge cases.

## Why

USPTO models and the patent assignment transformer maintained the same normalization algorithm independently. Centralizing it removes a drift point while keeping field selection and domain behavior unchanged.

## Scope

This PR intentionally does **not**:

- introduce a universal identifier normalizer;
- tighten UEI, DUNS, or CAGE validation;
- change Patent or Neo4j keys;
- change extractor field-alias selection;
- overlap company-identity or schema-authority work.

## Validation

- 153 focused USPTO/model/transformer tests passed
- 33 USPTO integration tests passed
- Full unit suite: 4,826 passed, 7 skipped
- MyPy passed for all changed production modules
- Independent review found no blockers

### DIFF
--- a/sbir_etl/models/uspto_models.py
+++ b/sbir_etl/models/uspto_models.py
@@ -30,17 +30,7 @@ from pydantic import BaseModel, Field, field_validator
 # ---- Utilities ----
 # Use centralized date parsing utility
 from sbir_etl.utils.date_utils import parse_date as _parse_date
-
-
-def _normalize_identifier(val: str | None) -> str | None:
-    if val is None:
-        return None
-    v = str(val).strip()
-    if v == "":
-        return None
-    # Remove common punctuation and whitespace collapse
-    v = re.sub(r"[^\w\-]", "", v).upper()
-    return v
+from sbir_etl.utils.identifiers import normalize_uspto_identifier
 
 
 def _normalize_name(name: str | None) -> str | None:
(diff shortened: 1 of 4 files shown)


hollomancer:
## Review

No findings — approve as-is.

Verified against `origin/main`:

- Both replaced implementations were byte-equivalent to `normalize_uspto_identifier`; the model and transformer versions differed only in variable naming.
- `re` is still used in both `sbir_etl/models/uspto_models.py` (`_normalize_name`) and `sbir_etl/transformers/patent_transformer.py` (address/ZIP/conveyance parsing), so no orphaned imports.
- The legacy edge where punctuation-only input returns `""` rather than `None` is preserved and now pinned by the new parametrized test — good call testing that explicitly rather than quietly "fixing" it.
- `_normalize_identifier = staticmethod(normalize_uspto_identifier)` keeps the private callable available for any caller reaching for it.

Scope matches the stated non-goals: no UEI/DUNS/CAGE tightening, no key changes, no extractor alias changes.

---

PR #475, opened 2026-08-01
Implement an auditable label-free Phase III census

hollomancer:
## Stack

Depends on #474, which contains the schema-verified USAspending/SBIR.gov source materialization and Phase II award-grain correction. This PR targets `codex/phase-iii-source-materialization`, so its review diff is census-only.

The earlier mixed draft #472 is preserved as audit history and will be closed as superseded after this stack is linked.

## Executable Phase 0 freeze

- Commits the amended design and append-only amendment log before the implementation commit.
- Hashes the raw bytes of both files at materialization and fails closed on a missing or mismatched file before the census loads contracts, pairs, counts, or writes.
- Records the spec/log paths, SHA-256 digests, and revision in Dagster metadata.
- Discloses each amendment's reason, criteria impact, and which source facts—but no census results—were visible at approval.

Frozen digests:

- design: `b1cce1e75af840f13223ecfc543b6d801e63cc3329f65f7c0a5b9037c0e9fc48`
- amendment log: `f9cdf80cf94c75e2e4475ca6f736e2e0fab27a8e8d19650a7841b11b17801605`

## Phase 1 implementation

- Adds one shared, exact-UEI, transaction-grain pair boundary used by both legacy S1 and the label-free census.
- Preserves the legacy scoring path's award-grain coded exclusion, latest-transaction selection, agency gate, and public schema.
- Adds the deterministic cumulative drop-off ladder and all six pre-registered window × agency cells.
- Writes both parquet audit tables before a blocking Dagster asset check reads them.
- Replaces the invalid global 3× span with every adjacent one-factor edge: four window edges plus three agency edges. A complete 3×2 adjacent graph has seven comparisons; none was silently omitted to force the earlier prose count of six.
- Reports pair, firm, contract, and signed-dollar effects; only an adjacent window effect on firms/contracts can fail, and only when it exceeds both 3× and the largest adjacent core-clause effect.

## Materialization and control dependency

No production census, drop-off, or sensitivity table has been materialized or quoted.

Phase 2 is paused at the required stop-and-ask checkpoint: the full-history SBIR.gov record contains a material identifier-free population, so an exact UEI/DUNS anti-join cannot certify that a SAM entity has never received an SBIR/STTR award. The census will remain unmaterialized until the control eligibility rule is approved and reliable enough to run before a count is circulated.

## Scope boundary

This PR does not modify `phase_iii_candidates/assets.py`, `TransitionScorer`, weight constants, or anything under `packages/sbir-ml/sbir_ml/transition/detection/`. The census imports/calls no scorer, weight, threshold, embedding, classifier, or similarity ranking, and it builds no second firm-contract join.

## Validation

- 316 focused census, pairing, discovery, source, extractor, Phase II, model, and script tests passed.
- 75 full Phase III candidate and extractor-integration tests passed.
- Independent quality sweep: 81 focused tests passed; Dagster discovered the asset and asset check.
- Ruff check and format pass.
- Targeted mypy passes.
- `git diff --check` passes.


hollomancer:
Post-merge validation dependency: #478 is the deliberately blocked negative-control/placebo PR. Its failing release-gate test records that the Phase 1 census tables are audit artifacts, not validated Phase III results, until matched controls and the fixed-seed placebo exist. No census result has been materialized or quoted.

PR #478, opened 2026-08-01
Add Phase III control and placebo foundation

hollomancer:
## Purpose

This draft is the visible validation dependency for the frozen, label-free Phase III census. It adds only the approved data-independent control audit and placebo mechanics; it does not present the census proxy as a validated Phase III result.

## What changed

- Replaces the dependency stub with repository-convention requirements, design, and tasks documents.
- Audits exact strict-normalized UEI and DUNS matches against a caller-supplied complete available SBIR/STTR history, retaining every exclusion reason.
- Labels passing candidates exactly `no observed exact-identifier SBIR/STTR match`; it never certifies a firm as SBIR-negative.
- Flags exact normalized-name matches only against identifier-free award rows as a worst-case reporting stress set. The flag never changes eligibility and is not described as a contamination estimate or bound.
- Records that public inputs omit employee count, with no proxy or new band, and that this is a major residual-balance limitation.
- Implements the single seed-`20260801` placebo at unique prior-award grain, propagates each shuffled end date through pair fan-out, and passes one permuted frame to both unchanged census table builders.
- Keeps the evidence-backed release sentinel deliberately failing.

## Deferred by design

The spec leaves SAM and award-history provenance, candidate-frame scope, canonical names, NAICS/state fields, first-contract year, PSC family, matching/reuse/tie-breaking, organizational identity, SMD encoding, per-firm and pseudo-index grain, artifact contracts, and release sign-off as explicit questions. None is silently answered in code.

## Scope boundary

The PR contains no SAM reader, matcher, Dagster asset, empirical control/placebo artifact, result, scorer, weight, threshold, similarity method, embedding, classifier, or ML change. It does not modify the frozen census criteria or the transition-scoring path.

## Validation

- 81 focused controls, arm-blindness, census-criteria, and shared-pairing tests pass after integration with current `main`.
- Ruff check and format pass.
- Mypy passes for the new package.
- `git diff --check` passes.

## Expected CI state

This PR remains intentionally draft/red. The release sentinel may be replaced only after real, provenance-backed negative-control and fixed-seed placebo artifacts exist. Pure fixture tests are not release evidence.

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
@@ -22,6 +22,7 @@ from ..phase_iii_census.criteria import build_dropoff_ladder, build_sensitivity_
 
 PLACEBO_SEED = 20260801
 CONTROL_STATUS_LABEL = "no observed exact-identifier SBIR/STTR match"
+UNSCREENABLE_STATUS_LABEL = "no usable identifier; not screenable"
 
 EXACT_UEI_EXCLUSION_REASON = "observed_exact_uei_sbir_sttr_match"
 EXACT_DUNS_EXCLUSION_REASON = "observed_exact_duns_sbir_sttr_match"
@@ -30,6 +31,7 @@ _CANDIDATE_IDENTIFIER_COLUMNS = frozenset({"entity_id", "uei", "duns"})
 _HISTORY_IDENTIFIER_COLUMNS = frozenset({"uei", "duns"})
 _ELIGIBILITY_AUDIT_COLUMNS = frozenset(
     {
+        "has_usable_identifier",
         "passes_exact_identifier_screen",
         "control_status_label",
     }
@@ -92,6 +94,12 @@ def _validate_history(
         )
 
 
+def _control_status_label(passed: bool, screenable: bool) -> str | None:
+    if not passed:
+        return None
+    return CONTROL_STATUS_LABEL if screenable else UNSCREENABLE_STATUS_LABEL
+
+
 def audit_exact_identifier_eligibility(
     candidate_entities: pd.DataFrame,
     complete_award_history: pd.DataFrame,
@@ -101,6 +109,14 @@ def audit_exact_identifier_eligibility(
     Passing this screen means only that no exact usable identifier was observed in
     the supplied history. It does not certify that an entity never received an
     SBIR/STTR award.
+
+    Passing is three-valued in substance and the output says so rather than leaving
+    it to be reconstructed from the normalized identifier columns. An entity that
+    was screened and matched nothing carries `has_usable_identifier` true and the
+    approved control label; an entity whose identifiers are absent or malformed
+    carries `has_usable_identifier` false and `UNSCREENABLE_STATUS_LABEL`, because
+    nothing about it was screened at all. Selecting on
+    `passes_exact_identifier_screen` alone silently mixes the two.
     """
 
     _require_columns(
@@ -127,6 +143,7 @@ def audit_exact_identifier_eligibility(
     normalized_duns = output["duns"].map(normalize_duns)
     exact_uei_match = normalized_uei.isin(history_ueis)
     exact_duns_match = normalized_duns.isin(history_duns)
+    has_usable_identifier = normalized_uei.notna() | normalized_duns.notna()
     passes = ~(exact_uei_match | exact_duns_match)
 
     reasons = [
@@ -148,20 +165,37 @@ def audit_exact_identifier_eligibility(
     output["exclusion_reason"] = pd.Series(reasons, index=output.index, dtype="string").mask(
         lambda series: series.eq("")
     )
+    output["has_usable_identifier"] = has_usable_identifier.astype(bool)
     output["passes_exact_identifier_screen"] = passes.astype(bool)
     output["control_status_label"] = pd.Series(
-        [CONTROL_STATUS_LABEL if passed else pd.NA for passed in passes],
+        [
+            _control_status_label(bool(passed), bool(screenable))
+            for passed, screenable in zip(passes, has_usable_identifier, strict=True)
+        ],
         index=output.index,
         dtype="string",
     )
     return output
 
 
-def flag_identifier_free_name_stress_set(
+def flag_identifier_unreachable_name_stress_set(
     eligibility_audit: pd.DataFrame,
     complete_award_history: pd.DataFrame,
 ) -> pd.DataFrame:
-    """Flag exact normalized names found on identifier-free award rows.
+    """Flag exact normalized names on award rows the identifier screen could not reach.
+
+    Which award rows are unreachable is a property of the *candidate*, not of the
+    history alone. A candidate with a usable identifier was compared against every
+    identifier-carrying award row, so only identifier-free award rows remain
+    unreachable for it. A candidate with no usable identifier was compared against
+    nothing, so the whole history remains unreachable for it — including award rows
+    that do carry identifiers.
+
+    Restricting the reference set to identifier-free award rows for every candidate
+    drops exactly the worst case this flag exists to surface: an unscreenable
+    candidate whose normalized name equals that of an identifier-carrying awardee.
+    That case would be reported as clean by the eligibility audit and absent here,
+    and it understates stress in the direction that flatters the control arm.
 
     The flag is for worst-case stress reporting only. It is not an eligibility
     rule, contamination estimate, upper bound, fuzzy match, or alias screen.
@@ -183,11 +217,12 @@ def flag_identifier_free_name_stress_set(
     history_duns = complete_award_history["duns"].map(normalize_duns)
     history_names = complete_award_history["company_name"].map(_normalized_company_name)
     identifier_free = history_uei.isna() & history_duns.isna()
-    reference_names = {
+    identifier_free_names = {
         normalized
         for normalized in history_names.loc[identifier_free]
         if isinstance(normalized, str)
     }
+    every_award_name = {normalized for normalized in history_names if isinstance(normalized, str)}
 
     output = eligibility_audit.copy()
     normalized_entity_name = output["entity_name"].map(_normalized_company_name)
@@ -196,8 +231,13 @@ def flag_identifier_free_name_stress_set(
         index=output.index,
         dtype="string",
     )
-    output["identifier_free_award_exact_name_match"] = (
-        normalized_entity_name.notna() & normalized_entity_name.isin(reference_names)
+    screenable = output["has_usable_identifier"].astype(bool)
+    unreachable_name_match = normalized_entity_name.isin(identifier_free_names).where(
+        screenable,
+        normalized_entity_name.isin(every_award_name),
+    )
+    output["identifier_unreachable_award_exact_name_match"] = (
+        normalized_entity_name.notna() & unreachable_name_match
     ).astype(bool)
     return output
 
--- a/packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
@@ -1,0 +1,296 @@
+"""Approved pure mechanics for Phase III controls and the fixed placebo.
+
+This module does not choose a control population, match firms, materialize an
+artifact, or change the frozen census criteria. Its inputs are already-canonical
+frames supplied by a future, separately approved orchestration layer.
+"""
+
+from __future__ import annotations
+
+from collections import Counter
+from datetime import date
+from typing import Any
+
+import numpy as np
+import pandas as pd
+
+from sbir_etl.utils.identifiers import normalize_duns, normalize_uei
+from sbir_etl.utils.text_normalization import normalize_company_name
+
+from ..phase_iii_census.criteria import build_dropoff_ladder, build_sensitivity_grid
+
+
+PLACEBO_SEED = 20260801
+CONTROL_STATUS_LABEL = "no observed exact-identifier SBIR/STTR match"
+
+EXACT_UEI_EXCLUSION_REASON = "observed_exact_uei_sbir_sttr_match"
+EXACT_DUNS_EXCLUSION_REASON = "observed_exact_duns_sbir_sttr_match"
+
+_CANDIDATE_IDENTIFIER_COLUMNS = frozenset({"entity_id", "uei", "duns"})
+_HISTORY_IDENTIFIER_COLUMNS = frozenset({"uei", "duns"})
+_ELIGIBILITY_AUDIT_COLUMNS = frozenset(
+    {
+        "passes_exact_identifier_screen",
+        "control_status_label",
+    }
+)
+_NULL_TEXT = frozenset({"", "<NA>", "NAN", "NAT", "NONE", "NULL", r"\N"})
+
+
+class NegativeControlInputError(ValueError):
+    """Raised when an input cannot support the approved audit or placebo."""
+
+
+def _require_columns(frame: pd.DataFrame, required: frozenset[str], *, label: str) -> None:
+    missing = sorted(required - set(frame.columns))
+    if missing:
+        raise NegativeControlInputError(f"{label} is missing required columns: {missing}")
+
+
+def _normalized_record_key(value: Any) -> str:
+    if value is None or value is pd.NA or value is pd.NaT:
+        return ""
+    try:
+        if bool(pd.isna(value)):
+            return ""
+    except (TypeError, ValueError):
+        pass
+    normalized = str(value).strip().upper()
+    return "" if normalized in _NULL_TEXT else normalized
+
+
+def _normalized_company_name(value: Any) -> str | None:
+    if _normalized_record_key(value) == "":
+        return None
+    normalized = normalize_company_name(str(value))
+    return normalized or None
+
+
+def _validated_entity_keys(candidate_entities: pd.DataFrame) -> pd.Series:
+    keys = candidate_entities["entity_id"].map(_normalized_record_key)
+    if keys.eq("").any():
+        raise NegativeControlInputError("Every candidate entity must have a nonblank entity_id")
+    if keys.duplicated().any():
+        raise NegativeControlInputError(
+            "Candidate entity_id values must be unique after case/whitespace normalization"
+        )
+    return keys
+
+
+def _validate_history(
+    candidate_entities: pd.DataFrame,
+    complete_award_history: pd.DataFrame,
+) -> None:
+    _require_columns(
+        complete_award_history,
+        _HISTORY_IDENTIFIER_COLUMNS,
+        label="complete SBIR/STTR award history",
+    )
+    if not candidate_entities.empty and complete_award_history.empty:
+        raise NegativeControlInputError(
+            "A nonempty candidate frame cannot be screened against empty SBIR/STTR history"
+        )
+
+
+def audit_exact_identifier_eligibility(
+    candidate_entities: pd.DataFrame,
+    complete_award_history: pd.DataFrame,
+) -> pd.DataFrame:
+    """Audit exact UEI/DUNS exclusions against supplied complete award history.
+
+    Passing this screen means only that no exact usable identifier was observed in
+    the supplied history. It does not certify that an entity never received an
+    SBIR/STTR award.
+    """
+
+    _require_columns(
+        candidate_entities,
+        _CANDIDATE_IDENTIFIER_COLUMNS,
+        label="candidate entity frame",
+    )
+    _validate_history(candidate_entities, complete_award_history)
+    _validated_entity_keys(candidate_entities)
+
+    history_ueis = {
+        normalized
+        for normalized in complete_award_history["uei"].map(normalize_uei)
+        if isinstance(normalized, str)
+    }
+    history_duns = {
+        normalized
+        for normalized in complete_award_history["duns"].map(normalize_duns)
+        if isinstance(normalized, str)
+    }
+
+    output = candidate_entities.copy()
+    normalized_uei = output["uei"].map(normalize_uei)
+    normalized_duns = output["duns"].map(normalize_duns)
+    exact_uei_match = normalized_uei.isin(history_ueis)
+    exact_duns_match = normalized_duns.isin(history_duns)
+    passes = ~(exact_uei_match | exact_duns_match)
+
+    reasons = [
+        ";".join(
+            reason
+            for matched, reason in (
+                (uei_match, EXACT_UEI_EXCLUSION_REASON),
+                (duns_match, EXACT_DUNS_EXCLUSION_REASON),
+            )
+            if matched
+        )
+        for uei_match, duns_match in zip(exact_uei_match, exact_duns_match, strict=True)
+    ]
+
+    output["normalized_uei"] = pd.Series(normalized_uei, index=output.index, dtype="string")
+    output["normalized_duns"] = pd.Series(normalized_duns, index=output.index, dtype="string")
+    output["exact_uei_sbir_sttr_match"] = exact_uei_match.astype(bool)
+    output["exact_duns_sbir_sttr_match"] = exact_duns_match.astype(bool)
+    output["exclusion_reason"] = pd.Series(reasons, index=output.index, dtype="string").mask(
+        lambda series: series.eq("")
+    )
+    output["passes_exact_identifier_screen"] = passes.astype(bool)
+    output["control_status_label"] = pd.Series(
+        [CONTROL_STATUS_LABEL if passed else pd.NA for passed in passes],
+        index=output.index,
+        dtype="string",
+    )
+    return output
+
+
+def flag_identifier_free_name_stress_set(
+    eligibility_audit: pd.DataFrame,
+    complete_award_history: pd.DataFrame,
+) -> pd.DataFrame:
+    """Flag exact normalized names found on identifier-free award rows.
+
+    The flag is for worst-case stress reporting only. It is not an eligibility
+    rule, contamination estimate, upper bound, fuzzy match, or alias screen.
+    """
+
+    _require_columns(
+        eligibility_audit,
+        frozenset({"entity_name", *_ELIGIBILITY_AUDIT_COLUMNS}),
+        label="eligibility audit",
+    )
+    _validate_history(eligibility_audit, complete_award_history)
+    _require_columns(
+        complete_award_history,
+        frozenset({"company_name"}),
+        label="complete SBIR/STTR award history",
+    )
+
+    history_uei = complete_award_history["uei"].map(normalize_uei)
+    history_duns = complete_award_history["duns"].map(normalize_duns)
+    history_names = complete_award_history["company_name"].map(_normalized_company_name)
+    identifier_free = history_uei.isna() & history_duns.isna()
+    reference_names = {
+        normalized
+        for normalized in history_names.loc[identifier_free]
+        if isinstance(normalized, str)
+    }
+
+    output = eligibility_audit.copy()
+    normalized_entity_name = output["entity_name"].map(_normalized_company_name)
+    output["normalized_entity_name"] = pd.Series(
+        normalized_entity_name,
+        index=output.index,
+        dtype="string",
+    )
+    output["identifier_free_award_exact_name_match"] = (
+        normalized_entity_name.notna() & normalized_entity_name.isin(reference_names)
+    ).astype(bool)
+    return output
+
+
+def _parse_prior_end_dates(values: pd.Series) -> pd.Series:
+    present = values.map(_normalized_record_key).ne("")
+    parsed = pd.to_datetime(values, errors="coerce", utc=True, format="mixed")
+    parsed = parsed.dt.tz_localize(None).dt.normalize()
+    if (present & parsed.isna()).any():
+        raise NegativeControlInputError(
+            "prior_period_of_performance_end contains an unparsable nonblank value"
+        )
+    return parsed
+
+
+def _date_multiset(values: pd.Series) -> Counter[int | None]:
+    return Counter(None if pd.isna(value) else int(pd.Timestamp(value).value) for value in values)
+
+
+def permute_prior_end_dates(pairs: pd.DataFrame) -> pd.DataFrame:
+    """Apply the one fixed placebo permutation at unique prior-award grain."""
+
+    required = frozenset({"prior_award_id", "prior_period_of_performance_end"})
+    _require_columns(pairs, required, label="placebo pair frame")
+    if pairs.empty:
+        return pairs.copy()
+
+    prior_keys = pairs["prior_award_id"].map(_normalized_record_key)
+    if prior_keys.eq("").any():
+        raise NegativeControlInputError("Every placebo pair row must have a prior_award_id")
+
+    prior_dates = _parse_prior_end_dates(pairs["prior_period_of_performance_end"])
+    award_rows = pd.DataFrame(
+        {"_prior_key": prior_keys, "_prior_end": prior_dates},
+        index=pairs.index,
+    )
+    date_variants = award_rows.groupby("_prior_key", sort=False)["_prior_end"].nunique(dropna=False)
+    if date_variants.gt(1).any():
+        raise NegativeControlInputError(
+            "Each prior_award_id must map to exactly one prior_period_of_performance_end"
+        )
+
+    unique_awards = (
+        award_rows.drop_duplicates("_prior_key", keep="first")
+        .sort_values("_prior_key", kind="stable")
+        .reset_index(drop=True)
+    )
+    donor_order = np.random.default_rng(PLACEBO_SEED).permutation(len(unique_awards))
+    shuffled_dates = unique_awards["_prior_end"].iloc[donor_order].reset_index(drop=True)
+    date_by_award = dict(zip(unique_awards["_prior_key"], shuffled_dates, strict=True))
+
+    output = pairs.copy()
+    output_dates = prior_keys.map(date_by_award)
+    output["prior_period_of_performance_end"] = pd.to_datetime(output_dates)
+
+    if len(output) != len(pairs) or not output.index.equals(pairs.index):
+        raise NegativeControlInputError("Placebo permutation changed pair row count or order")
+    for column in pairs.columns:
+        if column == "prior_period_of_performance_end":
+            continue
+        if not output[column].equals(pairs[column]):
+            raise NegativeControlInputError(
+                f"Placebo permutation changed prohibited non-date column: {column}"
+            )
+
+    output_award_dates = pd.DataFrame(
+        {"_prior_key": prior_keys, "_prior_end": output["prior_period_of_performance_end"]}
+    )
+    output_variants = output_award_dates.groupby("_prior_key", sort=False)["_prior_end"].nunique(
+        dropna=False
+    )
+    if output_variants.gt(1).any():
+        raise NegativeControlInputError(
+            "Placebo date was not propagated consistently across prior-award fan-out"
+        )
+    output_unique_dates = output_award_dates.drop_duplicates("_prior_key", keep="first")[
+        "_prior_end"
+    ]
+    if _date_multiset(output_unique_dates) != _date_multiset(unique_awards["_prior_end"]):
+        raise NegativeControlInputError(
+            "Placebo permutation did not preserve the award-grain date distribution"
+        )
+    return output
+
+
+def build_placebo_census_tables(
+    pairs: pd.DataFrame,
+    data_cut_date: date,
+) -> tuple[pd.DataFrame, pd.DataFrame]:
+    """Run one permuted frame through both unchanged frozen census helpers."""
+
+    placebo_pairs = permute_prior_end_dates(pairs)
+    return (
+        build_dropoff_ladder(placebo_pairs, data_cut_date),
+        build_sensitivity_grid(placebo_pairs, data_cut_date),
+    )
--- a/specs/phase-iii-negative-controls/design.md
+++ b/specs/phase-iii-negative-controls/design.md
@@ -15,12 +15,14 @@ materialization can be approved.
 The following decisions were approved on **2026-08-02 with no result visibility**:
 
 1. Retained controls are labeled **“no observed exact-identifier SBIR/STTR match,”** not
-   certified SBIR-negative.
+   certified SBIR-negative. A candidate with no usable identifier was not screened at
+   all and carries the distinct label **“no usable identifier; not screenable”** —
+   see the correction below.
 2. The core eligibility exclusion is exact strict-normalized UEI/DUNS against the
    complete available SBIR/STTR history.
-3. An exact normalized-name match to an identifier-free award recipient is only a
-   worst-case stress-set/reporting flag. It is neither an upper bound nor an
-   inclusion/exclusion rule, and exact names do not bound aliases.
+3. An exact normalized-name match to an award recipient the identifier screen could not
+   reach is only a worst-case stress-set/reporting flag. It is neither an upper bound
+   nor an inclusion/exclusion rule, and exact names do not bound aliases.
 4. Employee count is omitted, without a proxy or new bands. The limitation must be
    prominent in eventual methods and results.
 5. The single placebo uses seed `20260801`. It shuffles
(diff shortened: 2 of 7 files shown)


specs/phase-iii-negative-controls/design.md
```diff
@@ -0,0 +1,52 @@
...
+
+This stub makes the validation dependency visible before a count can acquire the
+appearance of a finished result. It adds no control implementation, resolution method,
+new numeric cutoff, or result.
+
+## Inherited design
+
+The control and SBIR arms must use the identical frozen census filter and pseudo-index
+design from [`../phase-iii-census/design.md`](../phase-iii-census/design.md). The filter
+must not branch on arm membership; an `if is_control:` inclusion path is a design failure.
```
hollomancer: This is the most valuable sentence in the document and it's worth saying so. A negative-control design that lets the filter see arm membership can produce any separation it likes; naming `if is_control:` as a *design failure* rather than a style preference is the thing that makes the whole exercise falsifiable.

One suggestion to make it enforceable rather than aspirational: this is the requirement that a test could actually pin **today**, without any of the blocked data. A test asserting that the census filter's call signature takes no arm/label argument — or that the filter module contains no reference to the arm column — would be green now, would live on `main`, and would go red the moment someone adds the branch. That's a real gate, and it's the one piece of this spec that isn't blocked on SAM or SBIR.gov.

Also confirmed the inherited references check out: `specs/phase-iii-census/design.md` exists on `main`, and the `0.1` balance limit in requirement 3 is genuinely pre-authorized there (line 309–310), not a new cutoff introduced here. The "adds no new numeric cutoff" claim in the preamble holds.

---

specs/phase-iii-negative-controls/design.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Required work before this gate can close
+
+1. **Full-history eligibility:** certify that every control firm received no SBIR/STTR
+   award anywhere in the complete award history, and report every exclusion with its
+   reason.
```
hollomancer: **As written, this requirement can never be satisfied — which makes the gate permanent rather than blocking.**

"Certify that every control firm received no SBIR/STTR award anywhere in the complete award history" demands proof of a negative over a public record that the blockers section correctly says has a material population with no reliable UEI/DUNS. That gap doesn't close by finding a better join — SBIR.gov's historical records simply don't carry the identifiers for part of the population, and no future work changes that. A requirement whose blocker is a permanent property of the source is an unfalsifiable condition, not a dependency.

The standard framing for this in the matched-control literature is to **bound** the contamination rather than eliminate it:

- estimate an upper bound on the share of "controls" that are undetected SBIR recipients (name-based screen on the unlinkable population gives a rate, even if it can't adjudicate individuals),
- report that bound as a pre-registered quantity, and
- show the headline contrast survives the worst case — if every unlinkable-but-name-similar firm is assumed SBIR-positive, does the conclusion hold?

That converts an impossible certification into a measurable robustness claim, and it's consistent with the doc's own instinct to "report every exclusion with its reason." I'd rather see this changed now, while it's a scaffold, than have someone discover in six months that the gate has no reachable exit.

Requirement 2 has a milder version of the same issue: `employee-count band` is a required matching covariate *and* a listed blocker, so the matching spec isn't executable as written. The doc should state which way it resolves — drop the covariate (and say what that costs in balance), or name the source that would supply it — otherwise the resolution path is undefined and the next person has to re-litigate it.

---

tests/unit/phase_iii_negative_controls/test_release_gate.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+
+pytestmark = pytest.mark.fast
```
hollomancer, line 27: `pytest.mark.fast` puts this in the `Tests (unit-fast)` lane, which is where it currently fails — matching the PR description exactly, so CI is behaving as designed.

Two consequences worth being deliberate about:

1. **If this ever merges, `main` is red permanently.** `unit-fast` is a required check, so every unrelated PR in the repo inherits the failure, and the design doc forecloses the escapes ("Removing, skipping, or marking it expected-to-fail is not a resolution"). That's a real commitment device, but it means the repo cannot return to green until the blocked research completes — and one of those blockers may not be resolvable at all (see my comment on requirement 1). The draft status is doing the actual work of preventing that today; the doc should say plainly that merging is not intended until the gate can close.
2. **`pytest -x` locally aborts the whole suite here.** Anyone running the fast lane with fail-fast stops on this sentinel before reaching their own tests. A `# noqa`-style carve-out isn't available for a deliberate failure, so at minimum this is worth a note in the module docstring (the file currently has none) so someone hitting it locally understands it in one read rather than going to `git log`.

The failure message itself is well written — it tells the reader exactly what would justify replacing it, which is the right property for a sentinel.

---

hollomancer (commented):
## Review: block Phase III census release on negative controls

Reviewed both files. CI is red exactly as described — `Tests (unit-fast)` and `CI Summary` fail, everything else passes — so the "expected CI state" section is accurate.

**I think the instinct here is right and the scope discipline is genuinely admirable.** A PR that adds two files, materializes no result, quotes no count, and refuses to pick a resolution method it hasn't earned is the correct response to "Phase 1 audit tables are not validated Phase III results." The arm-blind requirement and the distributional-reporting requirement (full per-firm distribution and overlap coefficient, not a single headline ratio) are the two clauses that actually prevent a tautological census, and both are stated crisply. I verified the inherited references: `specs/phase-iii-census/design.md` exists on `main` and pre-authorizes the `0.1` SMD limit, so the "no new numeric cutoff" claim holds.

My concerns are about whether this **works as a gate**, and about one requirement that I think is unsatisfiable as written.

### The enforcement mechanism doesn't bind anything

The PR body concedes this: "it cannot technically prevent the census asset itself from being materialized." I'd put it more strongly — the red test constrains only the branch it lives on, and a draft PR nobody intends to merge was already unmergeable. So the sentinel buys visibility, not enforcement, and visibility was already achieved by opening the PR and writing the design doc.

A test that would actually bind, and that could be written **today** with no blocked data:

- **Arm-blindness** — assert the census filter takes no arm/label parameter and its module never references the arm column. Green now, red the instant someone adds `if is_control:`.
- **Artifact precondition** — assert that if a census output exists at the frozen path, a negative-control and placebo artifact exist alongside it. Green now (no census materialized), red exactly when someone publishes a count without controls.

Either one lives on `main`, keeps CI green, and fails on the specific event this PR is trying to prevent. That's the difference between a sentinel and a gate, and it's what would let this merge instead of sitting open indefinitely.

### Requirement 1 has no reachable exit

Detailed in the inline comment. "Certify that every control firm received no SBIR/STTR award anywhere in the complete award history" asks for proof of a negative against a source the doc itself says lacks the identifiers to support it. That's a permanent property of SBIR.gov's historical records, not a gap that future work closes — so the gate as drafted can never open. Bounding the contamination rate and showing robustness to the worst case turns it into something achievable without weakening it. Worth fixing while this is still a scaffold.

Requirement 2 has a smaller version: employee-count band is simultaneously a required covariate and a listed blocker, so the matching spec isn't executable and the doc doesn't say which way it resolves.

### Repo convention

`specs/` dirs carry `requirements.md` / `design.md` / `tasks.md` (19/17/17 on `main`). This one has `design.md` only, and its content is mostly requirements — "Required work before this gate can close" is a numbered SHALL list, not a design. Splitting it would also make the spec legible to `/review-spec` and the spec-status survey, which classify by exactly these filenames; a design-only directory will be classified oddly by both.

### Minor

- The "Base dependency / #475 merged while this draft was being opened" framing is now historical. Since the PR targets `main` and #475 is in, the dependency section could just state the invariant (Phase 1 tables are not validated results) without the merge-race narrative.
- The test module has no docstring — worth one, given it's a deliberate failure someone will hit locally.

Nothing here is a scope objection — the PR does exactly what it says and no more. The question I'd want answered before this merges is whether the gate should be a sentinel or an actual precondition test, and I'd argue for the latter.

---

hollomancer:
## Re-review — all three threads addressed in `7e3e274`

All three were accurate. Dispositions:

**1. The unfalsifiable certification (requirement 1)** — this was the one worth blocking on, and I've made the change now rather than leaving it as a scaffold. "Certify that every control firm received no SBIR/STTR award anywhere in the complete award history" demanded proof of a negative over a population whose missing UEI/DUNS is a *permanent property of the source*, not a pending dependency. As written the gate had no reachable exit. Rewritten to the bounding treatment you outlined: name-based screen over the unlinkable population to estimate an upper bound on contamination, pre-register that bound, and show the headline contrast survives the worst case where every unlinkable name-similar firm is assumed SBIR-positive. The corresponding blocker entry now says the source property is permanent and that the *remaining* work is building the bound — which is real, unstarted work, so the gate still blocks, just reachably.

**2. Requirement 2's self-contradiction** — confirmed: employee-count band was a required matching covariate *and* a listed blocker, so the matching design wasn't executable. Resolved in one direction rather than left undefined: the covariate is dropped, and the cost is stated where it will be read — firm size is the covariate most plausibly correlated with both SBIR participation and contract outcomes, so its omission is the largest known threat to balance and must be reported with the results, not just buried in the design. Reinstating it later is a strict improvement, not a precondition.

**3. The permanently-red-`main` hazard** — confirmed on both counts, and both are now written down rather than relying on the draft flag alone. The design doc and the test's new module docstring both say plainly that this branch is not intended to merge while the gate is closed, and why: `unit-fast` is a required check, so merging would inherit the failure to every unrelated PR in the repo, and the design forecloses the usual escapes. The sentinel had no docstring at all; it now explains itself in one read, including the `--deselect` incantation for anyone whose `pytest -x` run dies on it.

**On your enforceable-today point** — you're right that arm-blindness is the one requirement a green test could pin now with none of the blocked data, and it's the most valuable thing in this PR. I've recorded it in the design doc (and in the sentinel's docstring) as belonging on `main` as an independent green test rather than here. I did **not** write that test on this branch, deliberately: putting it behind a permanently-red sentinel is exactly where it *can't* do its job. It wants its own PR against the census filter in #475. Say the word and I'll open it.

**Mergeability: no — intentionally, and that's now documented rather than implicit.** CI is red by design (`Tests (unit-fast)` → `CI Summary`), which is the PR working as intended. Everything else on the branch is green. This should stay a draft until the contamination bound and the placebo artifacts exist.

---

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
```diff
@@ -0,0 +1,296 @@
...
+        for normalized in complete_award_history["duns"].map(normalize_duns)
+        if isinstance(normalized, str)
+    }
+
+    output = candidate_entities.copy()
+    normalized_uei = output["uei"].map(normalize_uei)
+    normalized_duns = output["duns"].map(normalize_duns)
+    exact_uei_match = normalized_uei.isin(history_ueis)
+    exact_duns_match = normalized_duns.isin(history_duns)
+    passes = ~(exact_uei_match | exact_duns_match)
```
hollomancer, line 147: **An entity with no usable identifier passes this screen identically to one that was actually screened — and those are the two populations the whole design turns on.**

`passes = ~(uei_match | duns_match)` is `True` in three distinct situations: the entity has a usable identifier and it wasn't in history (genuinely screened clean), the entity has no identifier at all, and the entity's identifiers are malformed enough that `normalize_uei`/`normalize_duns` reject them. Confirmed at `81a1991`:

```
  entity_id  normalized_uei  normalized_duns  passes  control_status_label
0        E1    ABC123DEF4G5        123456789    True  no observed exact-identifier SBIR/STTR match
1        E2            <NA>             <NA>    True  no observed exact-identifier SBIR/STTR match
2        E3            <NA>             <NA>    True  no observed exact-identifier SBIR/STTR match
```

E2 has no identifiers; E3's are junk. Both carry the same `control_status_label` as E1.

The docstring is scrupulous about this — "passing this screen means only that no exact usable identifier was observed" — and that's exactly right as a statement about the *method*. But the frame it returns doesn't carry the distinction, so a downstream consumer reading `passes_exact_identifier_screen` cannot tell "screened, clean" from "not screenable." A caller *could* infer it from `normalized_uei`/`normalized_duns` both being `<NA>`, but that's reconstructing a first-class fact from a side effect, and the first consumer to write `audit[audit.passes_exact_identifier_screen]` gets a control population silently diluted with unscreenable firms.

This matters more here than it would elsewhere because of where the design landed after the last round. The contamination bound is built over precisely the population that can't be linked by identifier — that's the whole point of the rewritten requirement 1, and `flag_identifier_free_name_stress_set` is the machinery for it on the *history* side. But the same partition on the *candidate* side is left implicit. The unscreenable candidates are the denominator of the bound you now owe, and right now nothing in the audit output names them.

Suggested change, small and local: emit `has_usable_identifier` (or `screenable`) alongside `passes_exact_identifier_screen`, and either leave `control_status_label` null for unscreenable rows or give them their own label — something like `"no usable identifier; not screenable"`. That keeps the honest three-way distinction the docstring already makes in prose, and it gives the future orchestration layer a column to count rather than a convention to remember.

Worth a fixture test in the same file: two candidates, one screened-clean and one identifier-free, asserting they are *distinguishable* in the output. That test would be green today, needs none of the blocked data, and pins the property that the docstring currently only promises.

---

hollomancer (commented):
## Re-review at `81a1991` — prior round fully addressed; one finding on the new code

**The three earlier threads are closed on the merits, not papered over.** Requirement 1 now asks for a pre-registered contamination *bound* over the unlinkable population with a worst-case robustness check, and the blocker entry states plainly that the missing identifiers are a permanent property of the source while the remaining work — building the bound — is real and unstarted. Requirement 2's self-contradiction is resolved in one direction, with the cost of dropping the employee-count covariate stated where results will be read rather than buried. And the permanently-red-`main` hazard is written down instead of relying on the draft flag: `test_release_gate.py` now opens with a docstring that explains the sentinel, says outright that the branch is not intended to merge while the gate is closed and why (`unit-fast` is required), and hands the reader the `--deselect` incantation for `pytest -x`. That last one is the difference between a commitment device and a trap.

`a6c90a7` adds real code since that round — `methods.py`, 296 lines, plus 332 lines of fixture tests — so this is a review of that, not a re-read of the docs.

**It holds up well.** The module scopes itself narrowly and says so ("does not choose a control population, match firms, materialize an artifact, or change the frozen census criteria"), and the code keeps that promise: `build_placebo_census_tables` delegates to the unchanged frozen helpers rather than reimplementing them, and `permute_prior_end_dates` permutes at unique prior-award grain with the permutation asserted afterward — row count, index, every non-date column, per-award date consistency across fan-out, and the award-grain date multiset all verified post hoc rather than assumed. Deriving the donor order from a `_prior_key`-sorted frame makes the mapping independent of input row order, and there's a test pinning exactly that. Fixture tests: **11 passed**; the sentinel fails as designed.

**One finding:** `audit_exact_identifier_eligibility` cannot distinguish a candidate that was screened and came back clean from one that has no usable identifier to screen with — both get `passes_exact_identifier_screen=True` and the same `control_status_label`. The docstring is careful about this in prose, but the returned frame doesn't carry the distinction, and the unscreenable candidates are precisely the population the newly-required contamination bound is computed over. Reproduction and a small suggested fix in the line comment. Not blocking the branch — nothing here is mergeable yet by design — but worth fixing while the module is still this small, since every future consumer will read `passes_exact_identifier_screen` as "clean."

Two notes, neither actionable:
- `permute_prior_end_dates` will map some awards to their own date, and is the identity for a single-award frame. That's correct behavior for a placebo and not something to "fix," but when the empirical result exists it's worth reporting the fixed-point rate alongside it, so a reader can see the permutation actually moved the data.
- `_parse_prior_end_dates` uses `format="mixed"`, which is per-element inference. Fine at this grain; if it ever runs over the full history, it's the first thing to bite on speed.

**Mergeability: no, still intentionally.** CI is red by design and the sentinel is now self-documenting about it. This stays a draft until the contamination bound and the placebo artifacts exist.

---

specs/phase-iii-negative-controls/design.md
hollomancer: Verified addressed at `81a1991` — resolving.

The design doc now records arm-blindness as a property that belongs on `main` as an independent green test, and the sentinel's new module docstring repeats it. Agreed with the reasoning for not writing it on this branch: a green invariant sitting behind a permanently-red sentinel in the same lane cannot do its job.

Leaving one thing on the record so it doesn't evaporate with the thread: the test itself is still owed, and it is the only piece of this spec that is unblocked today. It wants its own PR against the census filter in #475, and it should land regardless of what happens to this branch — including if this branch is eventually abandoned.

---

tests/unit/phase_iii_negative_controls/test_release_gate.py
hollomancer: Verified addressed at `81a1991` — resolving.

The sentinel now has a module docstring that covers both points in one read: that merging turns `main` red for every unrelated PR because `unit-fast` is required, and the exact `--deselect` node id for anyone whose local `pytest -x` dies on it. It also states plainly that the branch is not intended to merge while the gate is closed, which is the part that previously lived only in the draft flag.

---

specs/phase-iii-negative-controls/design.md
hollomancer: Verified addressed at `81a1991` — resolving, with one correction to the record.

Requirement 2 is cleanly resolved: employee count is dropped rather than left as a covariate that is simultaneously required and blocked, and `design.md` § "Employee count is deliberately absent" states the cost where it will be read — GSA classifies `assertions.sizeMetrics.averageNumberOfEmployees` as FOUO, so this isn't a pending dependency either, and firm size is the covariate most plausibly correlated with both arms. That's the right call and it's executable now.

Requirement 1 no longer demands proof of a negative, so the unfalsifiable gate is gone. But the shipped mechanism is **not** the bounding treatment, and the PR body is more accurate than my summary comment was: `flag_identifier_free_name_stress_set` is explicitly disclaimed as "neither an upper bound nor an inclusion/exclusion rule" (`design.md:22-23`), and the docstring repeats it. So what exists today is a stress-set flag; the pre-registered contamination bound is still entirely unbuilt.

That's a fine place to be — a flag you can compute honestly beats a bound you can't defend — but it means the gate's exit condition moved rather than shrank. Anyone reading this thread later should not take "requirement 1 rewritten" as "requirement 1 satisfied."

---

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
hollomancer: Still open at `81a1991` — leaving unresolved.

Re-confirmed against the current head: `passes = ~(exact_uei_match | exact_duns_match)` is unchanged, no `has_usable_identifier`/`screenable` column was added, and `control_status_label` is still written for every passing row regardless of whether the row was screenable. The three-way distinction remains prose-only.

See the new thread on `flag_identifier_free_name_stress_set` — the two functions partition on *different* sides of the same missingness, which makes this one worse than it looks in isolation.

---

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
```diff
@@ -0,0 +1,296 @@
...
+    _require_columns(
+        complete_award_history,
+        frozenset({"company_name"}),
+        label="complete SBIR/STTR award history",
+    )
+
+    history_uei = complete_award_history["uei"].map(normalize_uei)
+    history_duns = complete_award_history["duns"].map(normalize_duns)
+    history_names = complete_award_history["company_name"].map(_normalized_company_name)
+    identifier_free = history_uei.isna() & history_duns.isna()
```
hollomancer, line 219: **The stress set is defined by missingness on the history side, but unscreenability lives on the candidate side — so the worst case falls through both functions.**

`identifier_free = history_uei.isna() & history_duns.isna()` restricts `reference_names` to award rows carrying *neither* identifier. That is the right population if the question is "which history rows can't be joined by identifier." But it is not the population that makes a candidate risky.

Walk the case that matters. A candidate firm has no UEI and no DUNS, and its normalized name exactly matches an SBIR award row that *does* carry a DUNS:

- `audit_exact_identifier_eligibility` — the candidate's `normalize_duns` yields `<NA>`, so `exact_duns_match` is `False`, `passes` is `True`, and it gets `control_status_label = "no observed exact-identifier SBIR/STTR match"`. Nothing was screened; the label is earned by absence.
- `flag_identifier_free_name_stress_set` — the matching history row has a DUNS, so it never enters `reference_names`, and `identifier_free_award_exact_name_match` is `False`.

So the single most likely undetected SBIR recipient in the frame — no identifiers, name-identical to a real awardee — is reported as clean by the eligibility audit and absent from the stress set that exists precisely to surface worst cases. Both outputs are individually defensible and the composition has a hole.

This is the same missingness as the open thread on line 130, seen from the other side. That one is about the audit not distinguishing *screened-clean* from *not-screenable*; this one is about the stress set not covering the not-screenable candidates at all. Fixing line 130 alone doesn't close this: even with a `screenable` column, the stress flag would still miss the case above, because the flag's own reference population is derived from the wrong side of the join.

Two things worth separating in the design before either is coded further:

1. **History rows that can't be joined** (current `identifier_free`) — this bounds what an identifier join could ever have caught. Legitimate, and correctly what `reference_names` computes.
2. **Candidates that can't be screened** (no usable identifier after normalization) — for these, *no* history row is reachable by identifier, so the relevant name-comparison population is the **entire** history, not just its identifier-free subset.

The stress set the design says it wants — worst-case reporting — needs (2), or the union. Restricting to (1) understates it in exactly the direction that flatters the control arm, which is the direction this whole spec is built to guard against.

Since this is fixture-level mechanics and needs none of the blocked data, it's testable today: a candidate with no identifiers whose name matches a DUNS-carrying history row should be visible in the output as *something* — either not-screenable, or stress-flagged. Right now it is neither.

Not asking for the fix on this branch necessarily — but the design doc currently describes the flag in terms of "identifier-free award rows," so the narrower population looks deliberate rather than an oversight. If it is deliberate, the reason belongs next to it, because the reading above is the natural one.

---

hollomancer (commented):
## Re-review of #478 at `81a1991`

**Three of four prior threads are addressed and resolved.** The employee-count contradiction is cleanly resolved (dropped, with the FOUO source constraint and the balance cost written where they'll be read), the sentinel now explains itself in one read including the `--deselect` node id, and the permanently-red-`main` hazard is documented rather than resting on the draft flag.

**One correction to the record on requirement 1.** The unfalsifiable certification is gone, which was the blocking issue. But the shipped mechanism is a stress-set flag that the design explicitly disclaims as "neither an upper bound nor an inclusion/exclusion rule" — it is not the bounding treatment. The gate's exit condition moved rather than shrank; the pre-registered contamination bound is still unbuilt. Noted on that thread so it doesn't get lost.

**The line-130 thread stays open** — re-confirmed unchanged at head, no `screenable` column, `control_status_label` still written for unscreenable rows.

**One new finding**, on `flag_identifier_free_name_stress_set`: it derives its reference names from history rows missing *both* identifiers, but the candidates that need stressing are the ones missing their *own* identifiers. A candidate with no UEI/DUNS whose name matches a DUNS-carrying award row passes the eligibility audit as clean *and* is absent from the stress set. Both functions behave as documented; the composition leaves the highest-risk case invisible in both outputs, in the direction that flatters the control arm. Details inline.

**Mergeability: no, by design** — `unit-fast` red via the sentinel, which is the PR working as intended. Everything else on the branch is green.

**Still owed and still unblocked:** the arm-blindness test against the census filter in #475. It is the one property here that a green test could pin today, and it should land on `main` independently of this branch's fate.

---

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
hollomancer: Fixed at `eed96d2`. The three-way distinction is now first-class in the frame rather than reconstructible from a side effect.

- `has_usable_identifier` is emitted alongside `passes_exact_identifier_screen`: false when both strict normalizers reject the candidate's identifiers.
- `control_status_label` no longer collapses the two passing populations. Screened-clean rows keep the approved `CONTROL_STATUS_LABEL`; unscreenable rows get `UNSCREENABLE_STATUS_LABEL = "no usable identifier; not screenable"`. I took the distinct-label option over the leave-null one so the unscreenable population is something to *count* rather than something to detect by absence — which is the same reason you gave for not leaving it to `normalized_uei`/`normalized_duns`.

`passes_exact_identifier_screen` itself is unchanged. It still means what the docstring said it meant, and the caller who writes `audit[audit.passes_exact_identifier_screen]` still gets a diluted population — but now there is a column that says so, and the label makes the dilution visible in any groupby of the output.

The fixture test you asked for is in `test_exact_identifier_audit_separates_screened_clean_from_unscreenable`. The existing fixture had no screened-clean candidate at all — every screenable entity in it matched history — so `ENTITY-CLEAN` (usable UEI absent from history) was added to make the contrast expressible. It asserts all three of `ENTITY-CLEAN`, `ENTITY-NAME-STRESS` and `ENTITY-ID-BACKED-NAME` pass, and that the first is distinguishable from the other two on both the new column and the label. Green today, no blocked data.

The other side of this — the stress set — is fixed in the same commit; see that thread.

---

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
hollomancer: Fixed at `eed96d2` rather than documented — the narrower population wasn't deliberate, and the case you walked was reproducible exactly as described.

The reference set is now chosen per candidate, on the principle that "unreachable" is a property of the candidate and not of the history alone:

- **screenable candidate** → identifier-free award rows only (unchanged). Every identifier-carrying row was already adjudicated by the identifier screen, so a name flag there is noise.
- **unscreenable candidate** → the **entire** history, because the identifier screen reached none of it.

That is your option (2), not the union: for a screenable candidate the union would re-flag rows the identifier join already cleared.

The helper and its column are renamed — `flag_identifier_unreachable_name_stress_set` and `identifier_unreachable_award_exact_name_match` — because "identifier-free" now names the wrong population, and leaving the old name on the broader behavior would reproduce the same misreading in the other direction.

The fixture already contained the failing case and pinned it: `ENTITY-ID-BACKED-NAME` has malformed identifiers and normalizes to the same name as `AWARD-NAME-WITH-ID`, which carries a UEI — and the old test asserted it was **not** flagged. That assertion is now inverted, with `test_unscreenable_candidate_is_stressed_against_the_whole_history` covering the case and `ENTITY-UEI` covering the screenable side (shares a name with the identifier-carrying `AWARD-UEI`, must stay unflagged). Reverting the method change alone turns that test red; verified. 13 passed, sentinel still failing as designed.

**One thing needing your sign-off, not mine.** This changes approved decisions 1 and 3, which were approved dated and result-blind. I've recorded it in `design.md` under a "Correction (2026-08-03) — pending sign-off" heading rather than editing the approved wording silently. My argument for making the change on this branch: both edits are strictly conservative — they can only move rows *into* the unscreenable or stressed populations, never out — and no result exists or has been viewed, so nothing here is a retrospective answer. But the governance call is yours; if you'd rather the approved text stand and this ride as a documented known gap, say so and I'll revert the two behavior changes and keep only the note.

Unchanged either way: this is still a reporting flag, still not a bound, and per your note on the other thread the pre-registered contamination bound remains entirely unbuilt.

---

tests/unit/phase_iii_negative_controls/test_release_gate.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+import pytest
+
+
+pytestmark = pytest.mark.fast
+
+
+def test_negative_controls_and_placebo_release_gate_is_closed() -> None:
+    pytest.fail(
```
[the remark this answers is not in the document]
hollomancer: Declining this one deliberately — the `fast` placement is the mechanism, not an oversight, and the cost you describe is already bounded.

The sentinel is a commitment device from `specs/phase-iii-negative-controls/design.md`: no Phase III census count may be released as a validated result until provenance-backed negative-control and placebo artifacts exist. Moving it to a dedicated marker weakens exactly the property that makes it work. A gate in its own lane is one someone can decide to stop caring about; a gate in the required lane cannot be quietly downgraded, only closed with evidence.

On "poisoning required CI": that only bites on merge, and merging is foreclosed. The PR is a draft and both the design doc and the test's module docstring say plainly it is not intended to merge while the gate is closed — precisely because `unit-fast` is required and merging would turn `main` red for every unrelated PR. Adding a separate-but-still-required job wouldn't change that arithmetic; it would just relocate the red.

On local `pytest -m fast` ergonomics: that's real, and it's already handled the way you'd want a deliberate failure handled — the module docstring leads with the exact `--deselect` node id, so someone whose `pytest -x` dies here understands it in one read rather than going to `git log`. That was added in response to [an earlier thread](https://github.com/hollomancer/sbir-analytics/pull/478#discussion_r3696402633) raising the same two consequences.

The other two threads on this PR were accurate and are fixed at `eed96d2`.

---

hollomancer:
## Review — not mergeable, correctly; two threads fixed, one declined

Reviewed alongside #516, #515 and #504.

### CI

`Fast Tests (4/4)` is red. That is the release sentinel doing its job, not a defect — every other check on the branch is green, including `Lint, Types, and Guards` after this round's changes. **Not mergeable, and that is the intended state.** Merging would make `unit-fast` red for every unrelated PR in the repo, which the design forecloses; the draft flag and the sentinel docstring both now say so in writing rather than relying on someone remembering.

### This round's threads

- **Screened-clean vs. not-screenable** (`methods.py:130`) — fixed at `eed96d2`. `has_usable_identifier` is emitted, and unscreenable passing rows carry `UNSCREENABLE_STATUS_LABEL` instead of the approved control label, so the distinction the docstring made in prose is now a column and a value you can group by. New fixture test; the fixture had no screened-clean candidate at all, so one was added to make the contrast expressible.
- **The stress set's reference population** (`methods.py:185`) — fixed in the same commit. The reference set is now chosen per candidate: identifier-free award rows for a screenable candidate, the entire history for an unscreenable one. The old test asserted the hole (`ENTITY-ID-BACKED-NAME` — malformed identifiers, name-identical to a UEI-carrying awardee — was asserted *not* flagged); that assertion is inverted, and reverting the method change alone turns the new test red. Verified.
- **The sentinel's `fast` marker** (Copilot) — declined, with reasons on the thread. The required-lane placement is the commitment device; a gate in its own lane is one that can be quietly stopped caring about. The local-ergonomics cost is real and already handled by the `--deselect` node id in the module docstring.

### Still owed, so it doesn't evaporate

1. **Your sign-off on the design correction.** The two fixes change approved decisions 1 and 3, which were approved dated and result-blind. They're recorded in `design.md` under a "Correction (2026-08-03) — pending sign-off" heading rather than edited silently. My argument for making them here: both are strictly conservative — they can only move rows *into* the unscreenable or stressed populations, never out — and no result exists or has been viewed, so neither is a retrospective answer to a design question. The governance call is yours; say the word and I'll revert the behavior and keep only the note.
2. **The contamination bound remains unbuilt.** Per the earlier thread, "requirement 1 rewritten" is not "requirement 1 satisfied" — what exists is a stress-set flag, explicitly disclaimed as neither a bound nor a rule.
3. **The arm-blindness test.** Still the only piece of this spec that is unblocked today, still wants its own PR against the census filter in #475, and should land regardless of what happens to this branch.

### Interaction with #515

Both PRs create `packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/__init__.py`. Each merges cleanly into `main` alone; whichever lands second hits an add/add conflict. The resolution is a union of the export lists and needs doing by hand — resolving it with `--theirs` while testing broke #515's `scripts/data/extract_phase_iii_*.py` with an `ImportError` at collection. Given this branch isn't merging while the gate is closed, #515 will almost certainly land first and this branch will carry the rebase.

---

hollomancer:
## Merge readiness: not ready — `main` has moved past this branch, and the conflict is structural

The add/add conflict I reported earlier is still there, and having now looked at both sides properly, it is not a merge-resolution problem. It's a "this PR and `main` are two implementations of the same thing" problem.

```
CONFLICT (add/add): Merge conflict in packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/__init__.py
```

### What each side has

| | this branch | `main` |
|---|---|---|
| modules | `__init__.py`, `methods.py` | `__init__.py` + 10 modules (`placebo.py`, `identity.py`, `sam_eligibility.py`, `covariates.py`, `matching.py`, `outcomes.py`, `quarantine.py`, `source_keys.py`, `february_mirror.py`, `nih_reporter.py`) |
| `PLACEBO_SEED` | `methods.py:23` = `20260801` | `placebo.py:16` = `20260801` |
| census tables | `methods.py:326 build_placebo_census_tables` | `placebo.py:264 build_placebo_census_tables` |
| permutation | `methods.py:260 permute_prior_end_dates` | `placebo.py:243 permute_prior_end_dates_across_firms` |
| eligibility audit | `methods.py:103 audit_exact_identifier_eligibility` | `sam_eligibility.py:220 build_sam_eligibility_table` |

Same seed, same concepts, different names and different module layout. Resolving `__init__.py` by combining both export lists would land two parallel implementations of the placebo mechanics in one package — which is the thing the primitives contract in CLAUDE.md exists to prevent, and it would be quietly wrong rather than loudly broken.

### What is actually still unique here

The unscreenable-candidate work is not on `main`. Nothing under `phase_iii_negative_controls/` there mentions `UNSCREENABLE_STATUS_LABEL`, `has_usable_identifier`, or the identifier-unreachable name stress set. That distinction — separating "screened and clean" from "no usable identifier, therefore not screenable", and computing the stress reference set per candidate accordingly — is the substantive contribution this branch still carries:

```python
screenable = output["has_usable_identifier"].astype(bool)
unreachable_name_match = normalized_entity_name.isin(identifier_free_names).where(
    screenable, normalized_entity_name.isin(every_award_name),
)
```

**To do:** rather than merging this branch, port the unscreenable distinction onto `main`'s module layout — `has_usable_identifier` and the status label into `sam_eligibility.py` next to `build_sam_eligibility_table`, the per-candidate stress set into `placebo.py`, and the corresponding tests from `tests/unit/phase_iii_negative_controls/test_methods.py` onto the modules that now own the behavior. Then close this PR in favor of that one.

If you'd rather keep this PR as the vehicle, the equivalent is rebasing onto `main` and rewriting `methods.py` away entirely — same work, more merge friction.

I'd like your call on which route before I do anything to the branch.

---

hollomancer:
## Recommendation: close this PR

Per the [merge-readiness review on 2026-08-05](https://github.com/hollomancer/sbir-analytics/pull/478#issuecomment-5192235610), this branch and `main` carry two parallel implementations of the same placebo mechanics (same seed, same concepts, different names and module layout). Resolving the add/add conflict by combining both sides would land duplicate implementations, which the primitives contract exists to prevent.

**Action:** close this draft rather than invest in rebasing. The substantive contributions — `has_usable_identifier`, `UNSCREENABLE_STATUS_LABEL`, and per-candidate stress reference sets — should be ported onto `main``'s module layout (`sam_eligibility.py` and `placebo.py`) in a fresh PR.

This keeps the branch from silently diverging further and makes the porting work explicit rather than a merge-resolution accident.

hollomancer:
**Mergeability review: not mergeable as-is (`mergeable_state: dirty`, 162 commits behind main), and likely overtaken by events.** This draft was opened as the visible validation dependency for the frozen census, intentionally red until "real, provenance-backed negative-control and fixed-seed placebo artifacts exist." Those artifacts now exist — delivered through main's implementation path (#522's control matches, the frozen placebo, the materialization approval gate) and recorded in the 2026-08-03 audits under `studies/phase-iii-census/` (identity-eligibility Rev 11, matching balance Rev 12/13, placebo falsification, negative-control outcomes). Merging a parallel foundation now would introduce a second implementation of concepts the census has since frozen.

Two mechanical blockers on top of that, both post-dating this branch: `specs/phase-iii-negative-controls/requirements.md` lacks the now-required `**Target epistemic tier:**` declaration (`check_epistemic_tiers.py` fails CI), and the spec has no `specs/status.md` entry (the repository-hygiene guard fails docs-check). Plus the `phase_iii_negative_controls/__init__.py` conflict.

**Suggested path:** skim the head commit ("Distinguish unscreenable candidates in the control audit and stress set") for analysis distinctions that never reached main's audit path — that's the one piece of potentially unsalvaged value — cherry-pick anything worth keeping onto a fresh branch off current main, then close this PR, optionally archiving the spec trio under `specs/archive/superseded/` with a completion record pointing at the census study contract.

---

hollomancer:
## The last reason to keep this branch is gone — the unscreenable work is superseded, not unsalvaged

The [2026-08-06 review](https://github.com/hollomancer/sbir-analytics/pull/478#issuecomment-5208131392) left one open question before closing: skim the head commit for analysis distinctions that never reached `main`, because that was "the one piece of potentially unsalvaged value." I checked it properly rather than skimming, and the answer flips the disposition from *port-then-close* to *just close*.

### The symbols are absent from `main`, but the concept is not

`git grep` for `has_usable_identifier`, `UNSCREENABLE_STATUS_LABEL`, and `identifier_unreachable` returns nothing on `main` — which is what the earlier review saw, and why it read as unsalvaged. But `main` expresses the same distinction under a different and stronger design, in `sam_eligibility.py`:

| this branch (`methods.py`) | `main` (`sam_eligibility.py`) |
|---|---|
| `has_usable_identifier` boolean column | `EligibilityStatus` — an exhaustive `StrEnum`, three preregistered outcomes |
| `CONTROL_STATUS_LABEL` (screened clean) | `ELIGIBLE_SCREENED_NEGATIVE` |
| `UNSCREENABLE_STATUS_LABEL` (no usable identifier) | `INDETERMINATE_POSSIBLE_SBIR` + reason `missing_comparable_name_state_key` |
| `identifier_unreachable_award_exact_name_match` (reporting flag) | `matched_unresolved_name_state_keys` / `matched_unresolved_address_zip_keys` |

The three-way partition this branch's threads fought to make first-class is the enum on `main`, and it arrived as a status with an exclusion-reason vocabulary rather than a boolean column plus a label string.

### `main`'s version is strictly stronger on both fixes

**On the stress set.** This branch's flag is explicitly inert — the docstring at `methods.py:200-202` says it "is not an eligibility rule, contamination estimate, upper bound, fuzzy match, or alias screen," and nothing downstream consumes it. On `main`, the same name/address signal is not a flag at all: a quarantine-key intersection sends the candidate to `INDETERMINATE_POSSIBLE_SBIR` (`sam_eligibility.py:331-336`), which *removes it from the screened-negative population*. The signal changes eligibility instead of being reported next to it. That is the thing the original thread wanted and this branch declined to build.

**On unscreenable candidates.** `main` doesn't label them, because it refuses to admit them:

```python
if normalize_uei(row.unique_entity_id) is None:
    raise IdentityRecoveryError("every SAM candidate row must have a valid UEI")
```

`sam_eligibility.py:250-251`. The population `has_usable_identifier` exists to mark cannot enter the frame — it fails closed at the boundary. Porting a column to distinguish rows that `main` rejects outright would be strictly redundant.

### This moots the sign-off you were asked for

The [pending governance question](https://github.com/hollomancer/sbir-analytics/pull/478#discussion_r3703858475) — whether corrections to approved decisions 1 and 3 could ride on this branch — needs no answer. Both corrections were edits to `methods.py`, a module `main` never adopted. There is no approved-decision drift to ratify or revert, because the code the corrections applied to is not on the path the census actually took. I've resolved that thread on those grounds rather than leaving it waiting on a decision that no longer has an object.

### State of the branch

186 commits behind `main`, up from 162 at the last review. The three mechanical blockers are unchanged and I re-verified them: the add/add conflict on `phase_iii_negative_controls/__init__.py`, no `**Target epistemic tier:**` declaration in `specs/phase-iii-negative-controls/requirements.md` (fails `check_epistemic_tiers.py`), and no `specs/status.md` entry (fails the hygiene guard).

**Recommendation, now unambiguous: close this PR with no porting work.** The three prior reviews recommending closure each carried a caveat about unported value; that caveat is discharged. I have not closed it myself — that stays your call — but nothing further is owed to the branch first.

One item genuinely survives all of this, and it is not on this branch: the **arm-blindness test** flagged on 2026-08-01 as the only requirement here that a green test could pin with none of the blocked data. It was never written. It belongs against the census filter on `main` regardless of what happens to this PR, and closing this branch is the last moment it would still be visible.

---

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/methods.py
hollomancer: Resolving this without the sign-off, because the decision no longer has an object.

Both corrections were edits to `methods.py`. `main` never adopted that module — the census took the `sam_eligibility.py` / `placebo.py` path instead, and the per-candidate stress set here has no counterpart there to drift from. So there is no approved-decision change on the census's actual path to ratify or revert; the wording you'd be signing off on governs code that isn't in the lineage.

The substance did land, in a stronger form than this thread asked for. The fix here made the stress set cover unscreenable candidates but kept it explicitly inert — `methods.py:200-202` disclaims it as "not an eligibility rule." On `main`, the same name/address signal intersects the quarantine keys and moves the candidate to `INDETERMINATE_POSSIBLE_SBIR` (`sam_eligibility.py:331-336`), so it changes eligibility rather than being reported alongside it. Full comparison in the [summary comment](https://github.com/hollomancer/sbir-analytics/pull/478#issuecomment-5233688805).

---

hollomancer:
Closing per the [supersession analysis above](https://github.com/hollomancer/sbir-analytics/pull/478#issuecomment-5233688805).

Short version of why this is a clean close rather than a lossy one: the branch's remaining unique contribution — separating screened-clean from not-screenable candidates — exists on `main` as `EligibilityStatus`, an exhaustive three-value enum with an exclusion-reason vocabulary, and `main`'s treatment is stronger. The name/address signal there moves a candidate to `INDETERMINATE_POSSIBLE_SBIR` instead of setting an inert reporting flag, and `build_sam_eligibility_table` rejects candidates without a valid UEI at the boundary, so the population this branch labeled cannot enter the frame at all. No porting work is owed.

The validation artifacts this draft was opened to gate on now exist through `main`'s implementation path and are recorded in the 2026-08-03 audits under `studies/phase-iii-census/`, so the release sentinel has served its purpose.

**One item outlives this branch and is not on `main`:** the arm-blindness test, flagged on 2026-08-01 as the only requirement in this spec that a green test could pin with none of the blocked data. It was never written. It belongs against the census filter, and closing this PR is the last place it's visible — worth an issue if it isn't tracked elsewhere.

The spec trio under `specs/phase-iii-negative-controls/` is still on this branch only, so nothing needs archiving on `main`; it lacks the now-required `**Target epistemic tier:**` declaration and a `specs/status.md` entry, which is part of why this could not have merged as-is.

---

PR #483, opened 2026-08-02
docs: restructure research-questions.md for readability

hollomancer:
## Description

`docs/research-questions.md` is the canonical inventory of what this repo exists to answer, but it had become hard to read. Two commits: a structural pass, then a plain-language pass over the reader-facing text.

---

## Commit 1 — structural pass (presentation only)

No questions, citations, status labels, `A-CP#` identifiers, PR/branch tags, or spec links added, removed, or changed in substance.

### What was making it hard to read

1. **Five orthogonal axes crammed into single prose bullets.** The A-CP4 FOCI question carried a lens tag, a legacy ID, a lower-bound caveat, the question, two citations, an answerability label, and a deps list — in one 90-word sentence. Finding the question meant parsing the metadata wrapped around it.
2. **Many "questions" were noun phrases.** *"Capability density & choke-point concentration map per CET area"*, *"Concentration-as-fragility"* — the reader had to reverse-engineer what was being asked.
3. **Sentence length.** Several single sentences ran 3–6 lines with nested em-dashes and parentheticals.

### Changes

- **Fixed shape for every question:** bold title (plus lens/ID tags) → the question, as a question → caveat if one applies → `Status:` if answerability is contested → a trailing `Deps / Refs / Spec` line.
- Rewrote noun-phrase entries as interrogatives; split the long sentences.
- **Moved Section A's framing** (CET spine, statutory grounding, scope-consolidation history) into a `Section A framing notes` subsection at the end, so Section A opens on questions rather than ten lines of framing. Easy to revert if you'd rather it stayed up top.
- Promoted the bolded sub-groupings in A3/A4 to real headings.
- Dep glossary → table; added a `How to read this document` key; added cross-section anchor links.

---

## Commit 2 — plain-language pass

Commit 1 deliberately left wording alone. This one applies a bounded rule, now stated in the reading key: **the title, question, and Status slots are written for an outside reader and stay in plain language; Deps, Spec, and the implementation notes are maintainer-facing and keep pipeline shorthand.** Those three slots are exactly where the audience pointers send external readers.

Three kinds of change:

**1. House shorthand translated.** The thick/thin/dense metaphor family (*capability density*, *supplier-base thickness*, *transition-thinness*, *thin bases*) and the nominalized compounds (*Concentration-as-fragility*, *Acquisition-erosion*, *Predictive erosion*) were invented in this repo — they look like terms of art but cannot be looked up. They now say how many firms there are and what is happening to them. *Whitespace* → *Coverage gaps*, keeping the house term parenthetically so existing readers still find it.

**2. Pipeline vocabulary out of audience-facing slots.** *Materialized* / *materialization* is Dagster's word for "computed and stored"; in a Status line read by an oversight staffer it looks like hedging. Now "built" / "run in production". Same for *stacked changes* and *grain*. The M&A implementation note keeps the original vocabulary, per the rule.

**3. Terms of art kept, but glossed.** HHI, left-censoring, survival probability, fill rate, and crowd-in/crowd-out are the correct words for this audience and were **not** replaced — each gained a short in-line gloss on first use. *Lower-bound proxy* is now defined once in the reading key and used bare thereafter, replacing six slightly different restatements.

The heaviest rewrite is B2's follow-on census question, which stacked five invented modifiers before reaching a verb:

> **Before:** How many exact-UEI, post-completion contract actions survive a pre-registered, label-free uncoded-follow-on proxy, and how do the audit counts change across its frozen clauses and agency/window cells?

> **After:** When a firm wins a federal contract after its SBIR award ends — the same firm by exact UEI, with no Phase III label required — how many of those contracts hold up as likely follow-on work? And how does that count shift as we vary the matching rules, the agency, and the time window?

Its methodology terms (*pre-registered*, *frozen criteria*) moved to a `Method:` line, where they answer a reviewer's concern about post-hoc tuning rather than blocking a scanner.

Also de-circularized three questions that restated their own titles (transition effectiveness rate, survival probability, categorization vs. transition likelihood) and unpacked E1's reconciliation status line.

**No question changed meaning and no status label changed what it asserts.**

## Type of Change

- [x] Documentation update

## Testing

No code changed, so no unit or integration tests apply. Content preservation was verified mechanically against the pre-change file after **both** commits:

- **Citations** — all 33 `[L#]` entries present; only additions (citations pulled into explicit `Refs` slots), zero removals.
- **Identifiers** — all 37 `A-CP#` occurrences preserved exactly.
- **PR / issue refs** — all 15 PR refs and both issue refs preserved.
- **Links** — every link target and `<https://…>` URL from the original present; all relative links resolve on disk; all internal anchors resolve to real headings.
- **Question counts** — unchanged in every section. Net +3 bullets, all from splitting the Form D "Format:" run-on into a list.
- **Status labels** — every answerability label preserved; commit 2 rewords some but changes none of their assertions.
- **Inbound links** — no other file in the repo links to an anchor inside this doc, so heading changes break nothing.
- **markdownlint** — clean under `.markdownlint.yaml`. Also fixes two pre-existing MD052 adjacent-reference errors (`[L1][L3]` → `[L1], [L3]`).

## Checklist

The Maintenance footer records both passes and the prior 2026-06-27 review. The open `[TODO: verify]` items (NSF ~18:1, A3 4:1 NASEM attribution) carry over untouched, as does the next-audit checklist.

### DIFF
--- a/docs/research-questions.md
+++ b/docs/research-questions.md
@@ -28,9 +28,21 @@ Each question is written in a fixed shape:
   Deps / Refs / Spec: what it needs, what it benchmarks against, where it is specified.
 ```
 
+The **title**, **question**, and **Status** slots are written for an outside
+reader — a congressional staffer or program officer who arrived from
+[Where to start, by audience](#where-to-start-by-audience). Keep them in plain
+language. The *Deps*, *Spec*, and implementation notes are for maintainers and
+may use pipeline shorthand freely.
+
 **Status** appears only where answerability is contested or partial. A question
 with no status line has no special caveat attached — check the linked spec and
 the *Spec* slot to see whether it is built.
+
+**Lower-bound proxy** marks a question whose answer can only ever undercount.
+It appears wherever we detect ownership or capital through public filings: SEC
+EDGAR shows disclosed, structured ownership and disclosed private capital, so
+anything held privately or beneath a corporate layer is invisible to us. A
+"zero" on such a question means *we found none*, never *there are none*.
 
 **Spec** links point at spec directories and design docs. A `(PR #…)` tag means
 the work landed in that pull request. A `(branch: …)` tag means it is in
@@ -126,20 +138,24 @@ perspective, see [F. Capital formation & entrepreneurial finance](#f-capital-for
   DLA), phase, and vintage?
   *Deps: none · Refs: [L18]*
 
-- **Capability density & choke-point concentration map** (cap/vuln) (A-CP1 concentration map, A-CP2 supplier-base thickness, A-CP3 geographic distribution)
-  For each CET area, how many distinct awardees are there, how much award volume
-  do they hold, and how concentrated is that volume (awardee **HHI**) across
-  NAICS sector and geography (state, congressional district)?
-  High density reads as capability. The same HHI inverted is the choke-point
-  concentration map, flagging single- and thin-supplier clusters and
-  geographically narrow bases. GAO's program-wide Phase II HHI of ~11 [L14] is
-  the diffuse baseline that area-level concentration is measured against.
+- **Supplier base size and concentration per CET area** (cap/vuln) (A-CP1 concentration map, A-CP2 supplier-base thickness, A-CP3 geographic distribution)
+  In each CET area, how many distinct firms hold awards, how much award money do
+  they hold, and how much of that money is concentrated in a few of them —
+  measured by **HHI** (the Herfindahl-Hirschman Index, the standard antitrust
+  measure of market concentration; higher means fewer, larger players) and broken
+  out by NAICS sector and by geography (state, congressional district)?
+  Many independent firms in an area means a healthy supplier base. Few firms
+  means a choke point. It is the same number read two ways, and the second
+  reading also flags areas whose suppliers are clustered in one part of the
+  country. GAO's program-wide Phase II HHI of ~11 [L14] is the diffuse baseline
+  that area-level concentration is measured against.
   **Status:** Answerable now for the classified DoD subset.
   *Deps: CET, ER, NAICS · Refs: [L14], [L16], [L29] · Spec: [dod_supply_chain_initial_analysis.md](research/dod_supply_chain_initial_analysis.md) (reproducible baseline and its limitations)*
 
-- **Whitespace** (cap)
-  Which CET subfields show DoD demand signals but sparse SBIR coverage?
-  Surfaced via semantic search over award and solicitation text.
+- **Coverage gaps** (cap) (house shorthand: *whitespace*)
+  Which CET subfields does DoD appear to want work in, but few SBIR awards
+  cover?
+  Found by semantic search over award and solicitation text.
   **Status:** Answerable now.
   *Deps: CET*
 
@@ -153,8 +169,7 @@ perspective, see [F. Capital formation & entrepreneurial finance](#f-capital-for
   What share of awardees — and of award dollars — in each CET area sit under
   disclosed foreign ownership, control, or influence, when screened against the
   eight Pub. L. 119-83 restricted-entity lists?
-  *Lower-bound proxy:* EDGAR Exhibit 21 / 8-K plus entity resolution detect only
-  disclosed, structured ownership — not private beneficial ownership.
+  *Lower-bound proxy* — ownership is read from EDGAR Exhibit 21 and 8-K filings.
   **Status:** Answerable now for the SEC-filer subset; the private majority needs
   data acquisition.
   *Deps: ER, SEC EDGAR, M&A signals · Refs: [L26] (screening lists), [L30] (foreign-supplier dependence), [L17] (foreign-acquisition risk)*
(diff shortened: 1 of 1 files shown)


docs/research-questions.md
```diff
@@ -15,335 +16,1019 @@ respected so foundational work appears before dependents.
...
+  DLA), phase, and vintage?
+  *Deps: none · Refs: [L18]*
+
+- **Supplier base size and concentration per CET area** (cap/vuln) (A-CP1 concentration map, A-CP2 supplier-base thickness, A-CP3 geographic distribution)
+  In each CET area, how many distinct firms hold awards, how much award money do
+  they hold, and how much of that money is concentrated in a few of them —
+  measured by **HHI** (the Herfindahl-Hirschman Index, the standard antitrust
+  measure of market concentration; higher means fewer, larger players) and broken
+  out by NAICS sector and by geography (state, congressional district)?
+  Many independent firms in an area means a healthy supplier base. Few firms
```
hollomancer: This plain-language rewrite turns an award-data screening signal into a conclusion: many awardees do not establish a healthy supplier base, and few awardees do not by themselves establish a choke point. That conflicts with the document’s own limitation that this baseline cannot support physical sole-source conclusions. Please keep the uncertainty explicit—for example, say that many firms suggest a broader in-program base while few firms flag an area for choke-point review.

docs/research-questions.md
```diff
@@ -15,335 +16,1019 @@ respected so foundational work appears before dependents.
...
+- **Phase II → III latency**
+  What is the elapsed time between Phase II completion and the first Phase III
+  contract?
+  GAO documents the newer §638(qq)(3) performance-standard framework and notes
+  that commercialization progress is measured from multiple SBA data sources
+  [L14].
+  *Deps: ER, ID · Refs: [L14] · Spec: [phase-transition-latency.md](phase-transition-latency.md)*
+
+- **Phase II → III survival probability**
+  What share of Phase II awardees go on to win Phase III work, and how does that
```
hollomancer: The old entry named a survival probability; rewriting it as the observed share that ever wins Phase III changes the estimand. A survival/time-to-event estimate needs an explicit horizon and treatment of right-censored awardees, whereas a raw share does not. The maintenance note also says survival probability was only glossed. Please phrase the question in terms of the probability of transition within a defined horizon, accounting for awards still under observation, or rename it if a simple cohort share is intended.

docs/research-questions.md
```diff
@@ -15,335 +16,1019 @@ respected so foundational work appears before dependents.
...
+  Which awards have missing or null critical fields (amount, dates, recipient)?
+  *Deps: none*
 
-- Why is `award_date` missing on ~50% of records, and can it be recovered non-destructively? — [../specs/data-imputation/](../specs/data-imputation/). *(deps: E3)*
-- For each imputable field (award date, amount, contract dates, NAICS, identifiers), which methods are available and at what confidence tier (high ≥90%, medium 75–90%, low <75%)? *(deps: E3)*
-- What is per-method backtest accuracy / MAE against ground-truth holdouts? *(deps: IMP)*
-- Can solicitation topics be mapped to NAICS (agency-topic crosswalk top-1 accuracy ≥75%)? *(deps: IMP, CET)*
-- Does the phase-transition precision benchmark remain ≥85% when imputed values are included? *(deps: IMP + transition detection)*
-- Which downstream consumers (Neo4j, CET, transition detection) should use raw vs. effective values? *(deps: IMP)*
+### E4. Data imputation (Tier 2–3)
```
hollomancer, line 774: Moving the parenthetical status out of this heading changes its GitHub anchor. The rewrite drops four pre-existing deep-link targets: the former master-question heading, the long out-of-scope heading, E4, and E5. No in-repo links currently use them, but this is the canonical externally linked inventory, so external links cannot be audited. Please preserve the old targets with explicit legacy anchors (or retain compatible headings) while adding the shorter new anchors.

docs/research-questions.md
hollomancer: Fixed in 50ef4881. The text now treats awardee count and HHI as screening signals, explicitly says they do not establish supplier health or a physical choke point, and defines HHI in terms of award-dollar concentration.

docs/research-questions.md
hollomancer: Fixed in 50ef4881. The question now asks for transition probability within a defined follow-up period and explicitly treats incomplete follow-up as right-censored rather than failure.

docs/research-questions.md
hollomancer: Fixed in 50ef4881. Added explicit HTML aliases for all four former GitHub heading slugs while retaining the shorter readable headings; the preservation check now proves every old heading anchor remains addressable.

hollomancer (commented):
## Re-review at `50ef488` — all three threads resolved on the merits

Checked each fix against what was actually asked, not just that a commit landed:

**Overclaiming from award-data concentration.** The old text said many firms "means a healthy supplier base" and few "means a choke point" — conclusions the document's own limitation section says this baseline can't support. Now they're hedged to the right strength ("suggest a broader in-program supplier base," "flag potential concentration for review") with an explicit disclaimer that these signals don't establish supplier health or a physical choke point. HHI is also now defined in terms of award-dollar concentration rather than generic "market concentration," which matters because the measured universe is in-program awards, not a market.

**The changed estimand.** "What share of Phase II awardees go on to win Phase III work" is a cohort proportion; the heading promised a survival probability. Now phrased as probability of transition within a defined follow-up period, with incomplete follow-up named as right-censored rather than counted as failure. That's the estimand the heading claims.

**Dropped anchors.** Four `<a id="…">` aliases, matching exactly the four slugs the restructure would have orphaned — the former master-question heading, the long out-of-scope heading, E4, and E5 — while the shorter readable headings stay. Since this is the canonical externally-linked inventory and external links can't be audited from here, preserving them rather than accepting the breakage was the right call.

**Mergeable.** Clean against current `main` — no conflicts, and nothing on `main` has touched `docs/research-questions.md` since this branch diverged. Docs-only, and it's the only non-draft PR in this range.

---

PR #488, opened 2026-08-02
Fix Make error helper collision

hollomancer:
## Summary

- rename Make's shell error-reporting helper to avoid a GNU Make built-in collision
- update the affected failure paths
- add regression coverage for `server-validate-config`

## Root cause

`$(call error,...)` resolves to GNU Make's built-in `error` function during recipe expansion, so conditional failure branches abort targets even when validation succeeds.

## Impact

`make server-validate-config` and other targets with conditional error reporting now execute their shell conditionals correctly.

## Validation

- 10 passed: `tests/unit/test_server_runtime_contract.py`
- Ruff check passed
- `git diff --check` passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ GRAY   := \033[90m
 info = if [ "$(QUIET)" != "1" ]; then printf "$(BLUE)➤$(RESET) %s\n" "$(1)"; fi
 success = if [ "$(QUIET)" != "1" ]; then printf "$(GREEN)✔$(RESET) %s\n" "$(1)"; fi
 warn = if [ "$(QUIET)" != "1" ]; then printf "$(YELLOW)⚠$(RESET) %s\n" "$(1)"; fi
-error = if [ "$(QUIET)" != "1" ]; then printf "$(RED)✖$(RESET) %s\n" "$(1)"; fi
+failure = if [ "$(QUIET)" != "1" ]; then printf "$(RED)✖$(RESET) %s\n" "$(1)"; fi
 print-cmd = if [ "$(QUIET)" != "1" ]; then printf "$(GRAY)$$ %s$(RESET)\n" "$(strip $(1))"; fi
 
 define run
@@ -119,7 +119,7 @@ docker-verify: env-check ## Verify Docker setup is working correctly
 	    cypher-shell -u $${NEO4J_USER:-neo4j} -p $${NEO4J_PASSWORD:-test} 'RETURN 1' >/dev/null 2>&1; then \
 	   $(call success,Neo4j is accessible at bolt://localhost:7687); \
 	 else \
-	   $(call error,Neo4j is not accessible); \
+	   $(call failure,Neo4j is not accessible); \
 	   $(call warn,Check logs with: make docker-logs SERVICE=neo4j); \
 	   exit 1; \
 	 fi; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,7 +42,7 @@ GRAY   := \033[90m
 info = if [ "$(QUIET)" != "1" ]; then printf "$(BLUE)➤$(RESET) %s\n" "$(1)"; fi
```
hollomancer, line 42: **`info` is the same bug, three lines up, and it is worse — it doesn't abort the recipe, it corrupts it into a bash syntax error.**

The diagnosis in this PR is right: `$(call error,…)` resolves to GNU Make's builtin `error` function, which fires at recipe-expansion time regardless of whether the branch is ever taken. But `info` is *also* a builtin (`$(info text)`), so `$(call info,…)` hits the identical collision and survives the rename.

The failure mode differs because the builtins differ. `$(error)` aborts, so you get a loud `*** … Stop.` — which is how this got noticed. `$(info)` prints its argument at expansion time and expands to **the empty string**, which leaves a bare `;` in the recipe. Verified on this branch at `6e91c37`:

```
$ make --no-print-directory validate-config DOCKER_COMPOSE=true QUIET=1
Validating docker-compose.yml          # printed by the builtin, at expansion time
Checking .env file                     # no ➤, no color, and QUIET=1 ignored
/bin/bash: -c: line 9: syntax error near unexpected token `;'
/bin/bash: -c: line 9: `   ; \'
make: *** [Makefile:567: validate-config] Error 2
```

So `make validate-config` does not work on this branch — the target this PR's sibling `server-validate-config` is modeled on. Applying the same rename you applied to `error` fixes it:

```
$ sed 's/^info = /notice = /; s/$(call info,/$(call notice,/g' Makefile > Makefile   # then:
$ make --no-print-directory validate-config DOCKER_COMPOSE=true QUIET=1; echo $?
0
```

Silent, exit 0, QUIET respected. Same one-line rename, ~30 call sites.

Two smaller notes while you're here:

- `warn` is safe only by luck — the builtin is spelled `warning`, not `warn`. `success` and `print-cmd` are fine. After renaming `info`, the helper namespace is clean, and it's worth a one-line comment above the block saying *why* these names avoid `error`/`info`/`warning`, so the next person naming a helper doesn't walk back into it.
- The new test pins exactly one of the eight `$(call failure,…)` sites. That's the right shape of test — it invokes `make` for real with `DOCKER_COMPOSE=true`, so it exercises expansion rather than grepping the file, and I confirmed it goes red on the pre-rename Makefile (`exit=2`, `*** … Stop.`) and green after. But it can only ever catch a regression in `server-validate-config`. A cheap complement that covers the whole class: assert no user-defined helper in the `$(call …)` block shares a name with a GNU Make builtin function. That's a static check over one small list, and it would have caught `info` today.

---

hollomancer (commented):
## Review — correct root cause, incomplete sweep

The diagnosis is right and the fix is right: `$(call error,…)` collides with GNU Make's builtin `error`, which fires during recipe *expansion*, so `server-validate-config` aborted even on the success path where the failure branch is never reached. I reproduced it and confirmed the rename fixes it, and that the new test actually discriminates: red on the pre-rename Makefile (`exit=2`, `*** docker-compose.server.yml validation failed.  Stop.`), green after. `tests/unit/test_server_runtime_contract.py` — 10 passed.

**One blocking finding: `info` has the identical collision and is still in the tree.** It's a builtin too, and because `$(info)` expands to the empty string rather than aborting, it leaves a bare `;` in the recipe — `make validate-config` currently dies with a bash syntax error rather than a legible Make one. Details and the reproduction are in the line comment. Same one-line rename.

That also means the sibling target this PR is patterned on is broken in the same file, which is the strongest argument for the class-level test I suggest inline over a second single-target test.

Nothing else in the diff concerns me — the rename is mechanical and complete for `error` (8 call sites, all converted), and no other helper shadows a builtin.

---

PR #489, opened 2026-08-02
Fix procurement award identity and plan remaining P1 work

hollomancer:
## What changed

- Adds the four-phase remediation plan covering award identity, cold-start safety, source
  normalization/provenance, and ranking orientation/auditability.
- Introduces a versioned, award-grain key built from the full stable SBIR source lineage. Public
  tracking and contract identifiers remain available for evidence, but no longer control joins,
  deduplication, or candidate identity.
- Carries `prior_award_key` through S1/S2/S3 pairing, candidate/evidence artifacts, census
  validation, enrichment, and the monthly report ledger.
- Canonicalizes multiple SBIR.gov editions of one stable award and records edition counts.
- Makes legacy rollout fail closed when a public ID is ambiguous or conflicts with an explicit
  key; safe mixed keyed/keyless ledgers remain supported.
- Resolves each rendered transition path to its actual award, so **Built on** cannot silently use
  another award held by the same firm.

## Why

The report treated Agency Tracking Number as award grain even though it is reused. Contract number
is also reused. That collapsed candidate pairs and let awardee-first rendering attach the wrong
funded technology to a solicitation. The same conflation also affected descriptive enrichment and
the Phase III census. The root fix is to separate readable public lineage identifiers from the key
used for data identity.

## Impact and rollout

- The current 219,500-row local SBIR snapshot normalizes to 219,445 unique award keys. The 54
  repeated keys are published revisions of the same stable lineage and are retained as canonical
  records with edition counts (maximum three editions).
- Pre-migration normalized snapshots fail with an actionable error. Regenerate them from raw award
  snapshots when rolling this out.
- This draft intentionally implements only Phase 1. Phase 2 must add cache-miss/bootstrap bounds
  before Phase 3 expands matching coverage; Phase 4 restores deadline-primary ordering until the
  opportunity-to-firm model is production-validated.

## Verification

- `pytest -q -m fast tests/unit`: 4,222 passed, 7 skipped.
- Focused reporting/pairing/census/model/backtest/retrospective suite: 246 passed.
- Final changed-area regression suite: 98 passed.
- MyPy: 5 changed source modules passed.
- Ruff lint/format and `git diff --check`: passed.
- Full local SBIR snapshot: 219,445 normalized awards, 219,445 unique non-null keys.

## Known baseline issue

`tests/unit/phase_iii_census/test_import_isolation.py::test_census_import_and_freeze_verification_work_in_shallow_container_layout`
fails on unchanged `main`: its shallow fixture copies `sbir_analytics` but not the already-required
`sbir_etl.utils.award_identity` module. This PR does not broaden scope to repair that pre-existing
container-fixture issue.

### DIFF
--- a/sbir_etl/reporting/procurement_transition/core.py
+++ b/sbir_etl/reporting/procurement_transition/core.py
@@ -32,6 +32,8 @@ from sbir_etl.utils.procurement_text import (
 
 logger = logging.getLogger(__name__)
 
+_AWARD_KEY_VERSION = "sbir-source-v2"
+
 _ANNOTATION_FIELDS = (
     ("interest_alignment", "Mission-interest alignment"),
     ("technology_ecosystem", "Technology ecosystem"),
@@ -55,53 +57,100 @@ def _coalesce(df: pd.DataFrame, *names: str) -> pd.Series:
     return result
 
 
-# Fields that together identify one award. The public SBIR.gov snapshot reuses
-# Agency Tracking Numbers across tens of thousands of distinct award rows, so a
-# bare tracking number is a *firm/topic* key, not an award key: collapsing on it
-# makes unchanged rows look changed and silently merges distinct awards.
-_AWARD_GRAIN_FIELDS = (
-    "Company",
-    "Agency",
-    "Branch",
-    "Phase",
-    "Program",
-    "Proposal Award Date",
-    "Contract End Date",
-    "Award Title",
-    "Award Amount",
+# Stable source fields used to identify one award. Mutable report content such
+# as title, abstract, amount, and recorded end date deliberately stays out of
+# this key; changes to those fields belong in ``row_hash``.
+_AWARD_KEY_FIELDS = (
+    ("Agency Tracking Number", "agency_tracking_number"),
+    ("Contract", "contract_number"),
+    ("Agency", "agency"),
+    ("Branch", "branch", "sub_agency"),
+    ("Phase", "phase"),
+    ("Program", "program"),
+    ("Proposal Award Date", "award_date"),
+    ("Solicitation Number", "solicitation_number"),
 )
 
 
-def _natural_award_id(row: pd.Series) -> str | None:
-    for name in ("Agency Tracking Number", "Contract", "award_id"):
-        # ``_display`` rejects the nan/NaT/None/<NA> spellings ``str()`` would keep.
+def _row_value(row: pd.Series, *names: str) -> str | None:
+    for name in names:
         if (value := _display(row.get(name))) is not None:
             return value
     return None
+
+
+def _identity_component(value: Any) -> str:
+    return re.sub(r"\s+", " ", _display(value) or "").strip().upper()
+
+
+def _award_key_field_value(row: pd.Series, names: tuple[str, ...]) -> str | None:
+    value = _row_value(row, *names)
+    if names[0] != "Proposal Award Date" or value is None:
+        return value
+    text = value.strip()
+    match = re.match(r"^(\d{4})[-/](\d{1,2})[-/](\d{1,2})(?:\D|$)", text)
+    parts: tuple[str, str, str] | None
+    if match is None:
+        match = re.match(r"^(\d{1,2})/(\d{1,2})/(\d{4})(?:\D|$)", text)
+        parts = (match.group(3), match.group(1), match.group(2)) if match else None
+    else:
+        parts = (match.group(1), match.group(2), match.group(3))
+    if parts is None:
+        return value
+    try:
+        year, month, day = (int(part) for part in parts)
+        return date(year, month, day).isoformat()
+    except ValueError:
+        return value
+
+
+def _natural_award_id(row: pd.Series) -> str | None:
+    # Preserve the report's existing public lineage identifier for readable
+    # output and legacy candidate files. It is never used as award grain: the
+    # separate ``award_key`` owns joins, deduplication, and candidate identity.
+    return _row_value(
+        row,
+        "Agency Tracking Number",
+        "agency_tracking_number",
+        "Contract",
+        "contract_number",
+        "award_id",
+    )
 
 
 def _stable_award_id(row: pd.Series) -> str:
     natural = _natural_award_id(row)
     if natural is not None:
         return natural
-    material = "|".join(_display(row.get(name)) or "" for name in _AWARD_GRAIN_FIELDS)
-    return "sbir-" + hashlib.sha256(material.encode()).hexdigest()[:20]
+    return "sbir-" + _award_grain_key(row)[:20]
 
 
 def _award_grain_key(row: pd.Series) -> str:
     """Identity of one *award*, not one firm/topic.
 
-    ``award_id`` stays the natural identifier so it keeps joining to the
-    candidate frame, but it is not award-grain on its own — hence this separate
-    key for snapshot comparison.
+    ``award_id`` stays the natural identifier for readable lineage. It is not
+    award-grain on its own, so joins, deduplication, and snapshot comparison use
+    this separate stable source compound.
     """
 
-    material = "|".join(
-        [
-            _natural_award_id(row) or "",
-            *(_display(row.get(name)) or "" for name in _AWARD_GRAIN_FIELDS),
-        ]
+    if existing := _row_value(row, "award_key"):
+        if _row_value(row, "award_key_version") != _AWARD_KEY_VERSION:
+            raise ValueError(
+                "pre-migration award_key cannot be reused; regenerate identity from raw awards"
+            )
+        return existing
+
+    recipient = (
+        _row_value(row, "UEI", "uei", "recipient_uei")
+        or _row_value(row, "Duns", "duns", "recipient_duns")
+        or _row_value(row, "Company", "company", "recipient_name")
     )
+    components = (
+        "sbir",
+        recipient,
+        *(_award_key_field_value(row, names) for names in _AWARD_KEY_FIELDS),
+    )
+    material = "|".join(_identity_component(value) for value in components)
     return hashlib.sha256(material.encode()).hexdigest()[:24]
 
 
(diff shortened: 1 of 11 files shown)


sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -55,53 +57,100 @@ def _coalesce(df: pd.DataFrame, *names: str) -> pd.Series:
...
-    "Proposal Award Date",
-    "Contract End Date",
-    "Award Title",
-    "Award Amount",
+# Stable source fields used to identify one award. Mutable report content such
+# as title, abstract, amount, and recorded end date deliberately stays out of
+# this key; changes to those fields belong in ``row_hash``.
+_AWARD_KEY_FIELDS = (
+    ("Agency Tracking Number", "agency_tracking_number"),
+    ("Contract", "contract_number"),
```
hollomancer, line 65: **`Contract` and `Solicitation Number` are not stable across snapshots — including them reintroduces the exact bug this key exists to kill.**

The comment above this tuple draws the right line: mutable *report content* (title, abstract, amount, end date) stays out of the key. But the criterion that matters for a month-over-month diff isn't "content vs. identity," it's **"is this field ever populated later than the row first appears?"** — because if it is, the key changes and one award becomes a `disappeared` + `newly_observed` pair. `Contract` is the field most prone to exactly that: SBIR.gov publishes rows with a blank contract number and backfills it once the contract is executed. `Solicitation Number` has the same exposure on early-appearing rows.

Confirmed at `6800a9a` — same award, contract number backfilled, nothing else changed:

```
KEY no-contract: b1093c651b528012ab6c7a6b
KEY w/ contract: 1367da82cd6a16a5b92cf09e   SAME? False
```

and end to end through the cohort diff:

```python
prev = normalize_awards(pd.DataFrame([row("Acme", "T-1", contract="")]))
curr = normalize_awards(pd.DataFrame([row("Acme", "T-1", contract="FA8650-24-C-1234")]))
build_award_cohorts(curr, prev, report_month="2026-01")
# [{'award_id': 'T-1', 'newly_observed': True}]
```

A month where a batch of contract numbers lands is a month where the packet reports a wave of phantom new awards, and the prior snapshot's rows vanish. That's the same failure the old `_AWARD_GRAIN_FIELDS` key had (via `Award Amount` / `Contract End Date`), just moved to a less obvious field, so the fix looks complete while the class of bug survives.

Suggested resolution: treat a backfillable identifier as *evidence for* identity, not *part of* it. Build the key from the fields that are present-or-never at first publication — `Agency Tracking Number`, recipient, `Agency`/`Branch`, `Phase`, `Program`, `Proposal Award Date` — and keep `contract_number` / `solicitation_number` as carried columns (they're already emitted in `normalize_awards`, and `_cited_award_identifier` reads them from there, so nothing downstream needs them in the hash). If a contract number is genuinely load-bearing for disambiguating two awards that otherwise collide, that argues for a documented tie-break at *match* time, not for hashing a field that changes under you.

Whichever way you resolve it, this wants a regression test in the same shape as the reproduction above — normalize the same award twice with a field backfilled, assert `award_key` is unchanged. The current tests cover duplicate-edition collapse within one snapshot but not identity stability *across* snapshots, which is the property the key is actually for.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -970,13 +1151,62 @@ def _master(
...
+                        "legacy prior_award_id is ambiguous; regenerate candidates with "
+                        f"prior_award_key (examples: {preview})"
+                    )
+                joined.append(
+                    legacy.merge(
+                        award_fields,
+                        left_on="prior_award_id",
+                        right_on="award_id",
+                        how="left",
+                        validate="many_to_one",
```
hollomancer, line 1194: **The curated ambiguity error above is unreachable in production — `validate="many_to_one"` fires first, with a worse message.**

The intent here is good: catch a reused `prior_award_id` and tell the operator to regenerate candidates with `prior_award_key`. But pandas' `many_to_one` validation checks uniqueness of the merge key across the **entire** right frame, not just the rows a left key matches. The explicit check on line 1181 is narrower — it intersects with `candidate_ids`, so it only fires for duplicate ids the candidates actually reference.

Since a duplicated `award_id` in `awards` is the *normal* state of this data (that's the premise of the whole PR — `public_id_award_count` exists precisely because tracking numbers are reused), any awards frame with a reused tracking number anywhere blows up here first. Confirmed at `6800a9a` with candidates referencing only the unambiguous `T-1`:

```
award_ids: ['T-1', 'T-DUP', 'T-DUP']
RAISED: MergeError Merge keys are not unique in right dataset; not a many-to-one merge
Duplicates in right:
 award_id
   T-DUP ...
```

The operator gets a pandas internals message naming an award they never referenced, instead of "regenerate candidates with `prior_award_key`." The narrower check is the one that would actually help, and it's dead.

Two ways out, either fine:
- Deduplicate the right frame to the referenced-and-unambiguous subset before merging (`award_fields[~award_fields["award_id"].isin(ambiguous)]`), so `many_to_one` holds by construction and the curated error stays the only one that can fire; or
- Drop `validate=` on the legacy branch and assert the row count is unchanged after the merge — the explicit check already covers the fan-out case it was guarding against.

Worth a test either way: today `tests/unit/reporting/test_procurement_transition_outputs.py` exercises the ambiguity path only with the ambiguous id *referenced* by a candidate, which is the one arrangement where the curated error still wins.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -159,8 +219,39 @@ def normalize_awards(raw: pd.DataFrame) -> pd.DataFrame:
...
+            out.assign(
+                _end_sort=pd.to_datetime(out["recorded_end_date"], errors="coerce"),
+                _amount_sort=pd.to_numeric(out["amount"], errors="coerce"),
+            )
+            .sort_values(
+                ["award_key", "_end_sort", "_amount_sort", "row_hash"],
+                kind="stable",
+                na_position="first",
+            )
+            .drop_duplicates("award_key", keep="last")
```
hollomancer, line 251: **The edition collapse silently reorders the whole frame, and only when duplicates happen to be present.**

`sort_values(["award_key", …])` sorts every row, not just the colliding ones, so `normalize_awards` returns rows in `award_key` hash order — which is arbitrary — rather than source order. And because the block is guarded by `if duplicate_editions.any()`, the ordering *depends on whether this month's snapshot happened to contain a duplicate edition*. Confirmed at `6800a9a`:

```
input order:      Zeta, Alpha, Alpha(dup), Mid
ORDER (dup):      ['Alpha Inc', 'Mid Inc', 'Zeta Inc']     # hash order
ORDER (no dup):   ['Zeta Inc', 'Alpha Inc', 'Mid Inc']     # source order
```

For a module whose docstring is "Deterministic monthly procurement-center packet generation," that's the wrong kind of nondeterminism — it's deterministic given the input, but it flips between two orderings based on an incidental property of the snapshot. Any artifact that inherits this order (award cohort CSVs, anything diffed month over month for review) shows a whole-file reshuffle in the month a duplicate first appears, burying the real changes.

The fix is one line — restore the original order after collapsing:

```python
out = (
    out.assign(_source_order=range(len(out)), _end_sort=…, _amount_sort=…)
       .sort_values([...], kind="stable", na_position="first")
       .drop_duplicates("award_key", keep="last")
       .sort_values("_source_order", kind="stable")
       .drop(columns=["_source_order", "_end_sort", "_amount_sort"])
)
```

Same pattern as `_candidate_order` in `_master` below, which already gets this right — worth being consistent.

While here: `keep="last"` after sorting on `["award_key", _end_sort, _amount_sort, row_hash]` with `na_position="first"` picks the latest end date, breaking ties on the largest amount. That's a defensible "most current-looking edition" rule and the comment says so, but the *reason* end-date-then-amount is the right precedence isn't stated, and it's the kind of choice someone will otherwise flip on instinct. One clause would fix it.

---

hollomancer (commented):
## Review — the diagnosis is right, one field choice undoes it

The central move here is correct and overdue. Splitting readable lineage (`award_id`, still the public tracking number) from join/dedup identity (`award_key`, a compound hash) is the right shape, and the surrounding hardening is genuinely good work: `validate="many_to_one"` on the joins, `_award_for_entry` resolving each candidate path to its own award instead of reusing `group["award"]` for every row, `_cited_award_identifier` widening to contract and tracking number, the `award_key_version` fence that refuses pre-migration snapshots rather than silently mixing key generations. The built-column now varies per award in the summary table, which was a real attribution bug. `tests/unit/reporting/test_procurement_transition_outputs.py` + the phase_iii suites: **163 passed**.

Three findings, all verified against `6800a9a` with reproductions inline.

**1. `Contract` / `Solicitation Number` in `_AWARD_KEY_FIELDS` reintroduce the instability the key exists to prevent (blocking).** The key is supposed to be stable across snapshots, but SBIR.gov backfills contract numbers onto rows that first publish blank. Backfill one and the same award hashes differently, so the cohort diff reports it as `newly_observed` while the prior snapshot's row disappears — the same class of bug the old `_AWARD_GRAIN_FIELDS` had via `Award Amount`, relocated to a field where it's less visible. The distinction that matters isn't content-vs-identity, it's *can this field arrive late*. Reproduction and a suggested split in the line comment.

**2. The curated ambiguity `ValueError` in `_master` is unreachable.** `many_to_one` validates uniqueness across the whole right frame; the explicit check only covers ids the candidates reference. Since duplicated `award_id` is the normal state of this data, operators get a pandas `MergeError` naming an award they never referenced instead of "regenerate candidates with `prior_award_key`."

**3. `normalize_awards` reorders every row into hash order, but only in months where a duplicate edition exists.** One-line fix, and `_master`'s own `_candidate_order` already demonstrates the pattern.

Smaller things, none blocking: `_award_key_field_value` branches on `names[0] != "Proposal Award Date"`, which couples date parsing to a literal in a tuple that's likely to grow — a flag in the tuple or a small per-field normalizer map would survive edits better. And `row_hash` is now computed over a frame that includes `award_key_version`, so a future version bump marks every award content-changed; harmless while `_prepare` hard-rejects mismatched versions, but worth a comment so the coupling is deliberate rather than incidental.

Finding 1 is the one I'd hold the PR on — 2 and 3 are contained and mechanical.

---

PR #490, opened 2026-08-02
Document live Mac mini deployment contract

hollomancer:
## Summary

- add a root `AGENTS.md` entrypoint so Codex and other agents discover the live-deployment rules
- point both agent entrypoints to the canonical Mac mini runbook
- document the live checkout, release baseline, persistent storage, secrets, Dagster metadata, and Tailscale-only ingress
- record the initial materialized state and explicitly distinguish the bounded Neo4j seed from a full-universe graph

## Why

The live stack operates from a clean checkout separate from the active development checkout, but that identity was not present in tracked agent guidance. An agent could otherwise run deployment commands from the dirty development tree, replace local secrets, remove persistent volumes, enable public ingress, or overstate the completeness of the seeded graph.

## Operational state recorded

- deployed baseline: tag `0.1`, commit `054bd862`
- persistent data: `/Volumes/SSDmini/sbir-analytics`
- full-source `2026-Q3` drone-manufacturing snapshot
- bounded 25,000-record Neo4j validation seed, with broad schedules still disabled

## Validation

- `git diff --check`
- live Dagster materialization run `70e7d2f8-2a90-453f-a2a9-ae9bd4abc9c2` succeeded
- API snapshot and graph freshness endpoints verified
- all five server containers healthy after the initial Neo4j backup

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,21 @@
+# Agent Instructions
+
+Read [CLAUDE.md](CLAUDE.md) for the repository's project conventions, testing
+requirements, code standards, and scope rules. Those instructions apply to all
+coding agents; tool-specific agent definitions may differ by runtime.
+
+## Live deployment
+
+Before any deployment, server operation, or live Dagster materialization, read
+[the Mac mini runbook](docs/deployment/mac-mini-server.md#live-instance-on-this-mac-mini).
+
+- The only live checkout is `/Users/conradhollomon/projects/sbir-analytics-server`.
+  Never operate the live stack from the development checkout.
+- Preserve `.env.server`, `/Volumes/SSDmini/sbir-analytics`, and the Docker
+  `dagster_home` volume.
+- Ingress must remain Tailscale Serve over tailnet-only HTTPS. Never enable
+  Funnel or expose server ports to the LAN or public internet.
+- Treat materialization as a live-data mutation. Confirm the SSD is mounted,
+  the deployment checkout is clean, and the stack is healthy before running it.
+- Keep schedules disabled until their jobs have completed successfully by hand
+  with the inputs available on this host.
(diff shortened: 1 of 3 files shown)


hollomancer (commented):
## Review — right instinct, one duplication worth collapsing

The problem this solves is real and the framing is right: an agent with no tracked record of the live checkout will happily run `make server-*` from the dirty development tree, and every guardrail here (preserve `.env.server`, preserve the SSD path and the `dagster_home` volume, never `down -v`, Tailscale Serve only, never Funnel) names a specific irreversible action rather than gesturing at caution. The bounded-seed paragraph is the best thing in the diff — recording that Neo4j holds 24,794 rows from the first 25,000 source records **and** that it must not be represented as a full-universe graph closes off a misreading that would otherwise be very easy to make from the API alone.

Two notes, neither blocking.

**1. The live-deployment block is now duplicated verbatim in `AGENTS.md` and `CLAUDE.md`, and both are summaries of the runbook — three copies of the same rules.** They agree today. They will not agree after the first edit, and the failure mode is quiet: an agent reads whichever file its runtime loads, which may be the stale one, and the rules being duplicated are exactly the ones where being wrong destroys data. Since both files already link to `docs/deployment/mac-mini-server.md#live-instance-on-this-mac-mini`, I'd cut both summaries down to the pointer plus the one-line reason to follow it, and let the runbook be the single source. If you'd rather keep the inlined bullets for agents that don't follow links — a fair call, and probably the reason they're there — then say in both files that the runbook wins on conflict, so a drifted copy resolves instead of getting obeyed.

**2. `CLAUDE.md` is loaded in every session on every machine, including CI and cloud sessions where `/Users/conradhollomon/...` does not exist.** That's cheap enough to be fine, but the current phrasing ("The only live checkout is …") reads as a global invariant rather than a fact about one host. Scoping it — "on the Mac mini host, the only live checkout is …" — costs three words and stops an agent elsewhere from trying to reason about a path it can't see.

Smaller: the dated facts (baseline tag `0.1` / commit `054bd862`, the 2026-08-02 materialized state) will rot, and nothing in the repo can detect it. The runbook already handles this the right way for the baseline by telling the reader to verify with `git describe --tags --always --dirty` rather than trust the number — worth extending the same "verify, don't trust this number" note to the **Materialized state** section, which currently reads as current fact with no verification path.

Docs-only, no CI risk. Fine to merge once the duplication question is settled.

---

PR #493, opened 2026-08-02
Enforce architecture and evidence boundaries

hollomancer:
## Summary
- enforce the existing first-party dependency direction in CI and remove the sbir_etl to sbir_ml back-edge through dependency injection
- add versioned, machine-validated study contracts and enforce the Phase III census materialization gate at runtime before source reads or writes
- refresh the census contract to frozen revision R7 and its current authorized-but-non-citable production status
- require procurement-transition callers to choose an explicit fusion scorer, preventing silent production degradation
- prevent operational code and deployment runbooks from depending on archived scripts; promote the live vendor-filter builder into scripts/usaspending

## Architectural intent
This preserves the working package layers instead of rewriting them. Evidence maturity is modeled independently from operational materialization authorization. One exact transitional exception records the five server download jobs that still wrap existing CLI implementations; it blocks any additional package-to-scripts dependencies and identifies the next migration boundary. Identity consolidation remains a separate, output-sensitive change.

## Verification
- full unit suite: 5,393 passed, 7 skipped
- focused architecture, evidence, census, and reporting suite: 93 passed
- Ruff lint and format checks passed for changed files
- architecture, repository hygiene, removed-source-reference, and study-manifest guards passed


studies/phase-iii-census/study.yaml
```diff
@@ -0,0 +1,34 @@
...
+    sha256: f9cdf80cf94c75e2e4475ca6f736e2e0fab27a8e8d19650a7841b11b17801605
+implementation:
+  - path: packages/sbir-analytics/sbir_analytics/assets/phase_iii_census/assets.py
+    symbol: phase_iii_census
+identity_policy:
+  strategy: exact normalized UEI only; no name or DUNS fallback
+  version: phase-0-r4
+  negative_evidence_allowed: false
+materialization:
+  allowed: false
```
hollomancer: **Nothing blocks materialization. `allowed: false` is a declaration that no code reads.**

The PR summary says this registers the census "at its actual reproducible status **with materialization blocked**." Checked the whole branch:

```
git grep -n "study_manifest\|study.yaml\|studies/" -- 'packages/*' 'sbir_etl/*'
```

returns nothing outside `sbir_etl/quality/study_manifest.py` itself. `load_study_manifest` has exactly one caller: `scripts/ci/validate_study_manifests.py`. The `phase_iii_census` asset never consults the manifest, so `dagster asset materialize --select phase_iii_census` runs today exactly as it did before this PR, gate or no gate.

What `validate_study_manifests.py` actually enforces is *internal consistency of the YAML*: frozen-artifact hashes match, implementation symbols exist, `study_id` matches the directory, and `MaterializationGate.require_blocker_when_closed` ensures a closed gate names a blocker. All worthwhile. None of it is a gate on materialization — it's a gate on the manifest being well-formed.

That gap is worth closing here rather than later, for the reason the PR itself argues: the value of "evidence maturity as an orthogonal governance contract" is that it can't be bypassed by someone who didn't read the design doc. A contract that only validates itself is a well-formed README. The person most likely to materialize a blocked study is precisely the person who doesn't know the manifest exists.

The enforcement is small — the asset resolves its own `study_id` and raises when the gate is closed:

```python
manifest = load_study_manifest(REPOSITORY_ROOT / "studies" / "phase-iii-census" / "study.yaml")
if not manifest.materialization.allowed:
    raise Failure(f"materialization blocked: {'; '.join(manifest.materialization.blockers)}")
```

That turns `blockers: ["Phase 2 negative-control eligibility is unresolved."]` from a note into the actual error text an operator sees at the moment they try — which is where it does the most good, and it keeps the blocker list honest, since a stale blocker becomes visible friction rather than dormant prose.

Two smaller things while this file is open:

- `evidence_status: reproducible` sits next to `allowed: false`, and the ladder in `EvidenceStatus` doesn't relate the two. Worth stating somewhere whether `reproducible` is *expected* to have an open gate — otherwise the next study author has to guess whether these two fields are independent or ordered.
- The blocker names #478's unresolved eligibility work. Given that #478's own release sentinel is deliberately red and its exit condition just moved (the contamination bound is still unbuilt), this gate is not closing soon. That's the correct state, and it's an argument for enforcement rather than against it — a long-lived gate is exactly the kind that gets forgotten.

---

sbir_etl/reporting/procurement_transition/core.py
```diff
@@ -891,7 +893,12 @@ def _unranked(frame: pd.DataFrame, reason: str) -> pd.DataFrame:
     return rows
 
 
-def group_candidates_by_awardee(rows: pd.DataFrame, awards: pd.DataFrame) -> list[dict[str, Any]]:
+def group_candidates_by_awardee(
+    rows: pd.DataFrame,
+    awards: pd.DataFrame,
+    *,
+    fusion_scorer: FusionScorer | None = None,
```
hollomancer: **The DI is the right call, but `= None` makes the unwired case silent, and no test pins the one place the wiring actually happens.**

Removing the `sbir_etl → sbir_ml` back-edge by injection is clearly better than the lazy import it replaces. The concern is the default.

Before, an unranked report required a genuine failure — `sbir_ml` had to be missing or broken, and the `except Exception` path logged why. Now `fusion_scorer=None` is the *default*, and the degraded path is entered by omission:

```python
if fusion_scorer is None:
    return _unranked(rows, "no fusion scorer was configured")
```

Forgetting a keyword argument now produces a complete, plausible-looking report with candidates in deadline order instead of fusion-ranked order, and a `WARNING` in the log. Deadline order is a defensible ordering; nothing about the output announces that the ranker never ran. That's the failure mode you least want to be reachable by omission — a report that looks finished but isn't ranked.

The wiring is a single line in a single caller. `git grep` on this branch finds exactly one production binding:

```
scripts/data/monthly_procurement_transition_report.py:59:        fusion_scorer=score_pairs_with_fusion,
```

No test asserts it. Delete that line and every test in the suite still passes — `tests/unit/reporting/test_procurement_transition.py` and `test_procurement_transition_outputs.py` call `MonthlyReportBuilder(...)` and `group_candidates_by_awardee(rows, awards)` without a scorer in ~15 places, and after this PR those all silently exercise the unranked path. The three tests that *do* care were correctly updated to pass `score_pairs_with_fusion` explicitly, which is right — but it means the production wiring is now the only thing standing between "ranked" and "not ranked," and it is the one thing nothing covers.

Two options, either fine:

1. **Make it required.** Drop the `= None` default on `group_candidates_by_awardee` and `MonthlyReportBuilder.__init__` so a missing scorer is a `TypeError` at the call site rather than a quality downgrade at render time. Callers that genuinely want no ranking pass an explicit `None` and say so. This is the smaller change and it moves the failure to where it's cheapest.
2. **Pin the wiring.** A test that imports the script's entry point and asserts a scorer is bound. Keeps the ergonomic default, costs one test.

(1) fits this PR's own thesis better: the whole point is making an implicit dependency explicit, and an implicit `None` reintroduces exactly the kind of silent behavior the architecture check exists to prevent.

Also worth noting the layering result: after this change, the only place the real scorer is bound to the real report is a file under `scripts/`. `check_architecture_boundaries.py` correctly permits that direction, so it isn't a violation — but the production composition root for this feature now lives in `scripts/`, which is a slightly odd place for it given the PR's framing about operational code and archived scripts. Not a defect, just worth being deliberate about.

---

scripts/ci/validate_study_manifests.py
```diff
@@ -0,0 +1,123 @@
...
+def _repository_path(relative: str, repository_root: Path) -> Path:
+    candidate = (repository_root / relative).resolve()
+    try:
+        candidate.relative_to(repository_root.resolve())
+    except ValueError as exc:
+        raise ValueError(f"path escapes repository root: {relative}") from exc
+    return candidate
+
+
+def _module_symbols(path: Path) -> set[str]:
```
hollomancer, line 26: `_module_symbols` walks `tree.body` only, so it sees top-level `def`/`class` and nothing else.

That's correct for the one manifest that exists — `phase_iii_census` is a decorated top-level function, and a decorator doesn't change the node type. Two shapes that are entirely normal in this codebase would produce a misleading failure:

- an asset built by a factory and bound at module level (`phase_iii_census = build_census_asset(...)`) — an `Assign`, not a `FunctionDef`
- a symbol re-exported through a package `__init__.py`, which is how several assets are reached today

In both cases the guard reports `implementation symbol 'x' is missing from <path>`, which reads as "the study's implementation was deleted" when the truth is "the guard only recognizes one of several equally valid definition forms." Given this check gates CI, the false positive costs someone a confused half hour on a refactor that broke nothing.

Cheapest fix is to also collect module-level assignment targets:

```python
elif isinstance(node, ast.Assign):
    names.update(t.id for t in node.targets if isinstance(t, ast.Name))
elif isinstance(node, ast.AnnAssign) and isinstance(node.target, ast.Name):
    names.add(node.target.id)
```

Or, if the strictness is deliberate, say so in the docstring and in the error message — "symbol must be a top-level function or class definition" — so the failure explains itself. Either is fine; silently rejecting valid forms is the part worth fixing.

Unrelated but adjacent: `main()` returns 1 when `studies/*/study.yaml` matches nothing. That's the right default for a repo that has manifests, and it means deleting the last study fails CI rather than passing vacuously. Worth keeping.

---

hollomancer (commented):
## Review of #493 at `6bd07d1`

No prior threads on this one, so this is a first pass.

The architecture check is the strongest part. `check_architecture_boundaries.py` encodes the layering as data (`ALLOWED_FIRST_PARTY_IMPORTS`) rather than prose, walks the full AST so function-level imports don't slip through, catches literal `importlib.import_module`/`__import__`, and treats `scripts.*` as forbidden from every package. Removing the `sbir_etl → sbir_ml` back-edge by injection rather than by rearranging packages is the right instinct — it fixes the edge without a rewrite.

**Three findings, inline:**

1. **The materialization gate is inert** (the significant one). `studies/phase-iii-census/study.yaml` declares `materialization.allowed: false`, but `load_study_manifest` has exactly one caller — the CI validator — and nothing in `packages/` or `sbir_etl/` reads the manifest. `phase_iii_census` materializes today exactly as before. What the validator enforces is that the YAML is well-formed, not that the gate holds. The PR summary describes this as "materialization blocked," which isn't yet true.
2. **`fusion_scorer: ... = None` makes the degraded path reachable by omission.** Previously an unranked report required a real failure; now forgetting a kwarg yields a complete-looking report in deadline order with only a WARNING. The single production binding lives in `scripts/data/monthly_procurement_transition_report.py:59` and no test covers it — delete that line and the suite stays green. Suggest dropping the default so a missing scorer is a `TypeError`.
3. **`_module_symbols` only recognizes top-level `def`/`class`.** Fine for the current manifest; produces a misleading "symbol is missing" failure for factory-assigned assets or `__init__` re-exports, both of which exist in this repo.

**One question, not a defect:** `ALLOWED_FIRST_PARTY_IMPORTS["sbir_graph"] = frozenset()` bars `sbir_graph` from importing `sbir_etl`. That's true of the code today or CI wouldn't pass, but it's stricter than the layering the rest of the table implies (`sbir_ml` may import `sbir_etl`; `sbir_analytics` may import all three). Worth confirming it's an intended constraint rather than a snapshot of current imports — the table is the policy now, and a graph loader eventually wanting a shared normalizer from `sbir_etl` seems plausible.

**Mergeability: not yet — `mergeable_state` is `dirty`.** This needs a merge from `main` before anything else; note that #492 touches `packages/sbir-analytics/sbir_analytics/assets/transition/contracts.py`, and the two PRs interact on the archived-scripts rule (flagged on #492's provenance-key thread — its Copilot suggestion to import from `scripts/archive/` would fail this PR's new check, so `sbir_etl` has to own that constant).

Once the conflict is resolved, none of the three findings above blocks the architecture work itself; (1) is the one I'd want fixed before merge, since it's the difference between the study contract being enforced and being documented.

---

studies/phase-iii-census/study.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - E1
+estimand: >-
+  Among exact-UEI Phase II award and federal prime contract action pairs in the
+  retrospective source universe, estimate uncoded post-completion follow-on procurement
+  proxies under every preregistered time-window and agency-continuity cell.
+frozen_artifacts:
+  - path: specs/phase-iii-census/design.md
+    sha256: b1cce1e75af840f13223ecfc543b6d801e63cc3329f65f7c0a5b9037c0e9fc48
+  - path: specs/phase-iii-census/amendments.md
+    sha256: f9cdf80cf94c75e2e4475ca6f736e2e0fab27a8e8d19650a7841b11b17801605
```
hollomancer: **These digests are `phase-0-r4` and the census has already moved to r5 in #497 and r7 in #499. The manifest's own CI check turns `main` red when either lands.**

`b1cce1e…fc48` and `f9cdf80…1605` are exactly the two values #497 replaces with `d18734e…54d7` / `df66830…ba51`; #499 then moves them again to `33b5918…6df9` / `24ead99…e5ab`. I recomputed all six from the respective branches — the census-chain hashes are correct on their branches, and this manifest is pinned to the state before them. `identity_policy.version: phase-0-r4` is stale in the same way.

`validate_study_manifests.py` enforces frozen-artifact hash equality and is a required check, so this isn't a soft inconsistency — whichever of {#493, #497} merges second breaks CI for every unrelated PR until someone repins. Since the census chain is already ordered (#497 → #499), the cheap resolution is for this branch to rebase and pin r7 once that chain settles.

Worth noting the failure would be *correct behavior*: a manifest that goes red when its frozen artifacts move is the check doing its job. The problem is only that nothing in either PR surfaces the coupling until merge.

**The more interesting collision is semantic.** `materialization.allowed: false` here, blocked on unresolved negative-control eligibility, versus #497's design status recording that the owner "superseded only the prior sequencing pause and authorized attempting production materialization." After both merge the repository states both positions — and per the open thread above, the `allowed: false` side is the one no code consults, while the "authorized" side is the one an operator acts on. So the governance file's only load-bearing effect would be to assert the opposite of live policy, which is a worse outcome than not having it.

That strengthens rather than weakens the case in the other thread for making the gate enforced. But it also needs a decision this PR can't defer: if Revision 5's authorization is real, the gate should read `allowed: true` with the negative-control and placebo constraints staying where they already are under `limitations` — they're accurate there and don't need a closed gate to hold. If the authorization is narrower than #497's status paragraph reads, then this manifest is the accurate statement and #497's label is what should soften. Either way the two files should be reconciled deliberately, not by merge order.

The `permitted_claims` / `limitations` split is the strongest part of this file — those three limitations are each falsifiable and each name a specific failure mode rather than gesturing at caution. That content survives whichever way the gate resolves.

---

hollomancer (commented):
Re-checked at `6bd07d1`. The three earlier threads are unchanged at this head and remain open as written — no new commits since they were posted, so nothing to re-verify there.

Adding one new finding inline: this branch pins the `phase-0-r4` census digests, and #497 (r5) and #499 (r7) both move them. `validate_study_manifests.py` hash-checks those artifacts as a required check, so the two open PRs cannot both merge without breaking `main`. Underneath the mechanical conflict, `materialization.allowed: false` and #497's "authorized attempting production materialization" are contradictory statements of the same policy — which is a sharper version of the enforcement question already raised in the `study.yaml` thread above.

---

scripts/ci/validate_study_manifests.py
hollomancer: Addressed at `c49101d`.

`_module_symbols` now also collects module-level `Assign` targets and `AnnAssign` targets, so a factory-built asset (`census = build_asset(...)`) and an annotated binding are both recognized. Definitions keep working unchanged, including decorated ones — a decorator doesn't change the node type, which is why `phase_iii_census` passed before.

Parametrized over all six forms (`def`, `async def`, `class`, assign, annotated assign, decorated) plus a negative case asserting a name bound only *inside* a function is still reported missing, so the walk didn't quietly become `ast.walk`. 11 tests pass in `tests/unit/quality/test_study_manifest.py`.

Left the `main()` return-1-on-no-manifests behavior alone, per your note that it's the right default.

---

sbir_etl/reporting/procurement_transition/core.py
hollomancer: Took option 2 at `c49101d` — pinned the wiring, kept the ergonomic default.

`tests/unit/reporting/test_procurement_transition_wiring.py` asserts the single production call site passes `fusion_scorer`, that it isn't literally `None`, and that the bound name is imported. Verified it's a real guard rather than a tautology: deleting the `fusion_scorer=score_pairs_with_fusion,` line makes it fail, and I restored the line after checking.

Chose option 2 over option 1 deliberately, and it's worth recording why since you argued option 1 fits the PR's thesis better. Making the parameter required is a `TypeError` at ~15 existing call sites in `test_procurement_transition.py` and `test_procurement_transition_outputs.py`, none of which care about ranking — so it converts a targeted gap into churn across two test files on a PR whose subject is architecture boundaries. The narrower guard covers the actual hole you identified: the one line standing between ranked and unranked.

Two things this does not resolve, both still yours to call:

- The degraded path is still reachable by omission everywhere except the guarded call site. If a second production consumer appears, nothing forces it to bind a scorer. Option 1 remains the durable fix and is a clean standalone change once the test call sites can be touched without noise.
- Your note about the composition root now living under `scripts/` stands untouched. The guard actually cements it — the test asserts *that script* is where the binding lives.

---

PR #494, opened 2026-08-02
Retire the AWS data plane: Mac mini hosts ingestion (Phases 1–2)

hollomancer:
## Description

Moves the project off AWS so the Mac mini handles ingestion, processing, and serving, with GitHub Actions reduced to pure CI. This PR covers Phases 1 and 2 of a six-phase plan; the remaining phases land separately.

The server profile was already AWS-free — `.env.server.example` declares no AWS variables, `docker-compose.server.yml` never references them, and `config/base.yaml` ships `use_s3_first: false` with the `transition_*_s3*` keys empty. Nothing being served today depended on AWS. The one load-bearing piece was **ingestion**: GitHub Actions fetched upstream data and used S3 as the handoff, and Actions runners cannot reach a tailnet-only host.

**Phase 1 — delete the AWS control plane** (~2,620 deletions, nothing consumed any of it):

- `infrastructure/cdk/` entire tree: S3 bucket, GitHub OIDC role, Batch compute environments and job queue, plus their tests
- `packages/sbir-analytics/sbir_analytics/lambda/weekly_refresh_handler.py` — confirmed orphan, no references outside its own directory
- `ci.yml`'s `infrastructure-validate` job, its `detect-changes` filter and output, and three references in the summary job
- AWS deployment/Batch/EC2 docs, with `docs/deployment/README.md` and `docs/index.md` repointed at the Mac mini

Deleting this code does not delete the live bucket, which is torn down in Phase 6 after its contents are preserved.

**Phase 2 — rehost ingestion on the mini.** All four downloaders write locally by default and mirror to S3 only when a bucket is passed explicitly (or `S3_BUCKET` is set), which keeps `data-refresh.yml` working until Phase 4 retires it:

- `download_sbir.py` — writes the canonical `award_data.csv` plus a dated vintage under `history/` with a metadata sidecar. SBIR.gov serves only the current snapshot, so this preserves the longitudinal record the S3 dated-key scheme provided; the sidecar's sha256 drives change detection.
- `download_sam_gov.py` — adds `--dest` and a local parquet writer. The partial-vs-canonical guard now applies to local output, so a short paginated fallback cannot overwrite a full dataset on disk.
- `download_uspto.py` — `--local` already existed and is now the default; the local file is kept rather than deleted after upload.
- `download_database.py` (USAspending) — new local path resuming via HTTP Range with progress checkpointed beside the partial file, replacing the S3 `.checkpoints/` prefix and multipart upload. Adds a free-space guard and handles a server that ignores `Range` by restarting rather than appending to a stale prefix.

Four Dagster jobs and schedules in `assets/jobs/source_downloads.py` carry `data-refresh.yml`'s cron times. All default to **STOPPED**, since the runbook requires a successful manual run on the host before enabling a schedule.

`docs/deployment/aws-decommission-plan.md` records the full plan and the two ordering constraints that prevent a refresh gap.

## Type of Change

- [x] Documentation update

Not a breaking change: the S3 paths remain functional and opt-in throughout this PR.

## Testing

- [x] All tests pass locally

```
uv sync --extra stack-dev
uv run pytest tests/unit -q          # 5236 passed, 7 skipped (was 5169)
uv run mypy sbir_etl                 # Success: no issues found in 206 source files
uv run ruff check sbir_etl packages/*/[a-z]* tests   # All checks passed
```

67 new tests. The scripts had no coverage before this PR:

- `test_download_sbir.py` (13) — vintage layout, sidecar contents, change detection including missing and corrupt sidecars, vintage selection, S3 staying opt-in
- `test_download_sam_gov.py` (10) — local parquet writer, sidecar, partial/canonical isolation, threshold boundary
- `test_usaspending_download_local.py` (27) — URL resolution and discovery, checkpoint roundtrip, free-space guard, Range resume, server-ignores-Range restart, short-download detection, checkpoint retained for resume
- `test_source_download_jobs.py` (17) — jobs discovered, crons match the retired workflow, schedules default to STOPPED

All 16 workflow and composite-action YAML files parse. Lint was diffed against the pre-change baseline in `scripts/` (13 errors before, 12 after), so no new debt was introduced; the remainder is pre-existing and outside CI's ruff scope.

Not verified here: the USAspending job against the real dump. It is tested against mocks, and the first live run on the mini is the actual proof. The free-space guard fails fast if the volume cannot take it.

## Checklist

## Notes for reviewers

One bug was introduced and caught during development, worth a look since it was a silent failure. Removing the module-level `boto3.client` from `download_database.py`, the first attempt used a module `__getattr__` to make it lazy. PEP 562 does not fire for bare global lookups inside functions, so `s3_client` was undefined — and the checkpoint save caught the resulting `NameError` and logged a warning, meaning the S3 path would have appeared to work while never checkpointing. Replaced with an explicit cached `_s3()` accessor and verified it reaches the client.

Remaining phases: 3 (strip S3 from library code), 4 (delete `data-refresh.yml` and `etl-pipeline.yml`, remove `setup-aws-credentials`), 5 (tests), 6 (drop `boto3` and tear down the AWS resources). Phase 6's teardown depends on preserving two prefixes that hold non-reproducible dated history — the `raw/awards/` vintage series and the dated `processed/phase_transition/` outputs. The rest of the bucket is a cache this PR makes re-fetchable.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KdgyfxJQHecaX9GKDgXjUV)_


scripts/usaspending/download_database.py
```diff
@@ -101,12 +114,225 @@ def save_checkpoint(
...
         print(f"Cleared checkpoint: s3://{s3_bucket}/{checkpoint_s3_key}")
     except Exception as e:
         print(f"Warning: Failed to clear checkpoint from S3: {e}")
 
 
+def resolve_source_url(
+    database_type: str = "full",
+    date_str: str = None,
+    source_url: str = None,
+) -> str:
```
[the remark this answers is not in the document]
hollomancer: Fixed in 3ee99ced: both optional parameters now use `str | None` in `resolve_source_url()` and `download_local()`.

docs/deployment/aws-decommission-plan.md
```diff
@@ -0,0 +1,212 @@
...
+Phase 2 lands, the mini fetches from the original upstreams, so `raw/sam_gov/`,
+`raw/usaspending/`, `raw/uspto/`, `raw/transition/`, and everything derived under
+`enriched/` and `processed/sbir/` are re-fetchable or re-computable. Copying them
+is optional insurance.
+
+Two prefixes hold dated history that **cannot** be rebuilt:
+
+| S3 prefix | Why it is irreplaceable | SSD destination |
+|---|---|---|
+| `raw/awards/{YYYY-MM-DD}/award_data.csv` | A vintage series of SBIR.gov snapshots, each carrying a `sha256` in object metadata (`download_sbir.py:74`). SBIR.gov serves only the current file; past vintages do not exist upstream. | `data/raw/sbir/history/` |
```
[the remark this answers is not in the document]
hollomancer: Fixed in 3ee99ced: the plan now points to the stable `scripts/data/download_sbir.py` path and former `_upload_to_s3()` symbol instead of a drifting line number.

scripts/usaspending/download_database.py
```diff
@@ -47,549 +41,231 @@
...
-        print(f"Loaded checkpoint from s3://{s3_bucket}/{checkpoint_s3_key}")
-        return checkpoint_data
-    except s3_client.exceptions.NoSuchKey:
+    if date_str:
+        return USASPENDING_DOWNLOADS[database_type].format(
+            base=USASPENDING_DB_BASE_URL, date=date_str
+        )
+
+    print("No date specified - searching for latest available file...")
+    latest_file = find_latest_available_file(database_type=database_type, s3_bucket=None)
```
hollomancer: **[P1] Remove this now-invalid keyword.** `find_latest_available_file()` no longer accepts `s3_bucket`, so the normal no-date/no-URL path raises `TypeError` before any download starts. `download_usaspending_op()` calls exactly that path. Please call the current signature and add an autospecced discovery test so future signature drift is caught.

scripts/usaspending/download_database.py
```diff
@@ -47,549 +41,231 @@
...
+    with session.get(source_url, headers=headers, stream=True, timeout=300) as response:
+        response.raise_for_status()
+        # A server ignoring Range returns 200 with the whole body; restart cleanly.
+        if resume_from and response.status_code == 200:
+            print("Server ignored Range request - restarting from byte 0")
+            dest_path.unlink(missing_ok=True)
+            resume_from = downloaded = 0
+
+        mode = "ab" if resume_from else "wb"
+        with open(dest_path, mode) as fh:
```
hollomancer: **[P1] Do not write an incomplete download under the final discoverable `.zip` name.** `find_latest_usaspending_dump()` rglobs these final names, so an in-progress or failed download can be selected and handed to `pg_restore`. Resume into a non-discoverable `*.part` path, validate it, then atomically rename it to the final name; add a test proving partials are ignored.

.github/workflows/monthly-analysis.yml
```diff
@@ -67,9 +51,6 @@ concurrency:
   group: monthly-analysis
   cancel-in-progress: true
 
-env:
-  AWS_REGION: us-east-2
-  S3_BUCKET: sbir-etl-prod-data
   PYTHON_VERSION: "3.11"
```
hollomancer, line 55: **[P1] Restore the top-level `env:` key.** Its removal leaves `PYTHON_VERSION` and `SKIP_NEO4J_LOADING` nested under `concurrency`, which only supports `group` and `cancel-in-progress`. The monthly workflow will be rejected/lose these variables when it runs. An `actionlint` check would catch this; the current YAML-parser check does not.

.env.server.example
```diff
@@ -67,10 +67,24 @@ SERVER_BACKUP_DIR=./backups
...
 # Enable the weekly core refresh ONLY after a manual run succeeds.
+#
+# Heavy assets (CET, fiscal, modernbert, USPTO AI extraction) are LOADED so
+# their jobs can be triggered by hand, but nothing schedules them. They used to
+# run on AWS Batch with more headroom than this host has; measure runtime and
+# memory on a manual run before automating any of them. Set this back to false
+# if the code server starts hitting its 3G limit.
 # -----------------------------------------------------------------------------
-DAGSTER_LOAD_HEAVY_ASSETS=false
+DAGSTER_LOAD_HEAVY_ASSETS=true
```
hollomancer, line 77: **[P1] Loading heavy assets by default also makes the documented “core” refresh heavy.** `core_refresh_job` still uses `AssetSelection.all()`; the old `DAGSTER_LOAD_HEAVY_ASSETS=false` gate was the only thing excluding CET/fiscal/ModernBERT/USPTO-heavy assets. Enabling `weekly_core_refresh` now schedules all of them despite the 3 GB warning. Keep this false or give the core job an explicit non-heavy selection and test it.

.env.server.example
```diff
@@ -67,10 +67,24 @@ SERVER_BACKUP_DIR=./backups
...
 SBIR_ETL__DAGSTER__SCHEDULES__DAILY_ALL_ASSETS_ENABLED=false
 SBIR_ETL__DAGSTER__SCHEDULES__WEEKLY_CORE_REFRESH_ENABLED=false
+
+# Pipeline chaining: each sensor runs a pipeline after its download job
+# succeeds. All default to STOPPED; enable after a manual pipeline run works.
+SBIR_ETL__DAGSTER__SENSORS__SBIR_PIPELINE_AFTER_DOWNLOAD_ENABLED=false
```
hollomancer, line 83: **[P1] These new controls—and the downloader credentials—never reach the server containers.** `docker-compose.server.yml` uses an explicit `x-server-environment` mapping but does not forward any of these sensor/phase-transition toggles, the new download-schedule toggles, `SAM_GOV_API_KEY`, or `USPTO_ODP_API_KEY`. `--env-file` only supplies Compose interpolation; it does not inject unreferenced variables. Add explicit mappings (or a deliberate service `env_file`) plus a configuration test.

packages/sbir-analytics/sbir_analytics/assets/jobs/phase_transition_archive.py
```diff
@@ -0,0 +1,83 @@
...
+ARCHIVED_OUTPUTS = (
+    "data/processed/phase_ii_awards.parquet",
+    "data/processed/phase_ii_awards.checks.json",
+    "data/processed/phase_iii_contracts.parquet",
+    "data/processed/phase_iii_contracts.checks.json",
+    "data/processed/phase_ii_iii_pairs.parquet",
+    "data/processed/phase_ii_iii_pairs.checks.json",
+    "data/processed/phase_transition_survival.parquet",
+    "data/processed/phase_transition_survival.checks.json",
+    "reports/phase_transition/phase_transition_report.json",
```
hollomancer, line 33: **[P1] The replacement schedule never generates this report.** The retired workflow ran `scripts/phase_transition_analysis.py`; `monthly_phase_transition` now runs only the four latency assets, then this archive job copies whatever already exists. Because missing files merely warn, a stale report can be archived beside fresh parquet outputs. Make report generation part of the scheduled Dagster job and require every archived output to come from the current run.

hollomancer (commented):
Reviewed the current 87-file head. I fixed and resolved the two pre-existing threads in `3ee99ced`; the inline comments below are new runtime/data-integrity blockers found in the broader review.

Current CI is still red: Package MyPy fails in `source_downloads.py`, and integration tests still pass the removed `use_s3_first` argument in two SAM.gov tests. The container check was pending when this review was submitted (the previous head failed because the notice-corpus manifest is excluded from the image). The PR description also still says “Phases 1–2” and reports the earlier scope/test results, while this head contains the full multi-phase, 87-file change. Please refresh the description and verification after addressing the blockers.

Phase-6 cleanup is also incomplete outside the inline anchors: both runtime Dockerfiles still install `boto3`/`cloudpathlib`, `config/prod.yaml` still sets `use_s3_first: true`, and `pyproject.toml` retains an empty AWS `cloud` extra.

packages/sbir-analytics/sbir_analytics/assets/jobs/source_downloads.py
```diff
@@ -0,0 +1,232 @@
...
+    )
+
+    api_key = os.environ.get("SAM_GOV_API_KEY", "")
+    if not api_key:
+        raise ValueError(
+            "SAM_GOV_API_KEY is not set. Obtain a key from "
+            "https://sam.gov -> Account -> API Keys and add it to .env.server."
+        )
+
+    df: pd.DataFrame = _download_bulk_extract(api_key)
```
hollomancer: **[P1] Keep the optional return type until the `None` check narrows it.** `_download_bulk_extract()` returns `DataFrame | None`, so assigning it to `pd.DataFrame` is the current `Package MyPy - sbir-analytics` failure. Remove the narrowing annotation or annotate the union.

packages/sbir-analytics/sbir_analytics/assets/jobs/source_downloads.py
```diff
@@ -0,0 +1,232 @@
...
+    context.log.info("Downloading AI patents dataset")
+    results["ai_patents"] = stream_download(USPTO_AI_PATENT_URL, ai_dest, session)
+    _guard_html_shell(ai_dest)
+    results["ai_patents"]["path"] = str(ai_dest)
+
+    # 3: assignments, which need a real browser session.
+    from scripts.data.download_uspto_browser import download_assignments
+
+    context.log.info("Downloading patent assignments via browser automation")
+    assignment_results = asyncio.run(download_assignments(output_dir=dest_dir / "assignments"))
```
hollomancer: **[P1] Complete the browser-to-pipeline contract before scheduling this job.** The deployed `Dockerfile` installs neither Playwright nor Chromium, so this call cannot run on the server. After those are installed, `download_assignments()` catches per-file exceptions and returns `{"error": ...}` entries, but this op still succeeds; its `*.csv.zip` outputs are also invisible to the downstream raw USPTO assets, which discover only `.csv`/`.dta`/`.parquet`. Install the runtime, fail on any required-file error, and atomically extract (or add ZIP-member support) before triggering validation; cover the image/handoff end to end.

packages/sbir-analytics/sbir_analytics/assets/sensors/source_download_chaining.py
```diff
@@ -0,0 +1,128 @@
...
+def _download_was_unchanged(context) -> bool:
+    """True when the monitored download reported no upstream change."""
+    try:
+        records = context.instance.all_logs(context.dagster_run.run_id)
+    except Exception:
+        return False
+    for record in records:
+        event = getattr(record, "dagster_event", None)
+        data = getattr(event, "event_specific_data", None) if event else None
+        metadata = getattr(getattr(data, "materialization", None), "metadata", None)
```
hollomancer: **[P1] This reads the wrong Dagster event shape, so unchanged downloads still trigger the pipeline.** `context.add_output_metadata()` on these plain ops emits metadata on a `STEP_OUTPUT` (`StepOutputData.metadata`), not on `event_specific_data.materialization.metadata`; this loop therefore always falls through to `False`. Read step-output metadata (or carry freshness in a run tag/output) and test `changed=false`. Apply the same freshness/status guard to USAspending so `status="skipped"` does not start enrichment.

sbir_etl/utils/cloud_storage.py
```diff
@@ -1,637 +1,219 @@
...
-    - Static: raw/sam_gov/sam_entity_records.parquet
-    - Dated: raw/sam_gov/sam_entity_records_YYYYMMDD.parquet
+    canonical = base / "sam_entity_records.parquet"
+    if canonical.is_file():
+        return str(canonical)
 
-    Args:
-        bucket: S3 bucket name (defaults to env var)
-        prefix: S3 prefix to search under
+    dated = _newest(list(base.glob("sam_entity_records_*.parquet")))
```
hollomancer: **[P1] Exclude the explicitly partial SAM.gov file from fallback discovery.** This glob matches `sam_entity_records_partial.parquet`; on a fresh host with no canonical file, that partial becomes the full enrichment source, defeating the downloader's overwrite guard. Restrict fallback names to dated complete extracts (or explicitly reject `*_partial.parquet`) and test the partial-only case returns no source.

scripts/data/download_sbir.py
```diff
@@ -66,73 +73,101 @@ def download_sbir_awards(s3_bucket: str) -> dict:
...
+
+    Returns a result dict whose ``changed`` flag is False when the download
+    matches the newest vintage.
+    """
+    data, file_hash = _fetch()
+
+    history_dir = dest / "history"
+    previous, latest_vintage = _previous_hash(history_dir)
+
+    if previous and previous == file_hash:
```
hollomancer, line 138: **[P2] Repair the canonical file before returning unchanged.** This decision uses only the latest vintage hash. If `award_data.csv` is missing or was truncated by the direct write below, every rerun still returns `changed=false` and discovery keeps preferring the bad canonical path. Verify/recreate the canonical from the fetched bytes before returning, and stage canonical/vintage writes followed by atomic replacement.

scripts/data/download_sam_gov.py
```diff
@@ -495,42 +502,39 @@ def _normalise_chunk(chunk: pd.DataFrame) -> pd.DataFrame:
...
+    print(f"\n💾 Writing {len(df):,} rows to {path}")
+    df.to_parquet(path, index=False, engine="pyarrow")
+
+    metadata = {
+        "source": "api.sam.gov",
+        "row_count": len(df),
+        "downloaded_at": datetime.now(UTC).isoformat(),
+        "partial": name == PARQUET_NAME_PARTIAL,
+    }
+    (dest / META_NAME).write_text(json.dumps(metadata, indent=2))
```
hollomancer: **[P2] A partial write must not overwrite the canonical dataset's metadata.** `_write_local(..., name=PARQUET_NAME_PARTIAL)` leaves the full parquet alone but always replaces `sam_entity_records.meta.json` with the partial row count/provenance. Use an output-specific sidecar or leave canonical metadata unchanged for partial writes, and extend the isolation test to cover metadata.

packages/sbir-analytics/sbir_analytics/assets/modernbert/embeddings.py
```diff
@@ -190,45 +188,33 @@ def modernbert_embeddings_patents(
...
-
-        bucket = get_s3_bucket_from_env()
-        if not bucket:
-            raise ValueError("S3 bucket not configured. Set SBIR_ANALYTICS_S3_BUCKET env var.")
+        from sbir_etl.utils.cloud_storage import get_data_root
 
-        # Default to latest PatentsView patent data
-        s3_url = f"s3://{bucket}/raw/uspto/patentsview/2025-11-18/patent.zip"
-        context.log.info(f"Loading from S3: {s3_url}")
+        archives = sorted((get_data_root() / "raw" / "uspto").rglob("patent*.zip"))
```
hollomancer: **[P2] Discover the exact PatentsView archive contract.** The migrated direct downloader writes `g_patent.tsv.zip`, which this `patent*.zip` glob misses, while the broad pattern can select an unrelated `patent_assignments.zip`. Accept the known PatentsView names (`g_patent.tsv.zip` and the Dagster job's `patentsview_patent.zip`), explicitly exclude assignment/AI archives, and test both downloader handoffs.

PR #484, opened 2026-08-02
Phase III enrichment → T3 firm-ranking + lineage features (Lever A): 0.47 was the wrong task

hollomancer:
## What this is

A **spec-first** follow-on to the Phase III ground-truth validation (#481). That PR measured *why* the fusion ranker underperforms; this one proposes the fix.

## Why

#481 established the ranker's ceiling is **empty candidate text, not the algorithm**: it matches words, and when a Phase III contract's whole FPDS description is "SBIR PHASE III AWARD." there's nothing to match. Measured: sensing (rich text) **0.60** vs logistics/DLA (near-empty) **0.21**, thin-desc **0.29**. T7's two highest-leverage fixes are both **data, not model**.

## What the spec proposes

1. **Enrich candidate text** — assemble real words per contract from SBIR topic descriptions (via the prior award's Topic Code), PSC/NAICS descriptions, award title, and recovered notice text — feeding the existing fusion text path in place of the bare FPDS description.
2. **Add description-independent features** — firm prior-award lineage, agency/topic continuity, timing gap, NAICS ancestry — signal that survives empty text.
3. **Re-score on the frozen #481 ground-truth set** (`score_t6.py`) for a **per-domain before/after lift** vs the 0.467 baseline, with CIs and the hard-decoy variant.

## The honest gate (T1)

The enrichment sources might be **as empty as what they replace** (Phase III sole-source awards often have no solicitation at all). So T1 is a cheap **source-coverage inventory** — measure per-source enrichable %, then a go/stop decision *before* building anything. Mirrors the step-0 check that saved effort in #481.

## Out of scope

Verified negatives / gating, forward-solicitation validation, and an embeddings bake-off — all documented as separate efforts.

## Success criterion

Sparse-description cells (logistics, thin-text) lift toward the rich-text band (~0.6) **without regressing rich-text cells**, reported per-domain with CIs. If they don't move, the spec says so.

> Depends on #481 (reuses its frozen ground-truth set + `score_t6.py`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #495, opened 2026-08-02
ER consolidation: one canonical firm_key, kill inlined normalize_name copies

hollomancer:
## What this is

The consolidation cleanup flagged during the merge review: the Phase III transition work re-inlined `normalize_name` in several scripts instead of reusing the core. This lands **one canonical firm-match key** and points the copies at it.

## Change

- **Add `firm_key`** to `sbir_etl.utils.text_normalization` = `normalize_name(remove_suffixes=True)` — the single "firm name → match key" definition (the ER foundation, research-questions **E2**).
- **Replace the inlined `normalize_name`** in `resolve_firm_awards` (#481; re-exported as `normalize_name` for its importers like `consolidate_selflabeled`) and `measure_firm_ranking` (#484).

## Why it's safe

- **Empirically resolve-decision-preserving**: 200/200 sampled firms on `award_data` resolve identically under the old inlined key and `firm_key` — they differ only in **case** (canonical is lowercase), and every comparison normalizes both sides, so matches are unchanged.
- Tests updated to the lowercase canonical; a `firm_key` test locks the behavior.
- CI-target checks green (ruff on `sbir_etl`+`tests`, `ruff format --check`, `mypy sbir_etl`). The `transition/detection` import errors are **pre-existing on main** (confirmed by stashing this change — they error either way), unrelated to this PR.

## Deliberately scoped out (documented follow-ups)

- **Promote `firm_from_notice`** (notice→firm parser) to a shared module — secondary duplication.
- **Integrate `resolve_firm` with the existing `company_fuzzy_matcher`** — the light script-level lookup vs the heavy DataFrame enricher is a design question, not a mechanical cleanup; a rushed merge there risks the fuzzy-match behavior.

### DIFF
--- a/sbir_etl/utils/text_normalization.py
+++ b/sbir_etl/utils/text_normalization.py
@@ -114,6 +114,24 @@ def normalize_company_name(name: str | None) -> str:
     return normalize_name(name, remove_suffixes=False)
 
 
+def firm_key(name: str | None) -> str:
+    """Canonical firm-match key: suffix-stripped, punctuation-normalized name.
+
+    The single definition for "normalize a firm/company name into a match key" used by
+    the Phase III transition work (firm→award resolution, notice→firm attribution). Equal
+    to ``normalize_name(name, remove_suffixes=True)``; empirically resolve-decision-
+    equivalent (200/200 on award_data) to the earlier ad-hoc uppercase variants, differing
+    only in case. Prefer this over re-deriving a normalizer inline.
+
+    Examples:
+        >>> firm_key("Acme Photonics, Inc.")
+        'acme photonics'
+        >>> firm_key("Progeny Systems Corp.")
+        'progeny systems'
+    """
+    return normalize_name(name, remove_suffixes=True)
+
+
 def pluralize_col_key(col: str) -> str:
     """Convert a column name to a pluralized dict key.
 
--- a/scripts/phase3_benchmark/measure_firm_ranking.py
+++ b/scripts/phase3_benchmark/measure_firm_ranking.py
@@ -29,12 +29,7 @@ import pandas as pd
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.metrics.pairwise import cosine_similarity
 
-_SUFFIX = re.compile(r"\b(INC|LLC|CORP\w*|CO|COMPANY|LTD|LP|LLP|PC|PLLC|INCORPORATED)\b\.?", re.I)
-_NON_ALNUM = re.compile(r"[^A-Z0-9 ]")
-
-
-def normalize_name(s: object) -> str:
-    return " ".join(_NON_ALNUM.sub(" ", _SUFFIX.sub(" ", str(s or "").upper())).split())
+from sbir_etl.utils.text_normalization import firm_key as normalize_name
 
 
 def firm_bucket(agency: str, branch: str) -> str:
--- a/scripts/phase3_groundtruth/resolve_firm_awards.py
+++ b/scripts/phase3_groundtruth/resolve_firm_awards.py
@@ -16,7 +16,6 @@ human can spot-check the fuzzy ones.
 from __future__ import annotations
 
 import argparse
-import re
 from dataclasses import dataclass, field
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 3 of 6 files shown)


sbir_etl/utils/text_normalization.py
```diff
@@ -114,6 +114,24 @@ def normalize_company_name(name: str | None) -> str:
...
+    equivalent (200/200 on award_data) to the earlier ad-hoc uppercase variants, differing
+    only in case. Prefer this over re-deriving a normalizer inline.
+
+    Examples:
+        >>> firm_key("Acme Photonics, Inc.")
+        'acme photonics'
+        >>> firm_key("Progeny Systems Corp.")
+        'progeny systems'
+    """
+    return normalize_name(name, remove_suffixes=True)
```
hollomancer, line 132: **[P1] Preserve the firm partitions produced by the replaced normalizers.** This delegates to the general normalizer, so it changes more than case: dotted `L.L.C.` remains `l l c`, `PC`/`PLLC` stop being stripped, while general tokens such as `partnership` are newly stripped. Against the current award data, 218 of 34,460 resolver keys differ; this splits Pelletized Straw LLC's five awards 4+1 and merges Engineering Partnership Ltd with Engineering Inc. The ranker likewise gets 14 split groups and 2 merged groups, changing abstracts, density, and candidate pools. Please use a compatibility-specific suffix grammar (including dotted legal forms) and add partition-level regression tests rather than relying on the 200-query decision sample.

hollomancer (commented):
The focused unit tests pass, but the consolidation is not behavior-compatible with the two normalizers it replaces. I left one blocking semantic comment with concrete current-data impact.

hollomancer (commented):
**Recommend closing this PR: #496 landed the same consolidation, and it already satisfies the open P1.**

I did not push fixes here, because fixing it would mean re-implementing what is now on `main`.

## The open P1 is resolved on main

The review comment asks for "a compatibility-specific suffix grammar (including dotted legal forms)" and partition-level evidence instead of a 200-query decision sample. That is precisely what #496's `organization-key-v1` profile is, and it was audited across **all 34,464** distinct names rather than sampled.

I ran the three cases from the comment against merged `main`:

| Input | `organization-key-v1` on main | Comment's concern under this PR |
|---|---|---|
| `Pelletized Straw L.L.C.` | `PELLETIZED STRAW` | 4+1 split — **does not occur** |
| `Pelletized Straw LLC` | `PELLETIZED STRAW` | joins correctly |
| `Engineering Partnership, Ltd` | `ENGINEERING PARTNERSHIP` | false merge — **does not occur** |
| `Engineering Inc` | `ENGINEERING` | stays distinct |
| `PC Photonics` | `PC PHOTONICS` | leading `PC` retained |
| `Smith PLLC` / `Acme PC` | `SMITH` / `ACME` | trailing designators stripped |

`_organization_key_v1` strips only *trailing* legal designators and normalizes dotted forms via `_DOTTED_DESIGNATORS`, so it neither splits punctuated variants nor swallows a mid-name `partnership`. #496's audit calls out these same three cases explicitly as deliberate outcomes.

## Both target callers are already migrated

The inlined `normalize_name` copies this PR set out to remove are gone from `main`:

- `scripts/phase3_groundtruth/resolve_firm_awards.py:35-38` → delegates to `normalize_company_name(..., profile=ORGANIZATION_KEY_V1)`, still re-exported as `normalize_name` for `consolidate_selflabeled`
- `scripts/phase3_benchmark/measure_firm_ranking.py:35-36` → same

So the stated goal — one canonical firm-match key, no inlined copies — is met, and #496 additionally kept per-consumer versioned profiles so each caller's partitions were preserved intentionally rather than incidentally.

## Mechanical blocker

`mergeable_state` is `dirty`. #496 rewrote `sbir_etl/utils/text_normalization.py` (‑61 lines, migrated onto the identity contract), which is the same file this PR adds `firm_key` to. Rebasing would mean reconciling `firm_key` against a module that no longer owns that responsibility.

## If you want a `firm_key` alias anyway

Adding it to `sbir_etl.identity` as a thin, documented alias for `ORGANIZATION_KEY_V1` would be reasonable — a short, memorable name for the ER foundation (research question **E2**) without a second normalization path. That is a small follow-up against current `main`, not a rebase of this branch.

The two documented follow-ups in the description — promoting `firm_from_notice`, and reconciling `resolve_firm` with `company_fuzzy_matcher` — are untouched by #496 and still worth tracking separately.

---

PR #496, opened 2026-08-02
Centralize company identity policies

hollomancer:
## Summary
- add one `sbir_etl.identity` contract for versioned company-name normalization profiles and 0..1 similarity scores
- migrate company-master, USAspending, transition vendor, Phase 0, UCC, SEC EDGAR, Form D, and Phase III research callers through compatibility shims
- consolidate USAspending and both Phase III paths on the durable `organization-key-v1` policy while retaining consumer-named legacy profiles
- prohibit unreviewed direct RapidFuzz scorers outside the shared identity contract, with reviewed exceptions for grant-number and person-name matching

## Evidence semantics
Name similarity remains candidate evidence, not proof of identity. It cannot support a negative claim such as never received an SBIR award without a separate study-specific coverage contract over identifiers, aliases, source universe, and data cut.

## Consolidation audit
Compared all 34,464 distinct company names in `award_data.csv` (SHA-256 `efdf7ca5a398703002ebb33345275b0f68e50af3c5db361d48a2456266a23628`).

The new policy:
- preserves 34,292 exact-name groups, the same count as `recipient-v1`
- removes the false `Engineering Inc` / `Engineering Partnership, Ltd` collision
- joins punctuated/unpunctuated `Pelletized Straw L.L.C.` variants
- adds the `Coherent Photonics` legal-name variants
- removes the false `PC Photonics` / `Photonics, Inc.` collision
- eliminates the ranking profile's broad `CORP\w*` behavior, which collided unrelated `Corp...` firms

Phase 0 remains on `entity-resolution-v1` because its normalized key participates in stable canonical IDs and needs an explicit ID migration.

## Verification
- 5,203 unit tests passed; 7 skipped
- 52 consolidation-focused identity/caller/guard tests passed
- MyPy passed: sbir_etl 208 files, sbir-ml + sbir-analytics 165 files
- Ruff lint and format checks passed for changed/CI surfaces
- company identity boundary and repository hygiene guards passed

## Relationship to PR #493
This PR is independent of #493 and can be reviewed separately. PR #493 establishes package and evidence boundaries; this PR addresses the output-sensitive identity primitive.

### DIFF
--- a/docs/steering/company-identity.md
+++ b/docs/steering/company-identity.md
@@ -1,0 +1,54 @@
+# Company identity contract
+
+Company identity is an evidence-producing decision, not generic string cleanup. Live
+callers use `sbir_etl.identity` for company-name normalization and similarity. The shared
+API has two guarantees:
+
+1. every normalization policy is explicitly named and versioned; and
+2. every similarity score uses a `0..1` scale, even when a compatibility adapter exposes
+   RapidFuzz's historical `0..100` convention.
+
+## Decision order
+
+Identity resolution remains identifier-first:
+
+1. exact validated UEI;
+2. exact validated CAGE or DUNS when the source and study permit it;
+3. exact name under the declared profile;
+4. fuzzy name candidate generation under declared thresholds; and
+5. human or study-specific adjudication where required.
+
+Name similarity alone must not establish a negative claim such as “this firm never
+received an SBIR award.” Absence claims require a study-specific coverage contract over
+the identifiers, aliases, source universe, and data cut being searched. The Phase III
+census study manifest therefore keeps `negative_evidence_allowed: false` until that
+contract exists.
+
+## Compatibility profiles
+
+The initial profiles preserve existing behavior so migration cannot silently change
+research outputs:
+
+| Profile | Current consumer |
+|---|---|
+| `matching-v1` | local company-master matching |
+| `recipient-v1` | USAspending recipient matching |
+| `entity-resolution-v1` | Phase 0 cross-source resolver |
+| `groundtruth-v1` | Phase III success-story award resolution |
+| `vendor-crosswalk-v1` | transition vendor crosswalk persistence |
+| `vendor-resolver-v1` | transition vendor resolver |
+| `form-d-join-v1` | agency/private-capital join keys |
+| `ucc-v1` | UCC debtor matching |
+| `sec-edgar-v1` | SEC company lookup and mention filtering |
+| `sec-edgar-trailing-v1` | single-suffix SEC mention/noise filtering |
+| `notice-key-v1` | Phase III notice attribution |
+| `phase3-ranking-v1` | Phase III firm-ranking benchmark |
+
+These profiles intentionally do not imply equivalent recall. Consolidating two profiles
+requires a versioned change, a before/after output comparison on the relevant corpus,
+and review of every changed auto-link and dropped candidate. Compatibility shims may keep
+old import paths temporarily, but the implementation belongs in `sbir_etl.identity`.
+
+Person-name matching, grant-number matching, and storage-only display cleanup are separate
+domains and do not use company-name profiles merely because their old helper was also
+called `normalize_name`.
--- a/docs/steering/company-identity.md
+++ b/docs/steering/company-identity.md
@@ -26,15 +26,20 @@ contract exists.
 
 ## Compatibility profiles
 
-The initial profiles preserve existing behavior so migration cannot silently change
-research outputs:
+`organization-key-v1` is the durable suffixless comparison key for ordinary company
+linkage. It folds case and accents, normalizes punctuation, and removes recognized legal
+designators only at the end of a name. Removing designator-like words from the middle of
+a name creates false collisions (for example, `PC Photonics` with `Photonics, Inc.`).
+
+The initial consumer-named profiles remain available as compatibility policies so
+migration cannot silently change research outputs:
 
 | Profile | Current consumer |
 |---|---|
 | `matching-v1` | local company-master matching |
-| `recipient-v1` | USAspending recipient matching |
+| `recipient-v1` | legacy USAspending recipient normalization |
 | `entity-resolution-v1` | Phase 0 cross-source resolver |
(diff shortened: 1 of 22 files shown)


hollomancer:
Additional equivalence audit: ran the original implementations from PR #493's pre-identity worktree and this branch against the same 11-name adversarial corpus across 11 consumers (shared text keep/remove, Phase 0, Form D, Phase III ground truth, Phase III ranking, notice attribution, vendor crosswalk, vendor resolver, UCC, and SEC). The sorted serialized result was byte-identical (2,158 bytes).

hollomancer (commented):
Reviewed the identity contract, the boundary guard, and the caller migrations. The consolidation audit and the byte-identical equivalence run across 11 consumers are convincing evidence for the behavior-preservation claim, and centralizing on one 0..1 contract is clearly the right shape.

Three findings, all verified by running the branch's code rather than by reading alone. Only the first is blocking in my view.

## 1. The boundary guard has two bypasses (`scripts/ci/check_identity_boundaries.py`)

`_uses_direct_rapidfuzz_scorer` inspects `ImportFrom.module` and `Import` alias names, which catches the two idiomatic forms but misses two others. Probing `scan_file` directly against the branch:

| Source | Caught |
|---|---|
| `from rapidfuzz import fuzz` | yes |
| `from rapidfuzz.distance import JaroWinkler` | yes |
| `from rapidfuzz import fuzz as f` | yes |
| `import rapidfuzz` then `rapidfuzz.fuzz.ratio(...)` | **no** |
| `from rapidfuzz import process` then `process.extractOne(...)` | **no** |

The second miss matters more than it looks: `process.extractOne` defaults to `fuzz.WRatio`, so that path is a direct scorer with no profile and no 0..1 normalization — exactly what the guard exists to prevent, reachable without touching `fuzz` or `rapidfuzz.distance` by name.

Both are cheap to close: also flag a bare `import rapidfuzz`, and treat `process` like `fuzz` in the `ImportFrom` branch. Worth adding the two cases to `tests/unit/scripts/test_identity_boundaries.py` so the guard's own coverage documents them.

## 2. Two `SUFFIX_TOKENS` entries are unreachable, and `ucc-v1` keeps a collision the PR fixes elsewhere

`SUFFIX_TOKENS` contains `'l.l.c'` and `'l l c'`, but `UCC_V1` collapses punctuation to spaces *before* filtering, then filters token-by-token. Neither multi-token entry can ever match a single token, so both are dead:

```
'Pelletized Straw L.L.C.'  ->  'pelletized straw l l c'
'Pelletized Straw LLC'     ->  'pelletized straw'
```

Those two do not join under `ucc-v1`. The PR description calls out joining exactly this punctuated/unpunctuated pair as a win for `organization-key-v1`, so it reads as an oversight that `ucc-v1` still splits them.

If preserving UCC recall exactly is deliberate under the compatibility-profile rule, that is a defensible call — but then the two dead entries should go, since they advertise handling that does not happen. If it is not deliberate, running the suffix filter before punctuation collapse fixes it.

## 3. The `ImportError` fallback is unreachable, and would change the metric if it ran

`rapidfuzz` is a hard runtime dependency (`pyproject.toml:28`, in `dependencies`, not an optional extra), so the `try/except ImportError` at the top of `company_names.py` and everything it guards — `_fallback_token_set`, the `SequenceMatcher` tail — cannot execute in any supported install.

If it ever did, `CompanyNameMetric.JARO_WINKLER` would fall through to `SequenceMatcher(...).ratio()`, a different algorithm, silently ignoring `prefix_weight`. A metric substitution is worse than an import error for a module whose whole purpose is making scoring semantics explicit.

I would drop the fallback and import `rapidfuzz` directly. If it stays, the JW branch should raise rather than degrade.

(I initially suspected `prefix_weight` was not plumbed through at all. That was my test string, not the code — `MARTHA`/`MARHTA` at `prefix_weight` 0.0 vs 0.25 matches `JaroWinkler.similarity` exactly. No issue there.)

## Notes, no action needed

- **CI is green.** All 22 check runs are success or skipped. The two "⚠️ Performance Check — No report generated" bot comments are **not caused by this PR**: on `main` the regression step is gated behind `steps.check_aws.outputs.aws_available == 'true'`, so without AWS credentials it skips while the summary step still posts. That is a pre-existing defect; #494 removes the gate so the benchmark always runs on the fixture.
- `_organization_key_v1` strips only *trailing* designators, so `PC Photonics` keeps its leading `PC`. Good — that is what makes the `PC Photonics` / `Photonics, Inc.` de-collision in the audit hold.
- Keeping Phase 0 on `entity-resolution-v1` pending an explicit ID migration is the right call; canonical IDs are not something to change as a side effect of this refactor.

---

hollomancer:
Correction to my review, and the guard fix is pushed as c6fe2fb.

**I was wrong about `process`.** My review claimed `from rapidfuzz import process` was an unintended bypass because `process.extractOne` defaults to `fuzz.WRatio`. It is not a bypass here — all four live callers (`company_fuzzy_matcher`, `usaspending/enricher`, `phase0/resolve_entities`, `phase3_groundtruth/resolve_firm_awards`) pass a scorer in from `sbir_etl.identity`, so `process` drives the search while the contract still owns the scoring. `test_plain_process_import_is_allowed` was a deliberate decision, not an oversight, and I should have read it before filing the finding. Left allowed in both import forms.

**The bare-import gap was real**, so that half stands. `import rapidfuzz` reaches `fuzz` and `distance` through attribute access without naming either submodule, which the AST check missed:

```python
import rapidfuzz
score = rapidfuzz.fuzz.token_set_ratio(a, b)   # previously not flagged
```

c6fe2fb flags the bare package import and `import rapidfuzz.fuzz` / `.distance`, and leaves `process` alone in both forms. Repo-wide scan stays at **0 violations**; full unit suite is **5,207 passed, 7 skipped**. Added regression tests for the bare import, the scorer-submodule form, and the allowed `process` submodule form, plus a comment recording why `process` is exempt so the distinction survives future edits.

Findings 2 (`ucc-v1` dead `SUFFIX_TOKENS` entries) and 3 (unreachable `ImportError` fallback) are untouched — both are judgment calls that belong to you, and neither blocks.

---

PR #497, opened 2026-08-02
Authorize Phase III census materialization

hollomancer:
## What changed

- append Revision 5 to the Phase III census approval log
- record the owner's pre-result decision to materialize Phase 1 while deferring unresolved control and placebo questions
- update the executable freeze revision and exact raw-byte SHA-256 values
- retain the existing fail-closed hash tests

## Scope

This is a sequencing amendment only. It does not change any inclusion clause, clause order, exact-UEI pair universe, source field, source snapshot, window endpoint, agency predicate, sensitivity cell, output metric, estimand, or diagnostic rule.

At approval and recording, source/provenance information and implementation status were visible. No exact-UEI pair count, cumulative drop-off count, six-cell sensitivity result, negative-control result, or placebo result had been materialized or seen.

## Validation

- exact freeze/hash tests: 10 passed
- complete Phase III census unit directory: 71 passed in a lockfile-synced Python 3.11 environment
- Ruff passed
- `git diff --check` passed

The census will be materialized only from this pinned commit after this pre-count record is published.


packages/sbir-analytics/sbir_analytics/assets/phase_iii_census/assets.py
```diff
@@ -92,7 +92,7 @@ def __init__(
 DATA_CUT_ENV = "SBIR_ETL__PHASE_III_CENSUS__DATA_CUT_DATE"
 SBIR_AWARDS_ENV = "SBIR_ETL__PHASE_TRANSITION__SBIR_AWARDS_PATH"
 PHASE_II_OUTPUT_ENV = "SBIR_ETL__PHASE_TRANSITION__PHASE_II_OUTPUT_PATH"
-FROZEN_SPEC_REVISION = "phase-0-r4"
+FROZEN_SPEC_REVISION = "phase-0-r5"
```
hollomancer, line 95: **This revision bump collides with #493, which pins the r4 digests in a manifest that CI hash-checks. Whichever merges second turns `main` red.**

Verified the freeze first, which is clean: `sha256(specs/phase-iii-census/design.md)` on this branch is `d18734e…54d7` and `sha256(specs/phase-iii-census/amendments.md)` is `df66830…ba51`, exactly the new constants below. Revision 5's "criteria impact: none" claim holds against the diff — the only design change is the status paragraph and the answerability label, and no clause, cell, window, predicate, estimand, metric, or source field moves. The r4 → r5 transition is internally consistent.

The problem is external to this branch. #493 adds `studies/phase-iii-census/study.yaml` carrying:

```yaml
frozen_artifacts:
  - path: specs/phase-iii-census/design.md
    sha256: b1cce1e75af840f13223ecfc543b6d801e63cc3329f65f7c0a5b9037c0e9fc48
  - path: specs/phase-iii-census/amendments.md
    sha256: f9cdf80cf94c75e2e4475ca6f736e2e0fab27a8e8d19650a7841b11b17801605
identity_policy:
  version: phase-0-r4
```

Those are precisely the two digests this PR replaces. `scripts/ci/validate_study_manifests.py` enforces frozen-artifact hash equality as a required check, so the manifest and the design doc cannot both be right after either PR lands. Both are open against `main`, neither references the other, and each is individually green. #499 then moves the same digests again to r7, so the drift is three revisions deep on one side and static on the other.

Two things follow, and they're separable:

1. **Sequencing.** Someone has to update `study.yaml`'s digests and `identity_policy.version` in whichever PR merges last. Cheapest is for #493 to rebase onto the final census revision and pin r7, since #493 is the newer artifact and the census chain (#497 → #499) is already ordered.

2. **The manifests disagree about the thing they exist to state.** More interesting than the hash mismatch: after both merge, `design.md` says the owner "authorized attempting production materialization," and `study.yaml` says `materialization.allowed: false` with blocker "Phase 2 negative-control eligibility is unresolved." Those are contradictory answers to the single question a reader most wants answered — and per the open thread on #493, the `allowed: false` side is the one nothing enforces, while the "authorized" side is the one an operator actually acts on. So the repo would ship a governance file whose only load-bearing effect is to state the opposite of current policy.

That's worth resolving deliberately rather than by merge order. Either the gate reflects the Revision 5 authorization (`allowed: true`, with the negative-control and placebo limits kept as `limitations`, which they already are), or Revision 5's authorization is narrower than the design status reads and the gate is the accurate statement — in which case this PR's answerability label is the thing to soften. The two files are currently written by different reasoning and only agree by accident of not having met.

One narrower note on the label itself: "production materialization authorized but not yet verified" is doing careful work, and the surrounding sentence correctly forecloses quoting a headline cell. Worth confirming that the post-write one-factor sensitivity check it defers to is enforced in the asset rather than expected of the operator — the label reads as if the check is a gate, and if it's procedural, saying so keeps this consistent with the honesty the rest of the amendment log maintains.

---

hollomancer (commented):
Reviewed at `2a77de5`. Small, well-formed change: a sequencing authorization recorded as Revision 5, the two frozen digests rebased onto it, and the test updated. I recomputed both hashes from the branch content and they match the new constants exactly, and the "criteria impact: none" claim holds — the design diff is confined to the status paragraph and the answerability label.

One comment inline. It isn't about this diff in isolation, which is fine, but about a collision with #493: that PR adds `studies/phase-iii-census/study.yaml` pinning the *old* r4 digests under a CI check that enforces hash equality, so whichever of the two merges second breaks the required `validate_study_manifests` check on `main`. Underneath the mechanical conflict, the two files also state opposite positions on whether the census may materialize, which is worth settling on purpose rather than by merge order.

---

PR #499, opened 2026-08-02
fix(census): reconcile exact-key supplementals

hollomancer:
## What changed

- freezes the approved exact-key multi-supplemental reconciliation rule before any census count exists
- reconciles one or more exact-key SBIR.gov supplementals into exactly one federal Phase II award
- preserves every authoritative federal field, fills only missing NAICS/PSC from unanimous normalized values, and fails closed on taxonomy conflicts or multiple federal matches
- records the quality-sweep consistency correction as append-only Revision 7 and pins both frozen-file digests

## Why

The first Phase II materialization stopped on source key `140D0420C0002`, where one federal award has two retained SBIR.gov source rows. The approved rule consumes all exact-key supplementals together without selecting by row order or using recipient identity as a heuristic.

The post-implementation quality sweep also found a stale paragraph that contradicted the approved rule. Revision 7 corrects only that text; it changes no criterion, pair universe, threshold, sensitivity cell, or implementation behavior.

## Validation

- 120 relevant unit tests passed on Python 3.11
- Ruff lint and format checks passed
- MyPy passed for both changed source files
- `git diff --check` passed
- dedicated test-fixer and quality-sweep audits completed

No census drop-off or sensitivity count had been materialized or seen when these commits were created. This draft is stacked on the Phase 1 sequencing PR; the census materialization will be pinned to the final commit in this branch.


packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_ii.py
```diff
@@ -414,12 +414,12 @@ def _unify(contract_phase_ii: pd.DataFrame, sbir_gov_phase_ii: pd.DataFrame) ->
     for source_key in shared_source_keys:
         federal_indexes = list(federal.index[federal_source_keys.eq(source_key)])
         supplemental_indexes = list(supplemental.index[supplemental_source_keys.eq(source_key)])
-        if len(federal_indexes) != 1 or len(supplemental_indexes) != 1:
+        if len(federal_indexes) != 1:
```
hollomancer, line 429: **Up to N SBIR.gov source rows are now consumed per federal award, and the output frame records that nowhere — even though it already carries the column pattern for exactly this.**

The relaxation itself checks out. The taxonomy loop below already iterated `supplemental.loc[supplemental_indexes, column]` over a list and unioned normalized nonblank values, so it was multi-supplemental-capable before this PR; dropping the `len(supplemental_indexes) != 1` half of the guard is genuinely the whole behavioral change, and `drop_supplemental.update(...)` was already set-based. Order independence isn't only asserted in the new test, it falls out of the structure — the union is over a set and the fill only fires when that set has one element. Good.

What's missing is the trace. Walk the `140D0420C0002` case the amendment cites:

- the federal row keeps `source_row_sha256 = None` (set for all contract rows at line 234),
- it keeps `phase_coding_reconciled = False` (line 270),
- both supplementals are dropped, taking their `source_row_sha256` values with them,
- `source_transaction_count` counts federal transactions only.

So an award that consumed two SBIR.gov source rows is, in the output, byte-identical to one that matched none — except possibly for a NAICS or PSC value that appeared from nowhere. The 19 keys from the audit are unrecoverable from the artifact.

That's an asymmetry worth closing, because the frame already solves this problem in the other direction. `source_transaction_count` exists precisely so that collapsing many FPDS transactions into one award grain is visible downstream rather than inferred. Supplemental consumption is the same kind of many-to-one collapse and gets no equivalent. A `supplemental_source_count`, or retention of the consumed `source_row_sha256` values, would cost one column and would make the audit reproducible from the artifact instead of from a one-time script.

This matters more than usual given where the frame is consumed. `_verify_phase_ii_provenance` in the census asset re-reads this parquet and asserts exact equality against the in-memory frame, so the census's provenance guarantee is "the frame is what was written" — it says nothing about how many source rows backed each row. Reconciliation is the one step in this path that discards evidence, and the one step with no counter.

Two smaller notes on the same block:

- The fill writes the **normalized** value: `federal.at[federal_index, column] = next(iter(supplemental_normalized))`, where `_normalize_source_key` strips and upper-cases. The new test pins this (`" ac12 "` → `"AC12"`). Federal-supplied values in the same column are stored raw. So `psc_code` carries two formatting conventions depending on provenance, with nothing marking which — the same missing-trace point in miniature, and here it's the *filled* rows that get the better-normalized form, which is the confusing direction.
- `_normalize_source_key` is now doing double duty as an award-identity normalizer and a taxonomy-value comparator. It happens to be correct for both, but a future change made for PIID normalization would silently alter which taxonomy values are judged conflicting *and* what gets persisted into `naics_code`/`psc_code`. Worth a distinct `_normalize_taxonomy_value` even if the body is identical today, so the two uses can diverge without one dragging the other.

---

hollomancer (commented):
Reviewed at `1984fb5`, against base `codex/phase-iii-census-r5-sequencing`.

Verified independently rather than taking the amendment's word for it:

- Both frozen digests match the branch content exactly — `sha256(specs/phase-iii-census/design.md) = 33b5918…6df9` and `sha256(specs/phase-iii-census/amendments.md) = 24ead99…e5ab`, equal to the constants at `assets.py:116-117`, with `FROZEN_SPEC_REVISION = "phase-0-r7"` matching the test. The freeze mechanism is intact through this revision.
- The code change is genuinely confined to the guard. The taxonomy union below already handled a list of supplementals, so this doesn't introduce multi-row merge logic — it stops rejecting the case the logic already covered. That's the smallest change that implements Revision 6, which is the right instinct for a frozen-criteria asset.
- Recording the incomplete-edit fix as its own Revision 7, with authority traced to the Revision 6 approval and no new policy claim, is the correct handling. The alternative — quietly folding it into r6 — would have been worse, and it costs nothing here since no result was materialized between them.

One substantive comment inline: the reconciliation now consumes an unbounded number of SBIR.gov source rows per federal award and leaves no trace of it in the output, in a frame that already carries `source_transaction_count` for the analogous federal-side collapse. Plus two smaller notes on normalized-vs-raw taxonomy values and the overloaded `_normalize_source_key`.

`_assert_prior_frames_equal` and its test are well built — pinning row reorder, column reorder, and value perturbation as things that must still fail is the right way to loosen a comparison predicate, and the null-canonicalization is correctly scoped to missingness only.

---

packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_ii.py
hollomancer: Partially addressed at `3ba4533` — leaving this open for the part that isn't.

**Done:** the taxonomy path now routes through `_normalize_taxonomy_value` instead of sharing `_normalize_source_key` with award-identity normalization. The body delegates and is behavior-identical, so the frozen digests are untouched; the point is only that a future change made for PIID matching can't silently move which taxonomy values are judged conflicting or what gets persisted into `naics_code`/`psc_code`. 129 tests across `tests/unit/phase_transition` and `tests/unit/phase_iii_census` pass, including the frozen-spec hash test.

**Not done, deliberately:** the missing reconciliation counter. Adding `supplemental_source_count` means changing `PHASE_II_COLUMNS`, which changes the frame the census hash-verifies in `_verify_phase_ii_provenance` — a schema change to a frozen-criteria artifact that wants its own amendment and its own PR, not a quiet addition here. The gap stands exactly as described: a federal award that consumed two SBIR.gov source rows is indistinguishable in the output from one that matched none.

The normalized-vs-raw formatting split on filled taxonomy values is the same missing-provenance issue and should be settled alongside the counter rather than patched separately.

---

PR #500, opened 2026-08-02
Stage 2: consolidate transition vendor identity keys

hollomancer:
## Summary

- add durable `vendor-key-v1` policy for transition vendor matching
- route both the vendor crosswalk and resolver through the same name key
- keep persisted canonical and alias names as human-readable display text
- retain the consumer-named Stage 1 profiles as compatibility policies
- document the Stage 2 corpus audit and matching semantics

## Why this is Stage 2

Stage 1 centralized the previously forked implementations behind `sbir_etl.identity` without silently changing most outputs. This PR consolidates the next demonstrably safe pair: the transition crosswalk and resolver had separate keys for the same vendor-resolution domain.

The crosswalk previously indexed display-cleaned names, while the resolver canonicalized case, punctuation, ampersands, and common long/short legal designators. Both now use `vendor-key-v1` for name indices, exact comparisons, and fuzzy-match inputs. Display cleanup remains separate so matching keys cannot overwrite persisted source-facing names.

## Real-data audit

Compared the legacy crosswalk key with `vendor-key-v1` over 34,464 distinct company names from:

```
award_data.csv
SHA-256 efdf7ca5a398703002ebb33345275b0f68e50af3c5db361d48a2456266a23628
```

Results:

- legacy crosswalk partitions: 34,459
- shared vendor-key partitions: 34,424
- newly joined groups: 35
- differences from the legacy resolver key: 0

All 35 joined groups were inspected. They are case, punctuation, ampersand, or legal-designator synonyms, including `Planetary Systems Corp.` / `PLANETARY SYSTEMS CORPORATION` and `Alme & Associates` / `Alme and Associates`. No ambiguous firms were joined.

## Verification

- `5,290 passed, 7 skipped` — full unit suite
- `179 passed` — focused identity, vendor-crosswalk, and vendor-resolver tests
- MyPy: no issues in 249 source files
- Ruff lint and format checks pass
- company identity boundary check passes
- corpus SHA and partition counts reproduced after implementation

## Scope boundary

This PR does not alter Phase 0 canonical IDs, UCC-specific abbreviation policy, SEC filing-name policy, Form D exact joins, or Phase III notice attribution. Those require their own output audits before consolidation.

### DIFF
--- a/packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
+++ b/packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
@@ -302,6 +302,11 @@ class VendorCrosswalk:
     ) -> tuple[CrosswalkRecord, float] | None:
         """
         Find best matching canonical record by name. Returns (record, score) or None.
+
+        A name that identifies more than one canonical record is ambiguous and returns
+        None. The shared vendor key deliberately joins spelling variants, so distinct
+        identifier-backed records can share a key; resolving that by index position
+        would select a record by insertion order.
         """
         if not name:
             return None
@@ -311,17 +316,17 @@ class VendorCrosswalk:
         # exact match
         cids = self._name_index.get(norm)
         if cids:
-            return self.records[cids[0]], 1.0
+            return (self.records[cids[0]], 1.0) if len(cids) == 1 else None
         # fuzzy choose best across canonical names
         best_score = 0.0
-        best_cid = None
+        best_cids: list[str] = []
         for canon_name, cid_list in self._name_index.items():
             score = _fuzzy_score(norm, canon_name)
             if score > best_score:
                 best_score = score
-                best_cid = cid_list[0]
-        if best_score >= fuzzy_threshold and best_cid:
-            return self.records[best_cid], best_score
+                best_cids = cid_list
+        if best_score >= fuzzy_threshold and len(best_cids) == 1:
+            return self.records[best_cids[0]], best_score
         return None
 
     def find_by_any(
@@ -374,8 +379,9 @@ class VendorCrosswalk:
             return False
         rec = self.records[canonical_id]
         alias_name_norm = _clean_display_name(alias_name) or alias_name
-        # avoid duplicates
-        if any(a.name == alias_name_norm for a in rec.aliases):
+        alias_key = _normalize_name(alias_name)
+        # avoid duplicates on the matching key, not on display text
+        if any(_normalize_name(a.name) == alias_key for a in rec.aliases):
             return True
         ad = AliasRecord(
             name=alias_name_norm,
@@ -385,9 +391,10 @@ class VendorCrosswalk:
         )
         rec.aliases.append(ad)
         # update name index
-        alias_key = _normalize_name(alias_name_norm)
         if alias_key:
-            self._name_index.setdefault(alias_key, []).append(canonical_id)
+            indexed = self._name_index.setdefault(alias_key, [])
+            if canonical_id not in indexed:
+                indexed.append(canonical_id)
         logger.info("Added alias '%s' to %s", alias_name, canonical_id)
         return True
 
--- a/packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
+++ b/packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
@@ -70,6 +70,14 @@ def _normalize_identifier(val: str | None) -> str | None:
 
 
 def _normalize_name(name: str | None) -> str | None:
+    """Build the shared transition-vendor key used for matching and indexing."""
+    if name is None:
+        return None
+    return normalize_company_name(name, profile=CompanyNameProfile.VENDOR_KEY_V1)
+
+
+def _clean_display_name(name: str | None) -> str | None:
+    """Clean persisted display text without turning it into an identity key."""
     if name is None:
         return None
     return normalize_company_name(name, profile=CompanyNameProfile.VENDOR_CROSSWALK_V1)
@@ -130,12 +138,12 @@ class CrosswalkRecord:
         return d
 
     def normalize(self) -> None:
-        self.canonical_name = _normalize_name(self.canonical_name) or self.canonical_name
+        self.canonical_name = _clean_display_name(self.canonical_name) or self.canonical_name
         self.uei = _normalize_identifier(self.uei)
         self.cage = _normalize_identifier(self.cage)
         self.duns = _normalize_identifier(self.duns)
         for a in self.aliases:
-            a.name = _normalize_name(a.name) or a.name
+            a.name = _clean_display_name(a.name) or a.name
 
 
 # ---------------------------------------------------------------------------
@@ -240,7 +248,7 @@ class VendorCrosswalk:
         # If canonical name differs, keep existing canonical_name but add alias
         if _normalize_name(base.canonical_name) != _normalize_name(rec.canonical_name):
             # add rec canonical name as alias if not present
-            alias_name = _normalize_name(rec.canonical_name) or rec.canonical_name
+            alias_name = _clean_display_name(rec.canonical_name) or rec.canonical_name
             if alias_name and alias_name not in existing_alias_names:
                 base.aliases.append(AliasRecord(name=alias_name, note="merged-canonical"))
         logger.info("Merged record %s into %s", rec.canonical_id, canonical_id)
@@ -365,7 +373,7 @@ class VendorCrosswalk:
         if canonical_id not in self.records:
             return False
         rec = self.records[canonical_id]
-        alias_name_norm = _normalize_name(alias_name) or alias_name
+        alias_name_norm = _clean_display_name(alias_name) or alias_name
         # avoid duplicates
         if any(a.name == alias_name_norm for a in rec.aliases):
             return True
@@ -377,7 +385,9 @@ class VendorCrosswalk:
         )
         rec.aliases.append(ad)
         # update name index
-        self._name_index.setdefault(alias_name_norm, []).append(canonical_id)
+        alias_key = _normalize_name(alias_name_norm)
+        if alias_key:
+            self._name_index.setdefault(alias_key, []).append(canonical_id)
         logger.info("Added alias '%s' to %s", alias_name, canonical_id)
         return True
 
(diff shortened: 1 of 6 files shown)


packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
```diff
@@ -70,6 +70,14 @@ def _normalize_identifier(val: str | None) -> str | None:
 
 
 def _normalize_name(name: str | None) -> str | None:
+    """Build the shared transition-vendor key used for matching and indexing."""
+    if name is None:
+        return None
+    return normalize_company_name(name, profile=CompanyNameProfile.VENDOR_KEY_V1)
```
hollomancer, line 76: **Merging the key partitions makes `find_by_name` return a `1.0` score for names that now identify two different firms, chosen by insertion order.**

The separation of matching key from display text is right, and the audit of the 35 newly joined groups is the correct diligence. The gap is that the audit checks whether the *groups* are safe, and the consequence lands somewhere else — on the lookup.

`find_by_name` (unchanged by this PR, but now reachable far more often) resolves an exact hit like this:

```python
cids = self._name_index.get(norm)
if cids:
    return self.records[cids[0]], 1.0
```

`_name_index` maps key → **list** of canonical ids, and this PR's own test now pins a two-element list as expected behavior:

```python
# tests/unit/transition/features/test_vendor_crosswalk.py
assert cw._name_index["acme corp"] == ["co-1", "co-2"]
```

The test comment is precise about what happens — "Both share the matching key while retaining separate identifier-backed records" — and that's the point: the records do *not* merge, because they carry distinct identifiers. Only the lookup collides. So `find_by_name("Acme Corp")` returns `co-1` with a perfect score, and `co-2` is unreachable by name, decided by which was inserted first. Nothing in the return value says the key was ambiguous.

This is pre-existing code, so it isn't introduced here — but this PR is precisely the change that widens the population reaching it, by design. Of the 35 groups the audit joined, any whose members are separate identifier-backed records land in exactly this state. The audit confirms the *names* are variants of one firm, which is reassuring, but it doesn't establish that each group is a single crosswalk record, and the new test demonstrates a case where it isn't.

It also sits awkwardly against the standard the repo applies elsewhere. The Phase II reconciliation rule in `specs/phase-iii-census/design.md` refuses to resolve a multi-way key by "row order, agency-name similarity, or another heuristic" and stops instead. `cids[0]` is row order.

The fix is local and doesn't require revisiting the key design:

```python
cids = self._name_index.get(norm)
if cids:
    if len(cids) > 1:
        return None          # or return the candidates, or degrade the score
    return self.records[cids[0]], 1.0
```

Whichever of those you pick, the property worth pinning in a test is that a name matching two distinct canonical records does not come back as a confident single answer. `find_by_any` propagates that `1.0` to callers as `method="name", score=1.0`, so an ambiguous name is currently indistinguishable from a UEI hit at the call site.

One smaller thing while this is open: after this change `vendor-resolver-v1` and `vendor-key-v1` share a single branch in `normalize_company_name`, and no non-test caller uses `VENDOR_RESOLVER_V1` anymore. `test_vendor_key_preserves_resolver_compatibility` therefore asserts a tautology — same code path, same input — so it will stay green even if the shared branch drifts. If the intent is to keep the legacy name as a frozen reference point, it needs its own implementation to be a real check; if the intent is that it's a deprecated alias, the docs table entry ("equivalent to `vendor-key-v1`") is the honest version and the test isn't buying anything.

---

packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
```diff
@@ -377,7 +385,9 @@ def add_alias(
         )
         rec.aliases.append(ad)
         # update name index
-        self._name_index.setdefault(alias_name_norm, []).append(canonical_id)
+        alias_key = _normalize_name(alias_name_norm)
+        if alias_key:
+            self._name_index.setdefault(alias_key, []).append(canonical_id)
```
hollomancer: Two things about this pair of lines, one a real defect and one a coupling worth naming.

**1. The duplicate guard above no longer matches the index key, so repeated aliases accumulate duplicate ids.**

The guard six lines up compares *display* names:

```python
alias_name_norm = _clean_display_name(alias_name) or alias_name
if any(a.name == alias_name_norm for a in rec.aliases):
    return True
```

`_clean_display_name` (`vendor-crosswalk-v1`) preserves case and does not canonicalize designators. So `add_alias(cid, "Acme Inc")` followed by `add_alias(cid, "ACME, INC.")` produces two `AliasRecord`s — and both reach the line below, both keying to `"acme inc"`:

```python
self._name_index["acme inc"] == ["co-123", "co-123"]
```

Before this PR those two forms produced two *distinct* index keys, so each list held one id and this couldn't happen. The collapse to a shared key is what surfaces it. `remove()` filters by id so it still cleans up, and `find_by_name` still returns the right record, so nothing is visibly broken today — but the index list grows on every re-add, and any future consumer that treats `len(cids) > 1` as "ambiguous name" (see my other comment, where I suggest exactly that) will read a self-duplicate as a genuine collision. Cheapest fix is to dedup on the key rather than the display text:

```python
if alias_key and canonical_id not in self._name_index.setdefault(alias_key, []):
    self._name_index[alias_key].append(canonical_id)
```

**2. Index keys are derived from display text, which the PR's own docs say must never be a match key.**

`docs/steering/company-identity.md` now reads `vendor-crosswalk-v1` → "transition crosswalk display cleanup only; never a match key". But every index key in this class is computed as `_normalize_name(_clean_display_name(x))`, not `_normalize_name(x)`: here explicitly, and in `add_or_merge`/`remove` implicitly, because `rec.normalize()` overwrites `canonical_name` with the cleaned form before it is keyed.

I checked and the composition is currently value-identical to keying from the raw name — `vendor-crosswalk-v1` only substitutes `,` `.` `/` `&`, and `vendor-key-v1` repeats all four substitutions before lowercasing and folding designators, so applying it twice is idempotent on this input. So there is no behavioral bug today. But it means the display profile is load-bearing for matching in exactly the way the doc line says it isn't, and a future edit to `vendor-crosswalk-v1` — the one profile the doc now advertises as safe to change freely, since it "only" affects display — would silently repartition the vendor index.

On this line the fix is one word: `_normalize_name(alias_name)` instead of `_normalize_name(alias_name_norm)`. For `add_or_merge` it needs the raw name captured before `rec.normalize()`. Either that, or a test asserting `_normalize_name(_clean_display_name(n)) == _normalize_name(n)` over a fixture set, so the invariant the code depends on is the invariant CI checks.

---

hollomancer (commented):
Reviewed at `45a3aab`. The core move — one `vendor-key-v1` for indices, exact comparisons and fuzzy inputs, with `vendor-crosswalk-v1` demoted to display cleanup — is the right shape, and keeping persisted canonical/alias text out of the key is the part that most easily goes wrong and didn't here. The audit over the same 34,464 names and pinned `award_data.csv` digest as the first consolidation is good practice; reusing the identical frame is what makes the 34,459 → 34,424 delta interpretable.

Two comments inline. The one that matters is that collapsing the key partitions changes what `find_by_name` does on a collision — it returns the first id in insertion order with a score of `1.0`, and this PR's own new test pins a two-record collision as expected state. The second is a duplicate-accumulation bug in `add_alias` that the shared key newly exposes, plus a note that index keys are currently computed *through* the display profile the docs now describe as never being a match key.

Nothing here argues against the consolidation itself.

---

packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
hollomancer: Addressed at `70953e0`.

`find_by_name` no longer resolves an ambiguous key by index position. The exact path returns `None` when the key holds more than one record, and the fuzzy path got the same treatment — it was doing `best_cid = cid_list[0]`, so fixing only the exact path would have left the identical order-dependence one branch over.

Checked the blast radius before changing behavior: `find_by_name` has no production callers. Its only caller anywhere is `find_by_any`, which is itself uncalled outside tests, so this is a semantics change to a currently-unwired lookup rather than to a live path.

New test builds the exact case from the comment — `Acme Corporation` / `Acme, Corp.` with distinct UEIs — and asserts the index holds `["co-1", "co-2"]`, that both name spellings return `None`, that `find_by_any(name=...)` returns `None`, and that UEI lookup still resolves. 539 tests in `tests/unit/transition` and `tests/unit/identity` pass.

On `test_vendor_key_preserves_resolver_compatibility`: left as-is. It's a tautology as you say, but it's a cheap one, and it does pin the thing the docs table now promises — if someone splits the shared branch and gives `vendor-resolver-v1` its own implementation, the test goes red and forces the decision to be explicit. Worth flagging rather than deleting.

---

packages/sbir-ml/sbir_ml/transition/features/vendor_crosswalk.py
hollomancer: Both addressed at `70953e0`.

**1.** `add_alias` now dedupes on the matching key rather than display text — `any(_normalize_name(a.name) == alias_key for a in rec.aliases)` — and the index append is guarded by `canonical_id not in indexed`. Belt and braces on purpose: the first prevents the redundant `AliasRecord`, the second makes the index self-correcting even if some other path appends. New test adds `"Acme Inc"`, `"ACME, INC."`, and `"Acme Incorporated"` to one record and asserts a single alias and `_name_index["acme inc"] == ["co-123"]`.

That guard matters more now than when I wrote the comment, since `find_by_name` treats `len(cids) > 1` as ambiguous — a self-duplicate would otherwise make a record unreachable by its own alias.

**2.** Took the one-word fix: `alias_key` is now computed from the raw `alias_name`, so this path no longer routes the matching key through the display profile.

Also took the test you suggested as the alternative, because it covers the `add_or_merge`/`remove` paths that the one-word fix doesn't reach — those still key from `canonical_name` after `rec.normalize()` has overwritten it. `TestVendorKeyIndexing` asserts `_normalize_name(_clean_display_name(n)) == _normalize_name(n)` over six fixtures, including the `PC Photonics` and `Planetary Systems Corp.` cases the docs call out. So the invariant the class depends on is now the invariant CI checks, and an edit to `vendor-crosswalk-v1` that would repartition the vendor index goes red instead of silent.

---

PR #504, opened 2026-08-02
Add the Dagster job test pattern; the three remaining migrations still open

hollomancer:
## Description

**Draft on purpose.** The test pattern below is implemented and passing; the three migrations it exists to serve are still open, and two of them need diagnosis before any code gets written. Opened as a doc-only stub and now carries real tests, so the metadata is corrected to match.

### Implemented: the test pattern

`weekly_awards_report_job` had **five passing tests** — discovery, cron, stopped default, and both op error branches — while being **unable to run at all**. Executing it raised `FileNotFoundError: Could not resolve SBIR awards CSV`, because it depends on a vintage `weekly_sbir_awards_download` produces. Wiring tests can't catch a job that is wired correctly to work it cannot do.

No new CI category is needed — layers 1–3a are ordinary unit tests, 3b is an ordinary integration test:

| Layer | Catches | Lives in | Runs |
|---|---|---|---|
| 1. Definitions load | A `_HOST_SCHEDULES` entry naming a job that doesn't exist | `tests/unit/assets/jobs/test_host_schedules.py` | every PR |
| 2. Op in isolation | Error branches, output paths | `tests/unit/…` | every PR |
| 3a. Job graph, script mocked | Job runs end to end and fails when it should | `tests/unit/…_execution.py` | every PR |
| 3b. Job graph, real script | Whether the script itself works | `tests/integration/dagster/` | `main` |
| 4. Real run on host | Bulk data, credentials, network | — | manual, per runbook |

**Layer 1 closes a live hazard.** `definitions.py` resolves each schedule with `_get_job(name)` and, on `None`, logs a warning and `continue`s. A typo doesn't fail — it silently drops the schedule. Since every host schedule defaults to STOPPED, a dropped schedule is indistinguishable from a disabled one until someone notices months of missing output. Same silent-rot shape as the broken `weekly.yml`.

**Layer 3b asserts an invariant, not a fixed outcome** — deliberately. The script resolves the CSV locally and falls back to downloading, so the result legitimately differs by environment: with a vintage or network it produces a report, without either it raises. Asserting either specifically would be flaky on whichever runner disagrees. What holds in both worlds is the contract the op enforces: *never report success without having written a report*. Flagging this because it's the assertion most likely to get "tightened" later, which would reintroduce the flakiness.

### Still open: the three migrations

**1. Weekly awards report** — job and schedule shipped in #502; what remains is operational. `weekly_sbir_awards_download` must be enabled **first**, or the report fails every Monday. Open question in the doc: whether a run-status sensor beats relying on cron spacing.

**2. Monthly benchmark — diagnose first.** Every month since May showed a ~30s `failure` then a ~60s `success` against a step budgeting 90 minutes. Pull a `run_manifest.json` artifact **before the 30-day retention expires** and establish what the successful runs actually produced.

**3. Monthly procurement report — diagnose first.** Never succeeded; its only run died 5s into the download step. The move also lets that step be deleted — `sbir_awards_download_job` already fetches the CSV weekly.

Layer 3b is where both diagnoses actually happen.

### Blocking all three

**Failure notification.** A stopped-by-default schedule on a tailnet-only host tells nobody when it breaks — or when it never starts. Same failure mode that opened this whole thread. Decide the path before enabling anything.

## Type of Change

- [ ] Bug fix
- [x] New feature (test pattern + guards)
- [ ] Breaking change
- [x] Documentation update

## Testing

- `tests/unit/assets/jobs` + `tests/integration/dagster` → **11 passed**
- `ruff check` / `ruff format` clean; `markdownlint-cli2` clean on the spec docs
- Verified the new unit tests land in the PR-time selection (`tests/unit -m "not slow"`) and the integration test does not

- [x] All tests pass locally

## Checklist


tests/integration/dagster/test_weekly_awards_report_real_script.py
```diff
@@ -0,0 +1,47 @@
...
+    from sbir_analytics.assets.jobs import weekly_awards_report as mod
+
+    monkeypatch.setenv(mod.DATA_ROOT_ENV, str(tmp_path))
+    # Keep the run bounded and offline-ish: shortest lookback, no AI calls, and
+    # skip the SBIR.gov API the script offers a flag for.
+    monkeypatch.setenv(mod.LOOKBACK_DAYS_ENV, "1")
+    monkeypatch.setenv("SKIP_SBIR_API", "1")
+    monkeypatch.delenv("OPENAI_API_KEY", raising=False)
+
+    try:
```
hollomancer: This makes the integration test vacuous whenever the job fails: `execute_in_process(raise_on_error=False)` normally returns an unsuccessful result, and the test then performs no assertion; an exception from an import, configuration regression, or broken script is also accepted. Assert the expected unavailable-data failure (or otherwise verify that the real script was reached) when `result.success` is false, so this test actually detects regressions beyond the success path.

hollomancer (commented):
Independent review: the layer-3b test needs to distinguish the explicitly permitted unavailable-data failure from unrelated job failures.

tests/integration/dagster/test_weekly_awards_report_real_script.py
hollomancer: Fixed at `62566b2`. The criticism was right and it was worse than "no assertion on the failure path": in this container there is no local awards vintage, so the failure branch is the one actually taken, meaning the test was asserting nothing at all where it runs.

Two changes:

- **The failure branch is now pinned.** On `result.success is False` the test pulls `failure_data_for_node("generate_weekly_awards_report_op")` and requires the rendered error to contain either `weekly_awards_report.py failed with exit code` or `Report script wrote no report to`. Both are raised by the op *after* `subprocess.run` returns, so either one proves the real script was invoked — which is the property this layer-3 test exists for and the one the mocked unit companion can't give. A missing `failure_data_for_node` now fails explicitly as "failed before reaching the report op", which is the job-graph regression case.
- **The blanket `try/except Exception: return` is gone.** With `raise_on_error=False`, step failures don't raise, so that handler was only ever catching framework-level errors — exactly the class of regression it then discarded.

Verified both directions rather than just re-running green:

- unmodified: 1 passed, taking the failure branch (`RuntimeError: weekly_awards_report.py failed with exit code 1`);
- with an `ImportError` injected at the top of the op to simulate a config regression: **failed**, as it should. Under the old test that case passed.

The docstring now says the failure branch is pinned too, so the next reader doesn't restore the permissive version thinking the environment-dependence justifies it.

---

hollomancer:
## Review — CI is green again; draft status is correct; one landmine with #516

Reviewed alongside #516, #515 and #478.

### CI was red; it is now green

`Lint, Types, and Guards` was failing on this branch — a broken Markdown link in `specs/actions-migration-followups/remaining-dagster-migrations.md:4`, which pointed at `../deployment/actions-migration-plan.md`. That resolves to `specs/deployment/…`; the plan lives under `docs/`. Fixed at `ca60102`, and all 13 checks now pass.

### On the substance

The layer 1–4 taxonomy is the right decomposition, and **layer 1 is the one earning its keep**. `definitions.py` resolving a schedule with `_get_job(name)` and doing `log warning; continue` on `None` means a typo doesn't fail — it silently drops the schedule, and since every host schedule is STOPPED by default, a dropped schedule and a disabled one are indistinguishable until someone notices months of missing output. That's the same silent-rot shape as the `weekly.yml` that started this thread, and a test that would have caught it is worth more than the migrations it's attached to.

Layer 3b is now pinned on both branches after the review thread above — the failure path asserts the *specific* unavailable-data failure rather than accepting any failure, so it can't pass vacuously on a runner with no local vintage (which is every runner without the CSV, including this one).

### Draft status is correct, and I'd keep it

By its own account two of the three migrations need diagnosis before code, and failure notification blocks all three — a stopped-by-default schedule on a tailnet-only host tells nobody when it breaks. That last one is worth treating as the real blocker: it's the failure mode that produced this PR, and enabling anything before it's decided reproduces it.

If you want the value early, the test pattern alone (layers 1–3b, no migration work) is self-contained and would be safe to split out onto `main` while the diagnoses proceed.

### Blocking interaction with #516 — no textual conflict, still lands red

#516 archives `docs/deployment/actions-migration-plan.md` → `docs/archive/deployment/actions-migration-plan.md`. The link fixed above points at the pre-archive path.

Git reports no conflict between the branches. I test-merged the pair onto `main` and ran the hygiene checker:

```
Missing local Markdown links were found in live docs/specs:
specs/actions-migration-followups/remaining-dagster-migrations.md:4: missing local Markdown link ../../docs/deployment/actions-migration-plan.md
```

Merging both — **in either order** — turns `main` red, silently, because neither PR's own CI can see the other. Since #516 is mergeable now and this is a draft, the clean sequence is: land #516, then repoint this link at `../../docs/archive/deployment/actions-migration-plan.md`. I deliberately did *not* pre-point it there, since that path doesn't exist until #516 merges and doing so would turn this branch red today.

---

PR #509, opened 2026-08-03
docs: tier the repo by epistemic status, and make the tiers operative

hollomancer:
## Description

Establishes tiering by epistemic status as the repository's organizing pattern, and wires it into the places that actually govern day-to-day work. Documentation and agent definitions only — no code moves, no directories created.

The motivating problem is maintenance cost. As research questions accumulate, a repository with one undifferentiated standard has to hold everything to the standard of its most rigorous artifact. The tiers cap that by making it *legitimate to not maintain most of the code*: `exploratory` work is allowed to be untested and stale because it is labeled non-citable, and in exchange `evidence` stays small and genuinely trustworthy.

### 1. Architecture overview — drift fixes and the tier rationale

Corrected an AWS deployment path (Step Functions, S3 data lake, EC2 Neo4j) that is not the live architecture, an `infrastructure/` CDK tree that is not in the repository, and a Lambda path that doesn't exist. Added the FastAPI `api/` package and `studies/` to the tree.

New §1.3 records the measured dependency direction: 141 imports from `packages/` into `sbir_etl`, zero the other way, including under `TYPE_CHECKING` and in deferred imports. The layering is acyclic. The section also notes the tree understates it — `sbir_etl/` sits at the root as an apparent peer of its own dependents — and that relocating it to `packages/sbir-etl/` would make layout match imports.

New §10 gives the architectural rationale for the four tiers and records the Phase III census as the `evidence` reference implementation.

### 2. `docs/steering/epistemic-tiers.md` — the canonical contract

New steering doc, and the single source the agents read: per-tier contracts, a four-question classification procedure, the declare-and-promote rules, and a mapping from `research-questions.md` answerability labels to the tier that has to stand behind each one. A question cannot be marked answerable on exploratory-tier work.

| Tier | Contract |
|---|---|
| `primitives` | One implementation per concept, versioned behavior, comprehensive tests |
| `pipelines` | Deterministic, reproducible from a declared data cut, no inference |
| `evidence` | Frozen spec + SHA enforcement + blocking asset checks + declared estimand — all four, no partial admission |
| `exploratory` | Labeled non-citable. Nothing else required. |

The doc records current status honestly: no tier directories exist yet; `sbir_etl/identity/` and the Phase III census already meet their contracts; the identity boundary checker runs in no workflow; config has 15 loader-bypassing `yaml.safe_load` sites.

### 3. Agents — enforcement at the points where tier is decidable

- **`scope-guard`** gains a tier check and a `RETIER` verdict, catching both directions of mis-tiering — a citable claim resting on uncontracted work, and exploratory code carrying hardening it will never need — plus silent promotion, which no other stage sees.
- **`spec-implementer`** establishes the tier before writing code, builds to it and not above it, and refuses to promote by importing `scripts/` code or quoting an exploratory number.
- **`quality-sweep`** and **`test-fixer`** scale effort by tier instead of uniformly. Untended exploratory code is explicitly the design working rather than a backlog. Neither may resolve a failure by weakening a contract — downgrading a blocking asset check or loosening a pinned hash is a contract change, not a cleanup.

### 4. `CLAUDE.md` / `AGENTS.md` — single-sourced

CLAUDE.md carries the tier summary and rules. AGENTS.md is now a pure routing file with no rules of its own — it had duplicated the live-deployment rule set from CLAUDE.md since before this branch. Nothing was dropped: the two bullets AGENTS.md uniquely carried moved into CLAUDE.md (both were already in the runbook), and two qualifiers that were *stronger* in AGENTS.md are now the CLAUDE.md wording — ingress as Tailscale Serve over tailnet-only HTTPS, and LAN/public port exposure prohibited alongside Funnel.

### Findings worth flagging

The premises this started from were partly stale; the corrections are in the docs:

- **The identity consolidation already landed.** `sbir_etl/identity/` exposes one `company_name_similarity` contract over versioned `CompanyNameProfile` policies, and all nine previously-forked call sites route through it. The profiles preserve divergent recall deliberately — the module docstring is explicit that uniform behavior is not the goal. The remaining gap is narrow: `scripts/ci/check_identity_boundaries.py` is wired into no workflow, Makefile target, or pre-commit hook, and the profile governing the census control frame isn't documented next to the estimand.
- **There is no back-edge.** `sbir_etl` → `packages/` is zero, not one.
- Config is the one primitive genuinely still forked: 15 `yaml.safe_load` sites. Deliberately **not** in this PR — see below.

## Type of Change

- [x] Documentation update

## Testing

Documentation and agent definitions only; no code paths touched, so no tests were added.

- `markdownlint-cli2` on every changed and added file: the new steering doc is clean; the remaining findings are pre-existing and none are in added text (counts verified identical on `main` by stashing).
- Every referenced path checked to exist — the 13 module, spec, script, and test paths quoted across the docs.
- Every figure re-measured against the working tree rather than carried over from the analysis that prompted this PR. Three were wrong on first write and corrected before commit: the `scripts/` subtree count, the back-edge count, and the "larger than any package" claim (`scripts/` at 42,867 lines is larger than `sbir-analytics` at 32,101 but smaller than `sbir_etl` at 59,902).
- Consolidation checked for loss by diffing the old AGENTS.md rule set against CLAUDE.md — all nine rule keys confirmed present.

## Checklist

## Follow-ups, deliberately not in this PR

1. **Config primitive consolidation** — 15 `yaml.safe_load` sites routed through `sbir_etl/config/loader.py`. Behavioral risk (validation, defaults, and env overrides would begin applying where they don't today), so it wants its own PR and its own test pass.
2. **Wire `check_identity_boundaries.py` into CI** — a guard that runs only in its own unit test is a guard against accidents, not drift. Roughly a one-line workflow addition.
3. **Document the census control-frame profile** in `specs/phase-iii-census/`, next to the estimand.
4. **Tier labels in spec `requirements.md` files** — the labeling step §10.5 calls the first useful move.
5. **`sbir_etl/` → `packages/sbir-etl/`** — cosmetic, cheap, and explicitly last.

### DIFF
--- a/.claude/agents/quality-sweep.md
+++ b/.claude/agents/quality-sweep.md
@@ -17,11 +17,32 @@ You are a code quality engineer for the SBIR Analytics project. Your job is to s
 6. **Run ruff format**: `uv run ruff format sbir_etl/` to standardize formatting
 7. **Run tests**: `uv run pytest tests/unit/ -x -q -m "not slow"` to verify nothing broke
 
+## Sweep Intensity by Tier
+
+Read `docs/steering/epistemic-tiers.md`. Cleanup effort is not uniform — spending
+it evenly is how maintenance cost grows without bound as research questions
+accumulate. Sweep hardest where things depend on the code:
+
+| Tier | Where | Effort |
+|---|---|---|
+| `primitives` | `sbir_etl/identity/`, `sbir_etl/config/`, `sbir_etl/models/` | Full: ruff, mypy, formatting, type annotations, docstrings |
+| `pipelines` | `sbir_etl/` (rest), `packages/` | Full ruff + mypy; annotate what you touch |
+| `evidence` | Phase III census assets and their specs | Ruff + mypy, but **change no behavior** — output hashes are pinned. If a lint fix would alter output, report it instead of applying it. |
+| `exploratory` | most of `scripts/`, all of `scripts/archive/` | Leave it alone unless asked |
+
+Do not "improve" `scripts/` on your own initiative. Untouched exploratory code is
+the tiering working as designed, not a backlog.
+
 ## Rules
 
 - Code standards are in CLAUDE.md — follow them
 - MyPy: Gradual typing (relaxed), Pydantic plugin enabled
 - Fix issues in batches by file, not one at a time
 - Don't add type annotations to code you didn't change (unless that's the goal)
+- Never let a cleanup change a pinned output. In `evidence` code, formatting and
+  annotations are safe; reordering operations, changing float handling, or altering
+  iteration order are not.
+- If a sweep would consolidate duplicate logic into a shared helper, that is a
+  primitive change — report it, don't do it inline
 - Run tests after each batch of fixes
-- Report what was fixed at the end
+- Report what was fixed at the end, grouped by tier
--- a/.claude/agents/scope-guard.md
+++ b/.claude/agents/scope-guard.md
@@ -53,23 +53,55 @@ For each item, answer these questions:
 - Does this produce an analytical output that replicates or exceeds a NASEM claim?
 - Is this building awards infrastructure (duplicative) or outcomes infrastructure (novel)?
 
+### 5. Tier
+Read `docs/steering/epistemic-tiers.md` first. Every change targets one tier:
+`primitives`, `pipelines`, `evidence`, or `exploratory`.
+
+- What tier does this claim? If the spec doesn't say, it is `exploratory` — hold
+  it to that and say so.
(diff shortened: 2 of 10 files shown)


hollomancer (commented):
Independent review: the executive-summary wording has been corrected and the original thread resolved. One remaining documentation inconsistency should be addressed: §6 still presents the old AWS/Lambda/S3 flow as the optional cloud architecture, despite the opening section now saying that deployment path is historical and non-live. Please either rewrite §6 to reflect the current Docker Compose deployment or explicitly mark it as archived historical material.

PR #510, opened 2026-08-03
fix(config): one strict YAML mapping read, and three crash sites fixed

hollomancer:
## Description

Follow-up to #509, which listed "config primitive consolidation — 15 `yaml.safe_load` sites routed through `sbir_etl/config/loader.py`" as the next step. **That framing was wrong**, and this PR does the smaller, real thing instead.

### Why the mass consolidation isn't the fix

I read all 15 sites. They are not 15 copies of one concept. Two are `loader.py` itself — that *is* the pipeline-config resolution path. The other thirteen load **domain data** — CET taxonomies, defense crosswalks, study manifests, census profiles, BEA fallbacks, feature configs, baselines — with deliberately different error policies:

| Policy | Sites |
|---|---|
| `or {}` — empty/absent file is a valid state | `tech_census` ×2, `transition_report_paths`, `fiscal_bea_mapper`, `patent_features`, `baselines` |
| swallow → `{}`, documented graceful degradation | `cet_vocabulary` |
| strict — must be a mapping, hand-rolled | `defense_taxonomy_crosswalk`, `study_manifest` |
| **strict by assumption, unchecked** | `local_cet_classifier` ×2, `taxonomy_loader` ×2 |

Routing these through `loader.get_config` would be actively wrong — that function returns a validated `PipelineConfig`, which a CET taxonomy is not. And a single helper covering all four policies would need a policy flag, turning a three-line check into a configuration surface. That's the "configurable everything" red flag in `scope-guard`'s own list.

### What this PR actually does

Adds `sbir_etl/config/yaml_io.read_yaml_mapping` — one implementation of the narrow concept the last two rows share: *read a YAML file that must contain a mapping, and fail with a message naming the file*. Six call sites now use it.

**The bottom row was a live bug.** Four calls did a bare `safe_load` and immediately treated the result as a mapping. On an empty or comment-only file `safe_load` returns `None`:

- `load_local_cet_rule_classifier` → `AttributeError: 'NoneType' object has no attribute 'get'`
- `TaxonomyLoader.load_taxonomy` → same
- `TaxonomyLoader.load_classification_config` → `TypeError` on `ClassificationConfig(**None)`

None of those named the offending file. Both `TaxonomyLoader` docstrings also promised `FileNotFoundError` / `ValueError`, neither of which the code could raise — corrected to what actually comes out (`FileSystemError`, `ConfigurationError`, `ValidationError`).

The two sites that already hand-rolled the `isinstance` check are a straight substitution, and `defense_taxonomy_crosswalk._load_yaml` goes away entirely.

### Deliberately unchanged

The ten remaining `yaml.safe_load` sites. `loader.py`'s two are the config-resolution path; the rest treat an absent or empty file as valid and would change behavior for no gain. Under the tier framing in #509 this is the `primitives` rule working as intended: one implementation per *concept*, and "read a taxonomy for graceful degradation" is a different concept from "read and strictly validate a manifest."

## Type of Change

## Testing

- **5414 unit tests pass**, 7 skipped (`pytest tests/unit/ -m "not slow"`).
- **9 new tests.** `tests/unit/config/test_yaml_io.py` covers the helper: valid mapping, missing file, empty file, comment-only file, three non-mapping top-level types, malformed YAML, and that the description reaches every failure message.
- **3 regression tests** pin the crash sites — empty taxonomy, non-mapping taxonomy, and empty classification config now raise `ConfigurationError` naming the file instead of `AttributeError`/`TypeError`.
- Confirmed no existing test depended on the old exception types, and no caller catches `FileNotFoundError` around the changed functions.
- `ruff check`, `ruff format --check`, and `mypy` clean (239 source files).
- All four CI guards pass, including the identity boundary guard newly wired up in #509.

## Checklist

## Note on #509

#509's follow-up list should be corrected — item 1 describes a consolidation that shouldn't happen. Happy to amend that PR body, or leave this one as the record.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FH27aKYTeWK8wEKq7YHBxd)_

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/assets/phase_iii_census/assets.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/phase_iii_census/assets.py
@@ -21,6 +21,7 @@ from sbir_analytics.assets.phase_transition.sbir_gov_source import (
     sha256_file,
     verify_sbir_gov_materialization,
 )
+from sbir_etl.exceptions import ConfigurationError
 from sbir_etl.quality.study_manifest import load_study_manifest
 
 from .criteria import (
@@ -128,7 +129,7 @@ def verify_materialization_gate() -> dict[str, Any]:
 
     try:
         manifest = load_study_manifest(STUDY_MANIFEST_PATH)
-    except (OSError, ValueError) as exc:
+    except (OSError, ValueError, ConfigurationError) as exc:
         raise CensusInputError(
             f"Phase III census study manifest is missing or invalid at {STUDY_MANIFEST_PATH}: {exc}"
         ) from exc
--- a/packages/sbir-ml/sbir_ml/ml/config/taxonomy_loader.py
+++ b/packages/sbir-ml/sbir_ml/ml/config/taxonomy_loader.py
@@ -7,10 +7,10 @@ Loads YAML configuration files and validates them against Pydantic schemas.
 from pathlib import Path
 from typing import Any
 
-import yaml
 from loguru import logger
 from pydantic import BaseModel, Field, field_validator
 
+from sbir_etl.config.yaml_io import read_yaml_mapping
 from sbir_etl.exceptions import FileSystemError
 from sbir_etl.models.cet_models import CETArea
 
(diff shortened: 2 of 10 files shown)


hollomancer (commented):
Independent review: no blocking issue found. The shared reader remains appropriately narrow, and the changed strict-mapping call sites and regression coverage align with its intended error contract.

PR #517, opened 2026-08-03
Enable opt-in Neo4j access over Tailscale

hollomancer:
## Summary

- add an opt-in, TLS-terminated Tailscale Serve route from tailnet port `17687` to loopback Neo4j Bolt
- preserve localhost-only Docker bindings and keep Neo4j Browser HTTP private
- manage route ownership, conflict detection, rollback, disable reconciliation, and removal through the existing server helper
- document the least-privilege operator grant and PocketGraph iPhone connection

## Why

The Mac mini currently exposes only Dagster and the read-only analytics API over Tailscale. A native iPhone graph client needs Bolt connectivity, but a manual Serve command would contradict the runbook and create untracked configuration drift.

This change keeps direct Neo4j access disabled by default. An operator must first apply the narrow tailnet grant, then set `NEO4J_TAILNET_BOLT_ENABLED=true` in the live `.env.server`. Setting it back to false removes the managed route on the next `server-tailscale-up` reconciliation.

## Validation

- `uv run pytest -q tests/unit/test_server_ops.py tests/unit/test_server_runtime_contract.py` — 28 passed
- `uv run ruff check scripts/server/tailscale-route-state.py tests/unit/test_server_ops.py`
- `uv run ruff format --check scripts/server/tailscale-route-state.py tests/unit/test_server_ops.py`
- `sh -n scripts/server/tailscale-serve.sh`
- `git diff --check`

No live deployment or Tailscale policy was changed.

### DIFF
--- a/.env.server.example
+++ b/.env.server.example
@@ -29,11 +29,14 @@ SERVER_LOOPBACK=127.0.0.1
 # Host ports (all hard-bound to 127.0.0.1)
 NEO4J_HTTP_PORT=7474
 NEO4J_BOLT_PORT=7687
+# TLS-terminated Tailscale Serve port for trusted Neo4j operators.
+NEO4J_TAILNET_BOLT_ENABLED=false
+NEO4J_TAILNET_BOLT_PORT=17687
 DAGSTER_PORT=3000
 SBIR_ANALYTICS_API_PORT=8010
 
 # -----------------------------------------------------------------------------
-# Neo4j credentials (private to the host; never exposed to the tailnet)
+# Neo4j credentials (used by the tailnet Bolt route; rotate and restrict carefully)
 # -----------------------------------------------------------------------------
 NEO4J_USER=neo4j
 # Required — the compose file fails fast if this is empty.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,8 +46,9 @@ materialization state belongs there, not in tracked documentation.
 The only live checkout is `/Users/conradhollomon/projects/sbir-analytics-server`;
 never operate the live stack from the development checkout. Preserve
 `.env.server`, `/Volumes/SSDmini/sbir-analytics`, and the Docker `dagster_home`
-volume. Ingress must remain Tailscale Serve over tailnet-only HTTPS; never enable
-Funnel and never expose server ports to the LAN or public internet.
+volume. Ingress must remain Tailscale Serve over tailnet-only HTTPS or explicitly
+enabled TLS-terminated TCP; never enable Funnel and never expose server ports to
+the LAN or public internet.
 
 Treat materialization as a live-data mutation: confirm the SSD is mounted, the
 deployment checkout is clean, and the stack is healthy before running one. Keep
--- a/Makefile
+++ b/Makefile
@@ -638,7 +638,7 @@ server-backup: server-env-check ## Dump Neo4j to $(SERVER_BACKUP_DIR) (default .
 	$(call run,SERVER_ENV_FILE=$(SERVER_ENV_FILE) COMPOSE_FILE=$(SERVER_COMPOSE_FILE) ./scripts/server/backup.sh)
 
 .PHONY: server-tailscale-up
-server-tailscale-up: server-env-check ## Configure persistent Tailscale Serve routes (443->Dagster, 8443->API)
+server-tailscale-up: server-env-check ## Configure persistent Tailscale Serve routes
 	@$(call info,Configuring Tailscale Serve routes)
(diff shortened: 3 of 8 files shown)


hollomancer:
## Review — the design is sound; one unvalidated env var can strand port 443 forwarding to Bolt

All 13 checks green, `clean`, draft. The shape of this is right: opt-in and default-off, ownership checked before every mutation, transactional rollback, disable-reconciliation on the next `up`, loopback bindings untouched, `serve` never `funnel`. The route-state classifier requires `AllowFunnel` to be falsy for a port before calling it `owned`, so a Funnel-enabled port is refused rather than adopted — the prohibition is defended in code, not just prose. Test coverage is real: enabled, disabled, removal-on-disable, foreign-route refusal, and full rollback after a late failure.

### Blocking: `NEO4J_TAILNET_BOLT_PORT` is never validated

It is read from `.env.server`, defaulted, and used directly. Nothing checks that it is numeric, in range, or **distinct from the other managed ports**. Set it to `443` and the failure is not a clean refusal:

```
➤ Configuring HTTPS 443 -> http://127.0.0.1:3000...
➤ Configuring HTTPS 8443 -> http://127.0.0.1:8010...
➤ Configuring TLS/TCP 443 -> 127.0.0.1:7687...
⚠ Serve port 443 changed after creation; leaving it untouched.
⚠ Rolling back the newly-created Serve port 8443 route.
⚠ Serve port 443 changed after creation; leaving it untouched.
✖ Tailscale did not install the expected TLS/TCP 443 route.
```

Exit 1, and the serve state it leaves behind:

```json
{
  "TCP": { "443": { "TCPForward": "127.0.0.1:7687", "TerminateTLS": "node.test.ts.net" } },
  "Web": { "node.test.ts.net:443": { "Handlers": { "/": { "Proxy": "http://127.0.0.1:3000" } } } }
}
```

Reproduced against the branch using this PR's own fake-`tailscale` harness with `NEO4J_TAILNET_BOLT_PORT=443`.

Three things go wrong, and they compound:

1. **The pre-flight check can't see the collision.** All three route states are read before any mutation, so port 443 reads `free` for *both* the Dagster check and the Neo4j check. Nothing compares the requested ports to each other.
2. **The second write silently overwrites the first.** `--tls-terminated-tcp=443` replaces the `TCP` entry that `--https=443` just created, leaving an orphaned `Web` handler behind.
3. **Rollback then refuses to clean up.** `rollback_expected_route` sees 443 in a state it doesn't recognize as `owned` and correctly declines to touch it — twice. The safety property that protects foreign routes is what prevents self-repair here, so the script exits non-zero having *left the damage in place*.

The end state is the one this PR exists to prevent: Dagster's ingress is gone, and Bolt is answering on the port `group:sbir-analysts` is granted (`tcp:443`), so the analyst/operator separation the grant creates is inverted. A failed run of `server-tailscale-up` should never be able to leave the live ingress in a worse state than it found it.

The fix is small and belongs before the first mutation — validate that the port is an integer in 1–65535 and not equal to 443 or 8443, and `exit 2` with a clear message. That is the same fail-closed treatment `NEO4J_TAILNET_BOLT_ENABLED` already gets for a non-boolean value. Worth a test asserting the serve state is untouched when the port collides. **Happy to push this if you want it.**

### Non-blocking

- **Stale security comment in the file that defines the boundary.** `docker-compose.server.yml:95` still reads `# Loopback-only. Neo4j is intentionally NOT exposed through Tailscale.`, while the header comment 80 lines above now documents `17687 -> Neo4j Bolt`. Same file, opposite claims. The header was updated and this one was missed; it's the comment an operator reads when they're looking at the port binding itself.
- **The grant ordering is load-bearing but unenforced.** The runbook is explicit that the operator grant must be applied *before* flipping the flag, and it can't be otherwise — ACLs are tailnet-side and the node can't read them. But if someone sets the flag first, the route comes up reachable by every tailnet device, with no signal that anything is wrong. Since it can't be checked, it should at least be *said* at the moment of risk: a `warn` when the route is created, naming the grant, rather than the current bare `info` with the `bolt+s://` URL. Cheap, and it puts the warning where the mistake happens.
- **No Funnel test on the new path.** `test_tailscale_down_refuses_funnel_enabled_route` covers 443 only. `classify()` handles the TLS/TCP branch correctly, but the highest-consequence invariant in this repo deserves a test on the new route too.

### On the `CLAUDE.md` amendment

Widening "tailnet-only HTTPS" to "HTTPS or explicitly enabled TLS-terminated TCP" is the honest edit — the alternative would have been to leave the stated invariant contradicting the code. It keeps both hard prohibitions (no Funnel, no LAN/public exposure) intact, and "explicitly enabled" carries the default-off property. No objection.

---

PR #521, opened 2026-08-03
chore: refresh agent guidance and prevent drift

hollomancer:
## Summary

- align scope review with the current A–F research-question inventory and recognized operating duties
- require spec agents and the review skill to check lifecycle status, research purpose, epistemic tier, and current implementation before acting on unchecked tasks
- update test and quality workflows to match current Make, `uv`, CI, package, and Dagster patterns
- add tracked Codex wrappers that route to the shared `.claude/agents/` role instructions
- add the shared `.agents` copy of `review-spec` and enforce parity between runtime copies
- extend repository hygiene checks to agent-document links, missing wrappers, and divergent skills

## Why

The tracked Claude role files, local Codex definitions, and shared skill copies had drifted. The scope guard still used the retired M1–M5 framing and referenced a deleted document, while spec agents could treat unchecked tasks in gated or deferred specs as current work. Agent files were also outside the documentation guard, so these failures were not detected.

## Impact

- Codex and Claude agents now follow one shared role body instead of separate instruction copies
- spec work starts from `specs/status.md`, the current workflow, and research questions A–F
- `make docs-check` fails when agent links break, Codex wrappers are missing or misrouted, or skill copies differ
- no ETL, data, deployment, or research-result behavior changes

## Validation

- `make docs-check`
- `pytest tests/unit/scripts/test_repository_hygiene.py -q` — 13 passed
- Ruff check and format check for the modified Python files
- architecture boundary check
- company identity boundary check
- study manifest validation
- `git diff --check`


hollomancer:
## Review — the guard genuinely works; one instruction reads ambiguously

Draft, `clean`, merges cleanly into `main`. 13 hygiene tests pass, the guard passes, and the four other CI guards pass.

### The enforcement is real, not decorative

Guidance-drift PRs usually assert their invariants. This one enforces them, so I tried to break it rather than take the tests' word:

| I broke | Guard said |
|---|---|
| appended a line to `.agents/skills/review-spec/SKILL.md` | `agent skill differs from .claude copy: review-spec` |
| removed `.Codex/agents/test-fixer.toml` | `missing Codex wrapper for shared agent role: test-fixer` |

Both caught, both with the file and reason named. The wrapper design is the right shape too — `.Codex/agents/*.toml` carries only a name, a description and a pointer to `.claude/agents/<name>.md`, so there is one role body rather than two that agree until they don't. Bringing `CLAUDE.md`, `AGENTS.md`, `.claude/` and `.agents/` inside `_is_documentation_file` means the link checker now covers the files agents actually execute, which is where a broken reference costs the most.

### The one thing I'd change: the `__future__` annotations rule

The relaxation is **substantively correct** — I checked rather than assuming, because the old blanket rule existed for a reason:

- `agency_private_capital/asset.py` has `from __future__ import annotations` *and* a `context: AssetExecutionContext` parameter, which is exactly the combination the old rule forbade;
- the full `Definitions` object resolves — 109 assets, that one included.

So Dagster resolves the postponed annotation fine, and the blanket prohibition was over-broad. Good change.

The problem is the wording. It says to follow "the local pattern in `phase_iii_census/assets.py`, `phase_iii_candidates/assets.py`, and `agency_private_capital/asset.py`" — but those files demonstrate **two different patterns**:

- `phase_iii_census/assets.py` sidesteps the issue entirely: its asset signature is `def phase_iii_census(context=None, ...)`, with no annotation for Dagster to inspect;
- `agency_private_capital/asset.py` annotates `context: AssetExecutionContext` and uses the future-import anyway.

A reader following "the local pattern" cannot tell which one is the rule, and the two imply opposite things about whether annotating `context` alongside the future-import is allowed. Since this is `CLAUDE.md` — instructions agents execute rather than prose humans skim — ambiguity here propagates into code. Naming the actual rule would fix it: the import is fine as long as every name in a Dagster-inspected signature is importable at module scope, with `agency_private_capital/asset.py` as the exemplar and the unannotated-`context` files described as a separate, older style rather than as the same pattern.

### Minor

`scan_agent_definition_routes` accepts a wrapper whose `developer_instructions` merely *contains* `.claude/agents/<name>.md`. A wrapper that mentions two role files passes while routing ambiguously. Unlikely given the wrappers are five lines, and the missing/misrouted cases are covered, so this is a note rather than a request.

### Mergeability

Clean into `main`, and clean alongside #518 — I merged the pair and ran the stricter guard against the result before #518 landed; it passed.

Unrelated flake worth knowing about: `tests/unit/loaders/test_neo4j_client.py::TestNeo4jClientBatchRelationships::test_batch_create_relationships_with_failures` failed once for me in a full parallel `-m "not slow"` run, then passed 3/3 in isolation and 2/2 across its own directory. It's order- or parallelism-dependent and lives on `main`, not in this PR — flagging only so a red shard on it isn't mistaken for a real regression.

---

PR #522, opened 2026-08-03
Build Phase III control matches

hollomancer:
## Summary

- builds an exact-UEI February contract-extraction frame for the SBIR and screened-control arms
- derives the five approved matching covariates through one arm-blind code path
- conservatively excludes any candidate carrying an FPDS `SR1`–`SR3` or `ST1`–`ST3` code anywhere in the extracted history
- matches up to three controls per treated firm by exact covariate stratum, without replacement
- persists source coverage, match counts, and level-wise standardized mean differences before enforcing the approved 0.1 balance stop gate

## Boundary

This PR is pre-outcome only. It does not run the census filter on either arm, compute criteria-met distributions, calculate overlap, quote a Phase III count, or modify the weighted scoring path. It reads `research` only to conservatively exclude any exact-UEI firm with an observed FPDS SBIR/STTR code across the full history; the materializer records that eligibility-only access separately from the earliest-contract matching fields and requires explicit review of coverage and balance artifacts before Phase 2 outcomes proceed.

The February mirror identity is pinned to the same archive ETag, TOC digest, 374-column schema digest, member CRC/SHA-256, and physical FPDS relation used by the verified Phase 1 census. The exact Phase II artifact is independently checksum-verified so the treated cohort is not reconstructed approximately.

## Validation

- `81 passed` across the Phase III negative-control unit suite and the two materialization scripts
- Ruff check passed for all changed Python files
- no stochastic selection; deterministic sorted allocation maximizes one control per treated firm before assigning second or third controls

## Review gate

Keep this PR in draft until the February extraction completes and the persisted coverage, matching-summary, and SMD tables are reviewed. No arm outcome should be built from this branch before that checkpoint.


hollomancer:
Pre-outcome production gate completed on the pinned February mirror. The provenance-verified extraction contains 82,513,002 transaction rows. Exact matching retained 1,029 pairs covering 712 of 5,539 match-eligible treated firms; 4,827 eligible treated firms had no exact control. All retained-pair SMDs are 0 by construction and no level exceeded 0.1. No census outcome filter was invoked. The complete eligibility, coverage, matching, balance, and artifact-hash tables are recorded in studies/phase-iii-census/control-matching-audit-2026-08-03.md. Per the frozen design, work stops here pending a decision on whether the limited treated-firm coverage is adequate.

PR #530, opened 2026-08-04
Implement solicitation substrate and bulk award linkage

hollomancer:
## Summary

- Add a lossless SBIR.gov solicitation normalizer with separate solicitation-version and hierarchical topic/subtopic tables while preserving the weekly-report interface.
- Use the existing versioned SBIR.gov bulk award download, not the unavailable solicitation endpoint, for initial exact award-to-solicitation/topic linkage.
- Add a fail-closed bulk linkage builder that validates the reviewed 42-column schema and source metadata hash, writes deterministic Parquet assertions, and manifests coverage by agency and award year.
- Keep Award Title and Abstract typed as award text; never relabel either as solicitation title, topic description, attachment text, or government requirement evidence.
- Add synthetic source-shape fixtures, focused tests, and a pinned research status report. Generated research artifacts remain outside version control.

## Why

The solicitation endpoint is currently unavailable, but the official bulk award export already carries solicitation numbers and topic codes for many award records. Making the endpoint a prerequisite unnecessarily blocked award linkage. The bulk snapshot is already downloaded with versioned metadata in the repository data workflow, so it provides a reproducible initial source while preserving a separate evidence gate for source-native solicitation text and attachments.

## Pinned bulk result

The 2026-08-03 award_data.csv snapshot was analyzed in full:

- 219,503 award rows and 42/42 reviewed columns
- 115,148 rows with a solicitation number; 115,147 unique link assertions after one exact duplicate was removed
- 107,848 rows, or 49.1%, with both solicitation number and topic code
- NSF award years 2022-2025: 1,613 of 1,614 rows, or 99.9%, with both identifiers, spanning 16 exact solicitation numbers
- deterministic assertion artifact SHA-256: 85465dc541ba34f058be3c53aa2bc1beddc7e1622d739751ee7dad0527706d72

Historical NSF coverage changes sharply by award year, so the report presents recent and historical cohorts separately rather than extrapolating the recent rate backward.

## Impact

This supports an initial, source-traceable current-NSF award cohort and supplies exact identifiers that can seed later retrieval of official NSF funding-opportunity publications. It does not add a Dagster schedule, graph loader, document crawler, or requirement classifier.

## Evidence boundaries

- Topic-only award rows are measured but not emitted as exact award-to-solicitation assertions.
- A solicitation identifier does not establish contract use, criticality, or a physical supply-chain dependency.
- Bulk award titles and abstracts are not solicitation or attachment text.
- NSF publication, SAM.gov attachment, and Grants.gov document adapters remain separate no-go gates pending manifested samples.
- FOCI remains excluded.
- DoD-14 and NDIS-8 crosswalks remain deferred until an authoritative mapping exists.

## Validation

- Review follow-up: analysis date is required, and a no-go run quarantines any stale Parquet at a hash-labeled superseded path.
- Full 219,503-row bulk snapshot: adapter go; source hash and size matched metadata.
- Focused solicitation, bulk-linkage, weekly-report, and Dagster job tests: 73 passed.
- Ruff lint and format checks on all changed Python: passed.
- mypy on the solicitation extractor: passed.
- Architecture-boundary checks: passed.
- Git diff whitespace check: passed.


hollomancer:
## Review at `fce9e5d` — the first artifact in this line that actually enforces its source hash; two findings, one should-fix

Reviewed the linkage builder line-by-line, the extractor diff, and the test suites (38 pass locally across the three changed test files). Merges clean.

### What this gets right — and two things worth naming because they fix repo-wide patterns

- **The source SHA gates instead of decorating.** `_source_metadata` compares the sidecar's recorded sha256 *and* byte size against the actual file, and a mismatch is a **blocker** that prevents materialization (`build_sbir_bulk_solicitation_links.py:598-600`), with `require_metadata=True` by default and an explicit `--allow-missing-metadata` escape hatch. Prior artifacts in this territory displayed the hash and enforced nothing; this is the pattern the evidence tier actually requires, implemented for the first time in this pipeline family.
- **The dedup is provably deterministic, and the reason is elegant.** `link_assertion_id` hashes five fields *including* `award_record_sha256` — which is itself a fingerprint over all 42 observed columns (`:218-219`). So any partition containing duplicates contains only byte-identical rows, and the `ROW_NUMBER()` tie-break (`:330-336`) cannot select a "wrong" survivor because there is no distinguishable survivor to select. The "one exact duplicate removed" claim in the PR body is guaranteed by construction, not by luck of the snapshot.
- **The fail-closed ladder is complete**: schema mismatch, row floor, metadata mismatch, zero links, and non-unique assertion IDs each independently block; links materialize only at zero blockers; `main` exits 2 on `no_go`. The manifest carries its own limitations list, and topic-only rows are measured but never emitted as exact links — the evidence-boundary discipline in the PR body is real in the code.
- Injection-safe SQL helpers, `ORDER BY link_assertion_id` in the COPY for stable parquet ordering, atomic writes throughout, and extractor tests that pin losslessness (`test_documented_fixture_round_trips_without_field_loss`), hierarchy, legacy-view compatibility, and schema-drift auditing.

### Medium (should-fix) — `--analysis-date` defaults to `date.today()`, which breaks the PR's own determinism claim

`parse_args` at `:670`: `default=date.today()`. But `analysis_date` is stamped into **every assertion row** as a column (`:319`), so rerunning the identical command on the identical snapshot tomorrow produces a byte-different parquet with a different SHA — and the PR body pins "deterministic assertion artifact SHA-256: `85465dc5…`" as a headline property. That SHA reproduces only if the operator happens to pass the same `--analysis-date`, which nothing requires or records as required.

This is the third appearance of this exact pattern in the repo (the concentration baseline's `--as-of`, the defense-lineage builder before its fix), and the fix is the same: make `--analysis-date` **required** with no default. Everything else in this builder refuses to guess; the run date shouldn't be the one exception. One line, plus updating the two script tests that rely on the default.

### Low — stale parquet survives a `no_go`

When blockers fire, `_materialize_links` is skipped — but a parquet from a previous successful run at the same default path is left in place, while the freshly written manifest beside it says `materialized: false`. A consumer who reads `award_solicitation_link_assertions.parquet` without consulting the manifest gets stale assertions under a manifest that appears to disown them. Cheap fix: on `no_go`, rename the existing parquet to `*.superseded` (or delete it), so the filesystem state agrees with the manifest. A test asserting "no_go run leaves no readable links parquet from a prior go run" would pin it.

### Noted as deliberate, not defects

- `matches_expected` requires exact column-**order** equality with the reviewed schema (`:139`), so an upstream column reorder — not just a schema change — triggers `no_go`. That's strict, but strict is the right default for a source you don't control; the blocker message says what happened.
- `MINIMUM_SOURCE_ROWS = 50` is a low floor, but combined with the metadata hash gate it's guarding against truncated files, not sampling adequacy — fine for its purpose.
- The NSF 2022+ vs historical stratification in the manifest matches the PR body's refusal to extrapolate the recent 99.9% rate backward — the coverage query structure enforces the separation rather than leaving it to prose.

### Scope

Not audited: the 378-line evidence-plan doc beyond spot checks, and `audit_sbir_solicitation_source_coverage.py` beyond its passing tests. The builder, the extractor's normalization core, and the determinism/enforcement claims — the load-bearing parts — are what this review covers.

---

hollomancer:
Addressed both actionable findings in 693791a2:

- --analysis-date is now required, so the date embedded in every assertion is an explicit reproducibility input.
- A no-go run now moves any prior links Parquet to a hash-labeled .superseded path and records that quarantine in the manifest.
- Added regression coverage for the required CLI argument and go-to-no-go stale-artifact transition.

Validation: 73 focused tests passed; Ruff, mypy, architecture boundaries, and the full 219,503-row pinned snapshot all pass. The reproduced assertion SHA remains 85465dc541ba34f058be3c53aa2bc1beddc7e1622d739751ee7dad0527706d72.

PR #516, opened 2026-08-03
docs: consolidate current architecture and operations guidance

hollomancer:
## Summary

- make the default install target provision the complete local workspace
- replace retired AWS, S3, Lambda, and GitHub Actions operating guidance with the Mac mini, local SSD, Dagster, and Tailscale model
- establish canonical owners for architecture, deployment, testing, configuration, specifications, data quality, and evidence status
- archive completed migration plans and legacy performance prose while deleting redundant runbooks and quick references
- distinguish computable, reproducible, validated, and citable research claims
- document current CET and transition boundaries, including the missing award-training loader and the difference between the Dagster scoring path and the richer ML library
- extend repository hygiene checks to validate Markdown anchors as well as linked files

## Why

The documentation accumulated overlapping sources of truth from several architecture eras. That made stale cloud topology, test commands, thresholds, and research-status language appear current even when newer guides were correct. This change assigns each fact to one maintained owner and preserves dated material under `docs/archive/`.

## Impact

- the active documentation surface is about 6,900 lines smaller
- live operations route through one Mac mini runbook
- CI and local testing route through one testing index and scheduling guide
- historical reports and ADRs retain provenance without acting as operational guidance
- policy briefs and technical findings remain separate because they serve different audiences

## Validation

- `uv run python scripts/ci/check_removed_src_references.py`
- `uv run pytest tests/unit/scripts/test_repository_hygiene.py -q` (7 passed)
- Ruff lint and format checks for the updated guard and tests
- `git diff --check`
- Docker Compose configuration validation
- full-stack workspace import smoke test
- 159 configuration tests
- 4 suite-integrity tests
- help smoke tests for every documented CLI

PR #534, opened 2026-08-04
Improve developer onboarding

hollomancer:
## Summary

Improve the fresh-clone developer experience without changing pipeline or research behavior.

- add `make doctor` and a 33-test, data-free `make test-smoke` target
- add `CONTRIBUTING.md` with repository scope, package boundaries, and verification guidance
- separate host-local and Compose-internal Neo4j addressing and unify development credentials
- rewrite the getting-started tutorial around `make sample-data` before the first Dagster materialization
- correct stale testing commands and document package-index versus repository-checkout installation
- reposition `scripts/setup_dev.sh` as a CI-compatibility path aligned with the `stack-dev` environment

## Developer impact

A new contributor now has one canonical sequence:

```bash
make install
cp .env.example .env
make setup-local
make doctor
make test-smoke
make sample-data
make dev
```

The tutorial's first `raw_sbir_awards` materialization uses ten generated awards and does not require external credentials or Neo4j. Host Python connects to `bolt://localhost:7687`; Compose services use `bolt://neo4j:7687`.

Package-index consumers retain normal installation commands:

```bash
pip install sbir-analytics
pip install sbir-etl
```

## Validation

- `make test-smoke` — 33 passed
- `make docs-check` — passed
- `docker compose --env-file .env.example --profile dev config --quiet` — passed
- `bash -n scripts/setup_dev.sh` — passed
- `git diff --check origin/main...HEAD` — passed
- GitHub CI, CodeQL, version policy, and setup-script verification — passed at `fe3d76f`

## Scope

This PR contains onboarding and developer-documentation changes only. The full-stack behavior of `make install` was already present on `main`; this PR adds diagnostics, smoke coverage, clearer environment boundaries, and a reproducible first-run path.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,11 @@
 # -----------------------
 # Neo4j authentication (username/password). Use a strong password for real deployments.
 NEO4J_USER=neo4j
-NEO4J_PASSWORD=change_me
+NEO4J_PASSWORD=test
+
+# Host-local processes connect through the published Docker port. Compose
+# overrides this with the internal service hostname for container processes.
+NEO4J_URI=bolt://localhost:7687
 
 # Neo4j plugins and memory tuning (optional examples)
 # Use JSON array notation for plugins; keep as example values here.
@@ -30,11 +34,11 @@ DOCKER_IMAGE_NAME=sbir-etl:latest
 # -----------------------
 # SBIR ETL application-specific variables
 # -----------------------
-# Canonical Neo4j Bolt URL used by the application (compose service host in dev)
-SBIR_ETL__NEO4J__BOLT_URL=bolt://neo4j:7687
+# Canonical Neo4j Bolt URL used by host-local application processes
+SBIR_ETL__NEO4J__BOLT_URL=bolt://localhost:7687
 
 # Individual connection pieces (application may prefer either URL or these pieces)
-SBIR_ETL__NEO4J__HOST=neo4j
+SBIR_ETL__NEO4J__HOST=localhost
 SBIR_ETL__NEO4J__PORT=7687
 SBIR_ETL__NEO4J__USERNAME=${NEO4J_USER}
 SBIR_ETL__NEO4J__PASSWORD=${NEO4J_PASSWORD}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,86 @@
+# Contributing
+
+Thanks for helping improve the SBIR/STTR commercialization analytics project.
+This is an experimental research codebase, so a useful contribution should make
+a research result more trustworthy, reproducible, or understandable without
+adding unnecessary operational complexity.
+
+## Before You Start
+
+Read the [research questions](docs/research-questions.md) first. They are the
+scope gate for features and analyses. For implementation work, also review the
+[architecture overview](docs/architecture/detailed-overview.md) and any relevant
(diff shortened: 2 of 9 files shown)


hollomancer:
## Review at `fe3d76f` — approvable; one paperwork fix before merge

Two review rounds: the first caught that the original head (`07b8e11`) was built on a **July 13 fork point** — its central fix (full-stack `make install`) had already landed on `main` via #516, six files conflicted, and the doc edits targeted pages that #516/#519 had since rewritten. The current head is a proper rebase onto today's `main` (v0.2.0): the dead install hunk is gone, the Makefile now only *adds* targets, and `mergeable_state` is clean. This comment reviews what's actually left.

### Verified against current `main`, not taken from the PR body

- **The strict hygiene guard passes** (`check_removed_src_references.py` — the post-#519 version with link *and* anchor validation, which didn't exist at the original fork point). The rewritten docs survive the rules that will actually gate them.
- **`docker compose --env-file .env.example --profile dev config --quiet` validates.**
- **Smoke suite: 33 passed** (`pytest -n 0` on the two chosen files). Note this is 33, not the body's 31 — `main` gained two asset-discovery tests since the body was written.
- **Every Make target the docs reference exists** (`lint`, `docs-check`, `sample-data`, `setup-local`), and the `make check` → `make lint` + `make docs-check` substitution in `docs/testing/index.md` is a *correction*, not a preference — `check` no longer exists on `main`, so the old doc pointed at a dead target.
- **`make doctor` is built right**: `uv run --no-sync` throughout, so the diagnostic never mutates the environment it's diagnosing; the import probe covers all four first-party packages and passes here.
- **The Neo4j host/container split** is the best fix in the PR — `.env.example` says `localhost:7687` for host processes, compose hardcodes the `neo4j` service hostname for containers, with the reasoning commented at both sites. That kills a classic newcomer trap (host process dialing `bolt://neo4j:7687`). The rebase also unified the dev credential story: the dev-profile fallback is now `test`, matching `.env.example` — the fallback `password` remains only in the common block other profiles use.
- **`CONTRIBUTING.md` is genuinely new** (absent on `main`), and its command sequence agrees with the getting-started page and the testing index — the three documents were clearly written against each other rather than independently.
- `setup_dev.sh` is honestly repositioned as a CI-compat script pointing newcomers at `make install`, and switched to `--extra stack-dev` so the two install paths produce the same environment instead of silently diverging.

### The one remaining item — the PR body describes the pre-rebase PR

The body wasn't updated with the rebase, so the merge-commit record would misdescribe the change:

- The first Summary bullet claims this PR makes `make install` install the full stack — that's pre-existing `main` behavior (#516); this diff doesn't touch the install targets.
- The whole "Why" section narrates the already-fixed install defect as the motivation.
- Validation numbers are stale: "31 passed" is now 33; the 138-package install and "364 links checked" were measured on the July tree under the old link rules.

Suggested body, roughly: *add `make doctor` and `make test-smoke`; add `CONTRIBUTING.md`; separate host-local vs Compose-internal Neo4j addressing and unify dev credentials; rewrite the getting-started tutorial around `make sample-data` before first materialization; correct dead doc references (`make check`, package README paths); reposition `setup_dev.sh` as CI-compat aligned to `stack-dev`.* Two minutes of editing, no code change — and with it, this has my approval. Blast radius is onboarding only; nothing here touches pipeline or research code.

---

PR #535, opened 2026-08-04
feat: add Massachusetts unmanned-systems SBIR/STTR report

hollomancer:
## Summary

- add a versioned physical unmanned-systems manufacturing profile covering SBIR and STTR awards across aerial, ground, surface, and undersea domains
- add normalized state filtering and preserve city, state, UEI, and DUNS in tech-census audit output
- expose the profile through the existing tech-census CLI and analytics tool
- publish the reviewed Massachusetts readout: 76 awardees, 238 awards, and $109,526,154.89

## Why

The existing `drone_manufacturing` profile intentionally covers aerial SBIR hardware only. This adds a separate profile for the broader question without changing that established definition, and makes state-scoped reports reusable through the existing census tool.

## Compatibility with current main

The branch is rebased onto `v0.3.0`. Snapshot publishing and private-API changes from the original branch were dropped because #536 retired that interface before this PR merged. The census profile, state/audit support, CLI/tool path, report, and their focused tests remain.

## Verification

- focused tech-census tests: 56 passed
- Ruff checks on all changed Python and test files: passed
- `make docs-check`: passed
- `git diff --check`: passed
- real-data candidate run from the original review reproduces 334 Massachusetts awards totaling $181,483,765.28
- report list checks at 26 platform/integrator firms plus 50 component/hardware firms

## Evidence boundary

The 238-award figure is the conservative human-reviewed result, distinct from the 334-award deterministic candidate stage. The original award-level review decisions were not persisted as a machine-readable ledger; the report calls this out explicitly so the reproducibility boundary is clear.

PR #536, opened 2026-08-04
Remove private analytics API

hollomancer:
## Summary

- remove the private FastAPI service, API-only snapshot store/publisher, console entry point, and tests
- remove FastAPI/Uvicorn dependencies plus local/server Compose services, bearer-token configuration, health checks, and the Tailscale 8443 route
- update current architecture and deployment guidance, and supersede the API-first ADR with a consumer-driven interface decision

## Breaking change

This removes `sbir-analytics-api`, `sbir_analytics.api`, the `SBIR_ANALYTICS_*` configuration contract, and the private HTTP endpoints. Existing Mac mini deployments should inspect `tailscale serve status` and remove the legacy 8443 route only if it still targets `http://127.0.0.1:8010`; the runbook includes the migration step.

## Validation

- `uv run pytest -q tests/unit/test_server_env_defaults.py tests/unit/test_server_ops.py tests/unit/test_server_prerequisites.py tests/unit/test_server_runtime_contract.py` — 50 passed
- `uv run pytest -q tests/unit/scripts/test_identity_boundaries.py` — 6 passed after staging the deletions
- CI-scoped Ruff — passed
- architecture boundary check — passed
- `uv lock --check` — passed
- `make docs-check` — passed
- server Compose config renders exactly `neo4j`, `dagster-code-server`, `dagster-webserver`, and `dagster-daemon`
- shell syntax and staged diff checks — passed

Full unit-suite audit reached 5,642 passed and 7 skipped. One remaining unrelated failure is caused by the ignored local `data/raw/sam_gov/sam_entity_records.parquet` fixture overriding a file-not-found test. The other initial failure was the deletion-sensitive identity scan and passes after staging, as recorded above.


hollomancer:
## Review at `e565a26` — a clean, complete amputation; approvable, with one suggestion and one blocking interaction with #535

For a deletion PR the review question is completeness, and this one proves out:

- **Zero orphaned references.** Repo-wide grep for `sbir_analytics.api`, `sbir-analytics-api`, `SBIR_ANALYTICS_API`, `analytics-api`, `tech_census_snapshot`, and stray FastAPI/Uvicorn imports comes back clean. The only surviving `8443`/`8010` mentions are exactly where they belong: the migration step in the runbook (`mac-mini-server.md:138-140`, with the correct "do not remove a route that has a different target" caveat) and the ADR record.
- **The four named suites pass (50 tests)**, the architecture and hygiene guards pass, and `docker compose -f docker-compose.server.yml --env-file .env.server.example --profile server config --services` renders **exactly** `neo4j`, `dagster-code-server`, `dagster-daemon`, `dagster-webserver` — the API is gone from the rendered config, not merely the file list.
- **The tailscale script edit is coherent end to end**: 8443 removed from route management, `CREATED_8443`/rollback state removed, and the Neo4j tailnet port-collision validator correctly narrowed from `{443, 8443}` to `{443}` — the secondary edit deletions usually miss.
- **ADR-004 is the right kind of decision record.** It names the real carrying costs (separate runtime, auth secret, dependency set, health check, ingress route — no current consumer), supersedes ADR-003 with links in both directions, and replaces the retired "API before MCP" rule with "any future interface starts from a concrete consumer and defines its own least-privilege contract" — so the deletion doesn't quietly re-create the old default the next time someone wants an adapter.

### Suggestion (non-blocking)

The legacy-route migration is manual: the operator inspects `tailscale serve status` and removes 8443 only if it targets `127.0.0.1:8010`. The script's own `route_state` helper exists precisely to make check-ownership-then-remove safe — a one-time cleanup in `cmd_up` (remove 8443 iff owned and targeting 8010; warn and leave it otherwise) would eliminate a hand-run command on the live host and use machinery that already has tests. Defensible either way given the runbook discipline here.

### Blocking interaction — #535 modifies five files this PR deletes

`packages/sbir-analytics/sbir_analytics/api/snapshots.py`, `scripts/data/build_tech_census_snapshot.py`, both their test files, and `docs/architecture/private-analytics-api.md` are all modified by #535 and deleted here — a guaranteed modify/delete conflict for whichever merges second, and the conflict is semantic, not textual: #535's summary includes exposing a new profile "through the private analytics API" that this PR retires with a same-day accepted ADR. The two PRs disagree about whether the API exists. Recommended order: settle the ADR question (this PR) first; #535's profile, state filtering, census CLI, and readout survive untouched — its API wiring is roughly 25 of its 777 lines. Full detail on the #535 thread.

---

hollomancer:
Addressed in `38b39785`:

- `cmd_up` now removes HTTPS 8443 only when it still targets the retired analytics API at `http://127.0.0.1:8010`.
- If 8443 has been reassigned, the helper warns and leaves it untouched.
- Added behavioral tests for both removal and preservation paths, and updated the runbook.
- Fixed the Ruff-format CI failure in `test_server_runtime_contract.py`.

The #535 interaction is being resolved by sequencing: #536 will merge first; #535 must then rebase onto `main` and drop its retired API/snapshot wiring.

PR #537, opened 2026-08-04
docs: generalize self-hosted server runbook

hollomancer:
## Summary

- rename the hardware-specific deployment runbook to `self-hosted-server.md`
- describe the live deployment in terms of its operating contract instead of a particular computer model, CPU architecture, checkout path, or storage mount
- move installation-specific paths and current state into an ignored `server-status.local.md`, with a tracked example template
- update documentation links, operational comments, and archived deployment references

## Why

The deployment documentation mixed portable safety requirements with details of the current host. That made the documented architecture appear tied to a Mac mini and encouraged host-specific paths to spread across the repository.

This keeps the important constraints explicit—dedicated clean checkout, persistent storage, preserved `.env.server` and Dagster metadata, loopback-only binds, Tailscale-only ingress, and no Funnel—while allowing the host implementation to change independently.

## Validation

- `make docs-check`
- `git diff --check`
- repository-wide scan confirms no Mac mini-specific wording or stale renamed anchors remain

### DIFF
--- a/.env.server.example
+++ b/.env.server.example
@@ -1,4 +1,4 @@
-# .env.server.example — Tailscale-only Mac mini server profile
+# .env.server.example — Tailscale-only self-hosted server profile
 #
 # Copy to `.env.server` and fill in real secrets. Never commit `.env.server`.
 #   cp .env.server.example .env.server
@@ -16,7 +16,7 @@ ENVIRONMENT=prod
 PYTHONUNBUFFERED=1
 SERVICE_STARTUP_TIMEOUT=120
 
-# Image coordinates (published multi-arch by CI; pull the arm64 manifest on M-series).
+# Image coordinates. Use a manifest compatible with the server host's architecture.
 IMAGE_NAME=ghcr.io/hollomancer/sbir-analytics
 IMAGE_TAG=latest
 
@@ -45,8 +45,8 @@ NEO4J_server_memory_heap_initial__size=512M
 NEO4J_server_memory_pagecache_size=512M
 
 # -----------------------------------------------------------------------------
-# Storage — defaults are repository-local. On the Mac mini, point these at the
-# external SSD, e.g. /Volumes/SSDmini/sbir-analytics/... (see the device guide).
+# Storage — defaults are repository-local. On the self-hosted server, point these at
+# durable host storage (see the server runbook).
 # -----------------------------------------------------------------------------
 SERVER_DATA_DIR=./data
 SERVER_REPORTS_DIR=./reports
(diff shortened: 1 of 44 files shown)


hollomancer:
## Review at `04c6939` — the generalization keeps every safety constraint tracked; one overreaching claim, one missing migration step, two collision warnings

Verified locally: the strict hygiene guard passes (it validates anchors, so the rename's new `#live-instance-on-the-server-host` anchor and all 40+ retargeted links are machine-checked, not just claimed), and `git check-ignore` confirms the new local-file name is ignored.

### What this gets right

- **The CLAUDE.md edit preserves the contract and delegates only the coordinates.** Every load-bearing rule stays tracked: operate only from the dedicated deployment checkout, never a development checkout; preserve `.env.server`, persistent data, and the `dagster_home` volume; tailnet-only ingress; never Funnel. What moves to the ignored local file is exactly what should never have been tracked — a personal home-directory path and a volume name. The safety posture is unchanged; the blast radius of a hardware change is now one ignored file.
- **The ignore pattern is smarter than the diff looks.** `/docs/deployment/*-status.local.md` covers both the new `server-status.local.md` *and* a live host's existing `mac-mini-status.local.md`, so an in-place upgrade doesn't suddenly show an untracked file and fail the "deployment checkout is clean" preflight.
- **The example template is complete and ends with the right prohibition** (no secrets; those stay in mode-`0600` `.env.server`) — the one line that prevents the local status file from becoming a shadow credential store.

### Medium-low — the validation claim overreaches the diff

The PR body states: "repository-wide scan confirms no Mac mini-specific wording … remain." That's not true of the tree at this head — the Mac-specific storage mount survives in seven files:

```
scripts/data/download_sam_gov.py:28        --dest /Volumes/SSDmini/sbir-analytics/...
scripts/data/download_uspto.py:8,271       --local /Volumes/SSDmini/sbir-analytics/...
scripts/data/download_sbir.py:11           --dest /Volumes/SSDmini/sbir-analytics/...
scripts/usaspending/download_database.py:11
tests/unit/assets/test_source_download_jobs.py:71-72   (asserts the SSDmini path round-trips)
tests/unit/utils/test_cloud_storage.py:33-34
```

Functionally harmless — these are docstring example invocations and env-var echo tests using it as an arbitrary value — but they are precisely the "host-specific paths spread across the repository" the Why section says this PR exists to stop, and the docstring examples are what an operator copy-pastes. Either sweep these seven (generic `/path/to/persistent-storage/...` examples; any placeholder in the tests) or narrow the validation claim to documentation and operational comments. The sweep is ~10 lines and makes the claim true.

### Low — no rename step for the live host's existing status file

The glob keeps the old `mac-mini-status.local.md` *ignored*, but nothing tells the operator it exists: CLAUDE.md and the runbook now reference only `server-status.local.md`, so after this merges, the live host has its real state in a file no documentation mentions, while the documented file doesn't exist yet. One line in the runbook's upgrade notes — "if `mac-mini-status.local.md` exists from a previous deployment, rename it to `server-status.local.md`" — closes the gap. (Precedent: #536 handles its legacy 8443 route with exactly this kind of explicit migration step.)

### Collision warnings — two open PRs edit the file this PR renames

- **#536** edits `docs/deployment/mac-mini-server.md` (the analytics-API migration step at old `:138-140`). Rename detection will usually carry the edit across, but whichever merges second should verify the hunk landed in `self-hosted-server.md`.
- **#533** (stacked on #523) adds ~90 lines to the same runbook documenting "a dated **Mac mini** canary" — content whose vocabulary this PR's ethos removes. When #533 rebases onto a `main` containing this rename, the section needs rewording to the generalized terms, not just relocation.

Neither blocks this PR; both are surprises waiting for whoever merges second, so they're on the record here.

---

PR #539, opened 2026-08-04
refactor: promote canonical SBIR award grain

hollomancer:
## What changed

- promote the immutable `sbir-source-v2` award-grain contract into `sbir_etl.identity.sbir_awards`
- add a canonical, chunk-bounded SBIR.gov award materializer in `sbir_etl.extractors.sbir_public_awards`
- retain `normalize_awards` as a compatibility entry point over the package pipeline
- migrate the procurement packet, config-driven technology census, and tech-area cohort loader to the shared materialization
- key tech-area deduplication and method overlap on `award_key`, not tracking/contract number or mutable amount fields

## Why

Award identity was private to the procurement report while other reports loaded the same source independently and defined different grains. That allowed reused tracking or contract numbers to collapse distinct awards and allowed mutable amount fields to participate in deduplication.

The identity behavior was already versioned. This change makes that contract load-bearing and reusable, and separates identity from deterministic source materialization.

## Compatibility and impact

- The existing award key and procurement `row_hash` are byte-identical for the same source row.
- Public `award_id` remains human-readable; `award_key` owns joins, counts, and deduplication.
- CSV identifiers remain strings, including leading zeroes.
- Raw CSV memory is bounded by a 50,000-row chunk while source-edition collapse remains independent of chunk boundaries.
- Tech-area composition now reports `duplicate_award_key_rows` instead of the misleading `duplicate_award_id_rows`.

## Validation

- 97 focused identity, extraction, procurement-output, census, and tech-area tests passed
- additional procurement wiring/AI tests passed; the optional full ML test requires the locally absent `scikit-learn` extra and is left to CI
- Ruff check and format check passed on all changed Python files
- MyPy passed on the primitive, materializer, and migrated census loader
- company identity boundary guard passed
- `make docs-check`
- byte-for-byte comparison of the pre/post extraction award key and row hash matched
- `git diff --check`

PR #529, opened 2026-08-04
docs: plan deterministic transition candidate assertions

hollomancer:
## Summary

- Add a repository-audited, three-PR implementation plan for replacing inferred causal topology with candidate assertions.
- Establish one canonical federal prime-contract award resolver: `generated_unique_award_id`, then an explicitly typed legacy agency/parent-IDV/PIID composite, then unresolved failure. Keys are namespaced as `USAID:` or `LEGACY:`; bare PIID is audit-only.
- Resolve assertion grain as Phase II source record × federal prime contract award, with contract actions retained as first-class supporting observations.
- Separate the detector-selected action, earliest award action, and earliest positive-obligation action. Preserve signed latency; a missing Phase II end produces typed temporal absence rather than dropping the assertion.
- Introduce typed identity, technology, and temporal assessments so unmeasurable or failed evaluation is not confused with a measured zero.
- Resolve the package boundary by narrowly permitting `sbir_graph` to consume `sbir_etl.assertions` while retaining exact-prefix architecture enforcement.
- Keep V1 candidate-only: `CANDIDATE`, support class C, and `INVESTIGATIVE_ONLY`.

## Shared boundary

The assertion producer and census use the same normalized UEI-only action-pair builder, contract-key resolver, and transaction identity, then apply independent downstream rules. A dedicated census projection preserves its audit-facing `target_id` without making PIID authoritative. The selected normalized census source keeps its existing generated-ID requirement, the hash-frozen census design is not edited, and assertion production imports no census predicates.

## Migration scope

PR 1 defines the canonical contract and content-addressed snapshot. PR 2 publishes assertion topology and deletes legacy `TRANSITIONED_TO` and `RESULTED_IN` edges after backup and verification. PR 3 proves DuckDB consumption and migrates remaining consumers. Review decisions, accepted claims, source-record graph nodes, and additional assertion families remain deferred until a consumer exists.

## Impact

This PR changes documentation only. It does not add production models, modify Neo4j, alter Dagster schedules, change frozen census artifacts, or touch the live deployment.

## Validation

- `make lint-boundaries`
- `make docs-check`
- `git diff --check`
- independent scope and shared-consumer audits: no remaining material contradiction

PR #540, opened 2026-08-04
refactor: promote versioned jurisdiction identity

hollomancer:
## What changed

- add a versioned U.S. jurisdiction identity primitive covering states, DC, and territories
- expose strict and legacy-permissive profiles instead of leaving divergent behavior unnamed
- migrate validation, fiscal refresh, patent transformation, geographic resolution, UCC matching, Neo4j loading, and nanotechnology liveness analysis to the shared contract
- preserve the patent transformer's historical two-letter pass-through and first-prefix behavior only through the named permissive profile
- make strict UCC matching reject unknown jurisdiction text instead of fabricating its first two letters
- extend the identity boundary guard to reject new full jurisdiction maps outside the primitive

## Why

The repository carried independent state tables in validators, transformers, matching, enrichment, graph loading, and research scripts. Their territory coverage, aliases, partial matching, and unknown-value behavior differed invisibly.

This promotes the concept—not merely the table—by making behavior named, immutable, tested, and mechanically single-source.

## Impact

- Canonical codes and names round-trip under `us-jurisdiction-strict-v1`.
- Explicit aliases such as `D.C.`, `Washington DC`, and `U.S. Virgin Islands` are accepted.
- Unknown and partial values fail closed under the strict profile.
- Existing `VALID_US_STATES` and GeographicResolver mapping attributes remain compatibility views over the primitive.
- Archived scripts are intentionally excluded from the active-code boundary.

## Validation

- 317 affected identity, validator, patent, fiscal, UCC, geographic-resolver, and boundary tests passed
- Ruff check passed on all changed Python files
- Ruff format check passed on package, guard, and test files; the exploratory nano script retains its existing formatting
- MyPy passed on the primitive, fiscal/UCC callers, and boundary guard
- identity boundary guard passed with the new file tracked
- `make docs-check`
- `git diff --check`

The Neo4j unit module requires the locally absent Dagster extra; its import and behavior remain covered by CI.

### DIFF
--- a/docs/steering/structure.md
+++ b/docs/steering/structure.md
@@ -86,6 +86,10 @@ docs/
 Company-name normalization and similarity belong in `sbir_etl/identity`; callers select
 an explicit versioned profile. See [company-identity.md](company-identity.md).
 
+U.S. state, district, and territory normalization belongs in
+`sbir_etl.identity.geography`. Callers use the strict profile unless they explicitly need
+the named permissive compatibility behavior; no caller carries its own jurisdiction map.
+
 ### Study Evidence Contracts
 
 Externally citable studies declare a versioned contract in `studies/<study-id>/study.yaml`.
--- a/packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
@@ -19,6 +19,7 @@ from dagster import (
 from loguru import logger
 
 from sbir_etl.config.loader import get_config
+from sbir_etl.identity.geography import normalize_us_jurisdiction
 from sbir_etl.models.award import Award
 from sbir_etl.utils.company_canonicalizer import canonicalize_companies_from_awards
 from sbir_etl.utils.text_normalization import normalize_name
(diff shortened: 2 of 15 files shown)


hollomancer:
## Review at `1e335e9` — the profile split is exactly how to consolidate divergent behavior; verified across all six call sites

Verified locally: 231 tests pass spanning the new primitive plus the consolidated call-site suites (geographic resolver, UCC matcher, validators, identity-boundary guard).

### The load-bearing question, answered by design

Six call sites previously carried their own jurisdiction normalization, and they did **not** all behave the same — the patent transformer accepted arbitrary two-letter codes and prefix-matched names. The wrong consolidation would have silently unified them; this PR instead names the divergence: `STRICT_V1` for the five strict sites, `PERMISSIVE_PREFIX_V1` explicitly documented as "preserves the patent transformer's legacy behavior," and verified at the call site — `patent_transformer.py:550-552` passes the permissive profile, everyone else gets strict by default. That's CLAUDE.md's "add a named profile if you need new behavior" rule executed literally, and it means this refactor changes **zero** outputs while making the divergence a reviewable, versioned fact instead of six accidents.

Also right:

- The alias table and name map are `MappingProxyType`-frozen with `_V1` suffixes — the primitive's behavior is versioned in its data, not only its function names.
- The identity-boundary guard is *extended* in the same PR (+78 lines) so future jurisdiction normalization outside `sbir_etl/identity/` fails CI — the promotion locks the door behind itself.
- All removed call-site code was module-private; no public import path breaks.

### One question worth answering in the PR body

`PERMISSIVE_PREFIX_V1`'s pass-through of *any* two-letter alphabetic value means `XX` or `ZZ` normalize to themselves under that profile. That's faithful legacy preservation (the point of the profile), but now that the behavior has a name, it's worth one sentence on whether the patent pipeline should eventually migrate to strict — i.e., is the permissive profile a permanent contract or a documented bridge with an exit condition, like the architecture guard's transitional entries? Either answer is fine; unstated, the profile will outlive its reason.

**Release-train note:** new public module (`sbir_etl.identity.geography`), no removed public surface → MINOR-class. Base for stacked #541 — must merge before it. Full train analysis on #538.

---

hollomancer:
## Merge readiness: not ready — the new guard fires on `main`

CI is red here in two places (`Lint, Types, and Guards`, `Fast Tests 4/4`) and both are the same root cause. It is not visible if you only test this branch's head, which is why I want to be specific about it.

### The guard this PR introduces now catches a file that landed on `main` after the branch point

```
Company identity boundary violations were found:
sbir_etl/utils/tech_census.py:43: U.S. jurisdiction map bypasses sbir_etl.identity.geography; use a named normalization profile
```

`tests/unit/scripts/test_identity_boundaries.py::test_current_repository_obeys_identity_boundaries` fails on the same violation, which is the `Fast Tests 4/4` failure.

I ran the guards against the bare head first and everything passed — the violation only appears against the merge with `main`, which is what CI actually builds. `sbir_etl/utils/tech_census.py:43` defines a `_US_STATE_NAMES` map, and it arrived on `main` via #535 (Massachusetts unmanned systems report) after this branch was cut. Reproduced:

```
$ git merge --no-edit origin/main && uv run python scripts/ci/check_identity_boundaries.py
Company identity boundary violations were found:
sbir_etl/utils/tech_census.py:43: U.S. jurisdiction map bypasses sbir_etl.identity.geography; use a named normalization profile
```

**To do:** merge `main` and migrate `tech_census.py`'s jurisdiction map to a named profile from `sbir_etl.identity.geography`. This is the guard doing its job — a second, unnamed copy of the jurisdiction normalization is exactly the drift it exists to catch — so the fix belongs here rather than in a suppression.

Worth checking whether `STRICT_V1` covers what `tech_census.py` needs or whether it wants `PERMISSIVE_PREFIX_V1`; if neither fits, a third named profile is the right answer, not a bypass.

### The docstring fix for my earlier comment isn't on this branch

I asked for `PERMISSIVE_PREFIX_V1` to be annotated as a bridge with a stated exit condition. Commit `2930694a` does exactly that and is well written — but it sits on `copilot/address-comments-529-up`, which is attached to no pull request and does not contain this branch's commits. It carries fixes for #529, #538, #539, #541 and #542 as well, all stacked on #538's branch, and it doesn't stand alone (its `phase_iii_negative_controls/identity.py` shim imports `sbir_etl.identity.exact_awards`, which only #541 creates — `ModuleNotFoundError` on import).

Those commits need to be cherry-picked onto the branches they belong to before any of this merges.

### Otherwise

On the bare head: ruff, ruff-format, mypy (252 files), architecture boundaries, repository hygiene, study manifests and `Definitions.validate_loadable` all pass. `Neo4j Integration Tests` and shards 1–3 are green. The primitive itself looks right; it's the migration of existing callers that's incomplete.

---

hollomancer:
## Review comments addressed

- **Identity guard on main**: Merged main and migrated tech_census.py's _US_STATE_NAMES to sbir_etl.identity.geography.normalize_us_jurisdiction (STRICT_V1 profile). Removed the 71-entry duplicate jurisdiction map.
- **PERMISSIVE_PREFIX_V1 docstring**: Annotated as a documented bridge with stated exit condition (migrate patent transformer to STRICT_V1). Added frozen-profile invariant to USJurisdictionProfile.

Guard now passes on merged tree. Ready for re-review.

PR #541, opened 2026-08-05
refactor: promote exact award-key identity

hollomancer:
## Stack

Depends on #540 (`agent/promote-geography-primitive`) because both extend the identity boundary guard and exports.

## What changed

- move generic exact award-key recovery and reconciliation into `sbir_etl.identity.exact_awards`
- name the immutable behavior `exact-award-key-v1`
- stamp new attempt and reconciliation outputs with `identity_profile`
- retain compatibility with previously retained attempt audits that predate the profile column
- reduce the Phase III negative-control module to a compatibility re-export
- migrate the negative-control package's internal imports to the primitive
- extend the identity boundary guard to reject a second resolver implementation

## Why

The resolver performs generic, fail-closed identity work and is shared by USAspending and NIH adapters, but it lived inside one study's asset package. That inverted the dependency: reusable identity behavior was owned by an analytical consumer and had no named version contract.

Source adapters still own source-specific key construction and data movement. The primitive owns only exact joins, recipient-identifier connectivity, exhaustive recovery statuses, provenance aggregation, and conflict quarantine.

## Impact

- Existing imports continue to work through the compatibility module.
- New outputs record `exact-award-key-v1`.
- Retained legacy audits remain rebuildable under the explicitly selected V1 compatibility path.
- Company names, titles, abstracts, and similarity signals remain excluded from the resolver.

## Validation

- 116 Phase III negative-control and identity-boundary tests passed
- Ruff check and format check passed on the primitive, package, guard, and tests
- MyPy passed on the primitive and identity guard
- identity boundary guard passed with the promoted file tracked
- `make docs-check`
- `git diff --check`


hollomancer:
## Review at `507501d` — clean promotion with a real compatibility shim; verified against the negative-controls suite

Verified locally: 153 tests pass across `tests/unit/identity/` and the full `phase_iii_negative_controls` suite, including the new `test_identity_compatibility.py`.

### What's right

- **The shim is a shim, not a fork.** `phase_iii_negative_controls/identity.py` collapses from 338 lines to pure re-exports of `sbir_etl.identity.exact_awards` — so the disjoint-set reconciliation, fail-closed conflict quarantine, and `RecoveryStatus` vocabulary that the Phase III identity-recovery work depends on now exist in exactly one place, under the identity-boundary guard, with `EXACT_AWARD_IDENTITY_VERSION` and a profile enum making the behavior versioned. Every sibling module (`sam_eligibility`, `quarantine`, `source_keys`, `february_mirror`, `nih_reporter`) now imports the primitive path directly, so the shim is for external stability, not internal convenience.
- **A dedicated compatibility test** pins that the old import path keeps working — the thing promotions usually assert in prose and break in practice.
- The identity-boundary guard is extended in the same PR, closing the door behind the move — consistent with #540's pattern.
- This machinery survived heavy adversarial review when it lived in the negative-controls package (the fail-closed properties were verified by execution during the #515 cycle); promoting it *unchanged*, with the suite still green, is the right way to carry that earned confidence across the move.

### Two notes

1. **Frozen-spec interaction worth one sentence in the PR body.** The Phase III census/negative-controls specs are frozen with SHA enforcement, and their design docs name the package-local module as the implementation home. If any frozen artifact hashes over text that references `phase_iii_negative_controls.identity` as the implementation location, this move needs the amendment path rather than a silent relocation. The shim means *behavior* is untouched — but the spec-freeze discipline in this repo is about text, not just behavior, so stating "no frozen artifact references the moved path" (or amending) closes the question.
2. **Stacking:** based on #540's branch, so it inherits #540's review state and must merge after it. The diff-vs-540 is clean; nothing from #540 is re-modified here.

**Release-train note:** new public module (`sbir_etl.identity.exact_awards`) with the old path preserved → MINOR-class, no breaking callout needed. Merges after #540. Full train analysis on #538.

---

hollomancer:
## Merge readiness: not ready — blocked by its base (#540)

This PR's own content is in good shape. The blocker is inherited.

### Inherited: #540's identity-guard failure

This branch stacks on `agent/promote-geography-primitive`, so it carries #540's new geography guard *and* the violation that guard finds on current `main`. Reproduced against the merge with `main`:

```
Company identity boundary violations were found:
sbir_etl/utils/tech_census.py:43: U.S. jurisdiction map bypasses sbir_etl.identity.geography; use a named normalization profile
```

Details are on #540 — short version: `_US_STATE_NAMES` in `sbir_etl/utils/tech_census.py` arrived on `main` via #535 after these branches were cut, and it needs migrating to a named `sbir_etl.identity.geography` profile. Nothing to do here beyond rebasing once #540 is fixed.

Everything else passes on the merged tree: architecture boundaries, repository hygiene, ruff, mypy.

### The frozen-spec answer to my earlier comment isn't on this branch

I asked whether any frozen artifact hashes over text naming `phase_iii_negative_controls.identity` as the implementation location. Commit `3df8b073` answers it — no frozen artifact in `studies/` or `specs/` does, and it records that finding in the shim docstring so it isn't re-investigated. That's the right resolution and the right place to put it.

But that commit is on `copilot/address-comments-529-up`, not on `agent/promote-exact-award-identity`. That branch has no pull request, doesn't contain this branch's commits, and can't even be imported on its own — it ships the shim that re-exports from `sbir_etl.identity.exact_awards` without shipping `exact_awards.py`:

```
$ python -c "import sbir_analytics.assets.phase_iii_negative_controls.identity"
ModuleNotFoundError: No module named 'sbir_etl.identity.exact_awards'
```

**To do:** cherry-pick `3df8b073` onto this branch (it's a docstring-only change to the shim this PR creates, so it should apply cleanly once `exact_awards.py` exists), and rebase after #540 is fixed.

---

hollomancer:
## Review comments addressed

- **Frozen-spec docstring**: Added to identity shim confirming no studies/ or specs/ frozen artifact hashes over text referencing phase_iii_negative_controls.identity as the implementation location.
- **Rebase on fixed #540**: Branch now stacks on updated geography-primitive branch.

Ready for re-review once #540 is ready.

PR #542, opened 2026-08-05
Promote source pipelines and remove report script bridges

hollomancer:
## Summary
- move SBIR, SAM.gov, USAspending, and both USPTO downloader implementations behind sbir_etl package APIs
- promote the USAspending availability probe too, removing the hidden script-to-script dependency
- retain all existing download scripts as thin source-checkout-compatible CLI entry points
- make Dagster source-download jobs call package APIs and delete the downloader architecture exception
- replace weekly-report and phase-transition subprocess calls with package APIs
- keep analytical report code explicitly exploratory, render non-citable notices, and stamp report metadata
- reduce the architecture allowlist from four package-to-script bridges to the one remaining tech-area cohort bridge

## Verification
- 69 focused downloader and architecture unit tests passed
- 5 USPTO CLI integration tests passed
- 36 Dagster source-download and weekly-report job tests passed
- 92 reporting, downloader, and boundary tests passed
- all downloader CLIs and the phase-transition CLI help paths exit successfully
- Ruff and mypy pass for promoted/changed modules
- make lint-boundaries, make docs-check, and git diff --check pass

## Stack
Depends on #538, which introduces the subprocess-aware architecture guard and tier-declaration enforcement.

The weekly and phase-transition analyses are deliberately not promoted to pipelines: they calculate or enrich findings. They are package APIs only to remove execution bridges and are labeled exploratory/non-citable.

PR #543, opened 2026-08-05
Replace tech-cohort shell-out with a named package API

hollomancer:
## Summary
- move the tech-area cohort engine behind sbir_etl.reporting.tech_area_cohort
- keep scripts/data/build_tech_area_cohort.py as a thin CLI entry point
- replace the Dagster subprocess bridge with a direct package call
- require every area config to select the named tech-area-cohort-v1 admission profile
- stamp composition, overlap, Dagster metadata, methodology, and policy scaffolds exploratory/non-citable
- use the shared strict YAML reader instead of two direct yaml.safe_load calls
- update the spec to distinguish deterministic mechanics from contestable award admission

## Epistemic decision
This is not a pipeline-tier promotion of the classifier. Method A/B/C award admission and the resulting analytical figures remain exploratory because the rules are contestable. The package API removes the execution bridge and makes the policy version explicit; the CLI remains an entry point.

## Verification
- 37 tech-cohort, Dagster-wrapper, and canonical-award-loader tests passed
- Ruff and mypy pass for the package API and asset wrapper
- CLI help exits successfully
- make lint-boundaries, make docs-check, and git diff --check pass

## Stack
Depends on #539 for the canonical chunk-bounded SBIR award loader and award-grain key.

After #538 and #542 are rebased over this stack, their subprocess-aware guard can delete the final tech-area execution exception. This PR itself removes the subprocess call; its base predates that guard.


hollomancer:
## Merge readiness: not ready — one CI-breaking regression, plus a stack problem

The refactor itself is clean and I'd merge it on the strength of the code. The subprocess bridge is genuinely gone, `build_tech_area_cohort.py` is a 14-line CLI entry point, the epistemic framing (`EPISTEMIC_TIER = "exploratory"`, non-citable stamped into `_cohort_metrics`) is the right call rather than a silent promotion, and `_repo_root()` correctly re-anchors on `config/transition_reports` now that the script it used to look for is gone. Ruff, ruff-format, mypy, `check_architecture_boundaries`, `check_removed_src_references` and `check_identity_boundaries` all pass locally on this head.

### Blocker: `verify_tech_area_figures.py` was left pointing at the gutted module

All four Fast Tests shards are red on this head. Reproduced locally:

```
scripts/data/verify_tech_area_figures.py:47: in <module>
    aggregate_composition = _btac.aggregate_composition
E   AttributeError: module 'build_tech_area_cohort' has no attribute 'aggregate_composition'
```

`verify_tech_area_figures.py:43-47` loads `build_tech_area_cohort.py` by file path with `importlib` and reads `aggregate_composition` off it. That function moved to `sbir_etl/reporting/tech_area_cohort.py:435` along with everything else, so the import raises at module scope and `tests/unit/scripts/test_verify_tech_area_figures.py` fails to collect — which is why every shard fails, not just one.

**To do:** replace the `importlib.util.spec_from_file_location` block with a direct import:

```python
from sbir_etl.reporting.tech_area_cohort import aggregate_composition
```

That also drops the `sys.path.insert` and `importlib` machinery the file only needed to reach a script.

Worth a grep for any other consumer doing the same file-path load of `build_tech_area_cohort.py` — this one was invisible to the architecture guard because it isn't a package-to-script import, it's a runtime file load.

### Stack problem: the base branch no longer merges into `main`

This PR targets `agent/promote-sbir-award-grain` (#539). Since #539 was pushed, #535 landed on `main` and touched the same code #539 rewrites:

```
CONFLICT (content): Merge conflict in sbir_etl/utils/tech_census.py
CONFLICT (content): Merge conflict in tests/unit/utils/test_tech_census.py
```

#539 rewrote `load_award_data_csv` to adapt the canonical materialization; #535's Massachusetts unmanned-systems work changed the same region. #539 has to be rebased before this stack can land, and this branch rebased on top of that.

### Sequencing note

Your PR body is right that the `TRANSITIONAL_SCRIPT_EXECUTIONS` entry can't be deleted here because the base predates the guard. Confirmed: at #542's head that dict holds exactly one entry, `transition_report.py → build_tech_area_cohort.py`, and this PR is what makes it removable. Whichever of #542/#543 lands second should delete it — leaving a stale exception is how the guard rots.

**Verification I ran:** full CI unit selection (`pytest tests/unit/ -m "not slow"`) on this head — 5535 passed, plus the 5 collection errors above. The `test_neo4j_client` and `test_sbir_awards` failures in my first run were parallel-collection artifacts in my worktree; both pass in isolation and are not attributable to this PR.

---

hollomancer:
## Review comments addressed

- **verify_tech_area_figures.py regression**: Replaced importlib-based file-path load of build_tech_area_cohort.py with direct import from sbir_etl.reporting.tech_area_cohort. The gutted module no longer blocks collection.
- **Rebase on fixed #539**: Branch now stacks on the updated promote-sbir-award-grain branch.

Ready for re-review once #539 conflict is resolved and CI runs.

hollomancer:
## Merge readiness at `55dc2f7`: your fix is verified — blocked only by what it inherits from #539

### Fixed and confirmed

`scripts/data/verify_tech_area_figures.py` now imports directly:

```python
from sbir_etl.reporting.tech_area_cohort import aggregate_composition
```

The `importlib.util.spec_from_file_location` block and the `sys.path.insert` it needed are both gone. `tests/unit/scripts/test_verify_tech_area_figures.py` collects and passes, which was what took all four shards down. Merges clean into `main` and into #539. Guards, mypy and ruff all pass; nothing in this PR's own diff is outstanding — the only file it adds beyond #539 is `sbir_etl/reporting/tech_area_cohort.py`.

### Blocker: inherited from the base branch

#539's merge commit `90c170fc` committed 13 untracked local files, and this branch carries all of them — including `infrastructure/cdk/cdk.out/` (CDK synth output, ~1,680 lines) and `specs/nih-commercialization-signal/`. Two CI checks fail on the merged tree as a result:

```
$ uv run python scripts/ci/check_epistemic_tiers.py
specs/nih-commercialization-signal/requirements.md: missing '**Target epistemic tier:** <tier>' declaration

$ uv run python scripts/ci/check_removed_src_references.py
specs/status.md:1: top-level spec is missing from status registry: nih-commercialization-signal
```

and `tests/unit/scripts/test_epistemic_tiers.py::test_current_repository_spec_declarations_are_valid` fails on the same violation. Full run: 5565 passed, that one failure. (The `test_neo4j_client` failure in my run is a parallel-collection artifact in my worktree — 37 passed in isolation.)

**To do:** nothing here. Clean the files off #539 and rebase; both PRs go green together. Details on the #539 thread.

### Reminder for whichever of #542/#543 lands second

`TRANSITIONAL_SCRIPT_EXECUTIONS` on `main` currently holds three entries. #542 removes two (`weekly_awards_report.py`, `phase_transition_archive.py`); this PR makes the third removable:

```python
"packages/sbir-analytics/sbir_analytics/assets/transition_report.py": frozenset(
    {"scripts/data/build_tech_area_cohort.py"}
),
```

The dict's own comment says each entry is removed when its API lands. Whoever merges second should delete the remaining entry rather than leave a stale exception — nothing enforces that it ever goes.

---

PR #565, opened 2026-08-09
Improve portfolio-facing repository guidance

hollomancer:
## Summary

- replace self-disqualifying authorship language with a factual account of project direction, evidence design, and AI-assisted implementation
- promote the runnable synthetic Army procurement-transition example and its committed expected output
- add a compact capability/status table that distinguishes implementation from evidentiary maturity
- replace invalid package-index installation commands with the verified workspace install path

## Why

The repository already contains strong technical and research artifacts, but the front page buries them and asks reviewers to discount the work before seeing the evidence. The package READMEs also advertise packages that are not published to PyPI, including an ML extra that does not exist.

This is a deliberately narrow first portfolio-readiness change. Runtime reliability, CI truthfulness, stale architecture material, and broader packaging metadata are deferred to focused follow-up PRs.

## Verification

- `make install`
- `make install-core`
- `make docs-check`
- `git diff --check`
- generated the documented Army packet and byte-compared it with `examples/army_science_technology_report.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,9 @@ position of any agency.
 The fastest end-to-end example is a deterministic Army procurement-transition
 packet built entirely from committed synthetic data:
 
+Run `make install` from the repository root first; `make install-core` omits the
+`sbir_ml` package used by this example.
+
 ```bash
 uv run python scripts/data/monthly_procurement_transition_report.py \
   --month 2026-06 \
@@ -45,8 +48,8 @@ The repository separates software capability from evidentiary maturity:
 
 | Capability | Current status | Evidence or boundary |
 | --- | --- | --- |
-| Procurement-transition reporting | Runnable synthetic demonstration | [Synthetic example and expected output](examples/README_ARMY_PROCUREMENT_TRANSITION.md) |
-| Award ingestion, entity resolution, and graph loading | Implemented; real-data setup required | [Getting-started guide](docs/getting-started/README.md) and [architecture](docs/architecture/detailed-overview.md) |
+| Procurement-transition reporting | Exploratory; runnable synthetic demonstration | [Synthetic example and expected output](examples/README_ARMY_PROCUREMENT_TRANSITION.md) |
+| Award ingestion, entity resolution, and graph loading | Implemented; real-data setup required | Operational capability, not an evidence claim; see the [getting-started guide](docs/getting-started/README.md) and [architecture](docs/architecture/detailed-overview.md) |
 | Phase III outcome analysis | Reproducible; not validated or approved for citation | [Phase III census study record](studies/phase-iii-census/study.yaml) |
 | Private-capital, M&A, and fiscal analyses | Exploratory and data-dependent | [Research output status index](docs/research/README.md) and the limitations below |
 
@@ -76,8 +79,9 @@ record to other public datasets. A few of the questions it explores:
   economic activity attributable to award spending, using BEA input-output tables
   where available and fallback assumptions when live BEA inputs are unavailable.
 
-The full, sourced inventory (organized by policy area and complexity) is
-in [docs/research-questions.md](docs/research-questions.md).
+The full, sourced inventory in
+[docs/research-questions.md](docs/research-questions.md) is the heart of the
+project: the code and studies exist to investigate and validate those questions.
 
 ## What it actually does
 
@@ -151,7 +155,7 @@ repository, start with these documents in order:
 
 ## Running it
 
-The project supports **Python 3.11 and 3.12** and uses
+The project targets **Python 3.11** and uses
 [`uv`](https://github.com/astral-sh/uv) for
 dependency management. There is intentionally no `requirements.txt` — the
 dependency set is defined by `pyproject.toml` and pinned in `uv.lock`. (If you
@@ -208,6 +212,9 @@ for compatibility boundaries, increment rules, and the release checklist.
 ## License
 
 MIT — see [LICENSE](LICENSE). Copyright (c) 2025 Conrad Hollomon.
+
+Contributions are welcome; see [CONTRIBUTING.md](CONTRIBUTING.md) for the local
+workflow and review expectations.
 
 ## Acknowledgments
 
--- a/README.md
+++ b/README.md
@@ -5,21 +5,50 @@ downstream commercialization signals (federal contracts,
 patents, private financing, and acquisitions) to better
 understand what happens after a small business wins an SBIR award.
 
-## About this project (please read first)
+## My role and use of AI
 
-- I am a federal employee working in the SBIR/STTR domain, not a
-  trained data scientist or ML engineer. I'm hoping to contribute which
-  questions are worth asking, how they map to the policy and
-  academic literature, and what data could plausibly answer them.
-- This is built with substantial help from AI agents. Much of the
-  implementation was generated and iterated with Claude and Codex. I
-  directed the design and validated the analysis, but this code definitely
-  isn't professional-grade engineering. This is a **personal side project,
-  not production software.** PRs welcome!
-- Recommend newcomers read **[docs/research-questions.md](docs/research-questions.md)**, 
-  a structured inventory of the questions this project exists to answer, each
-  tied to the relevant GAO/NASEM/CRS reports and peer-reviewed studies. That
-  document is the heart of the project, and the code is the work to validate it.
+- Defined the research agenda and functional requirements, starting with the
+  policy questions in [docs/research-questions.md](docs/research-questions.md).
+- Selected public data sources and specified the entity-linkage, analytical,
+  and reporting methods used to investigate those questions.
+- Set evidence and validation boundaries, including what the outputs can and
+  cannot support.
+- Used Claude and Codex extensively to implement and iterate on the software,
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -5,21 +5,50 @@ downstream commercialization signals (federal contracts,
...
+packet built entirely from committed synthetic data:
+
+```bash
+uv run python scripts/data/monthly_procurement_transition_report.py \
+  --month 2026-06 \
+  --awards examples/army_science_technology_awards.csv \
+  --candidates examples/army_science_technology_candidates.csv \
+  --opportunities examples/army_science_technology_opportunities.csv \
+  --output-root /tmp/procurement-transition-example
+```
```
hollomancer, line 39: **This is the first runnable command a portfolio reader will copy, and it fails after `make install-core` — which this same README recommends 130 lines below.**

`scripts/data/monthly_procurement_transition_report.py:12` imports `sbir_ml.transition.detection.fusion_scoring` at module level. `sbir_ml` is only reachable through the `stack-dev` extra (`pyproject.toml:78-83`); it is not in the base dependency set (`pyproject.toml:13-29`). I confirmed the gap with a non-mutating `uv sync --dry-run` of exactly what `make install-core` runs — it reports it would uninstall 109 packages including `sbir-ml==0.3.0`, `sbir-graph`, `sbir-analytics`, and `dagster`.

So the reader gets `ModuleNotFoundError: No module named 'sbir_ml'` before argparse even runs. The repo already knows this trap and documents it elsewhere — `docs/getting-started/README.md:105-106`: *"`dagster`, `pytest`, or a first-party package is missing: run `make install`, not `make install-core`."*

That's the worst possible place for it, given the PR's purpose. This section is the marquee "here is the work, go run it" moment, it sits ~130 lines above any install instruction, and the failure lands on someone evaluating the project. One line above the fence fixes it:

> Requires `make install` (the full workspace) — `make install-core` omits `sbir_ml`.

I did verify the example itself is genuinely reproducible, which is the more important claim and it holds: I ran the documented command and byte-compared the output against `examples/army_science_technology_report.md` — **identical**. Deterministic, offline, no API key. The flags in this block match the script's argparse exactly. (Also worth knowing: `uv run`'s default sync is inexact, so this command won't clobber a `make install` environment.)

**On the capability table below**, one gap against its own framing. It's introduced as separating "software capability from evidentiary maturity," and rows 2-4 each carry a maturity statement in house vocabulary — I checked rows 3 and 4 against `studies/phase-iii-census/study.yaml:4` (`evidence_status: reproducible`) and `docs/research/README.md:45-50` and both match exactly, which is good and conservative. But row 1 — the one the README foregrounds — says only "Runnable synthetic demonstration," with no tier. Its code declares `EPISTEMIC_TIER = "exploratory"` with the note "outputs are non-citable" (`sbir_etl/reporting/__init__.py:3-7`), and the driver lives in `scripts/`, exploratory by default.

This is a labeling gap rather than a tier violation — no number is quoted anywhere in the README, and lines 40-42 say "synthetic" three times, so nothing here is citable-looking. But a reader who sees three rows hedged with maturity language and one that isn't will reasonably infer the unhedged one sits outside the tier system, when it's actually at the bottom of it. Adding "exploratory" to that cell closes it.

Minor, same table: row 2's "Evidence or boundary" cell points at the getting-started guide and the architecture overview — both real, both how-to docs, neither an evidence statement or a boundary.

---

hollomancer (commented):
## Claims check out, including the one I most expected not to — one install prerequisite to fix

I verified the substantive claims rather than reading for plausibility:

- **The Army example reproduces byte-for-byte.** I ran the documented command and compared against the committed `examples/army_science_technology_report.md` — identical. Deterministic, offline, no API key needed. The README's flags match the script's argparse exactly, and both `examples/README_ARMY_PROCUREMENT_TRANSITION.md` and the expected output exist on the branch. Promoting this was the right call; a reviewer can confirm the whole evidence trail in one command.
- **The removed `pip install` commands were genuinely invalid, on two counts.** All four package names 404 on PyPI (`sbir-etl`, `sbir-analytics`, `sbir-graph`, `sbir-ml`), and `pip install "sbir-etl[ml]"` named an extra that doesn't exist at all — the real extras are `ucc1-pilot, uspto, uspto-browser, cloud, streaming, phonetic, monitoring, viz, dev, stack-dev`. Both replacement Make targets exist and do what the README says (`Makefile:152-155`, `:157-160`).
- **Tier discipline holds.** The README quotes no numbers, marks no research question answered, and the two rows touching research maturity match `studies/phase-iii-census/study.yaml:4` and `docs/research/README.md:45-50` exactly. For a portfolio-facing rewrite this is the thing most likely to go wrong, and it didn't.
- All 26 relative links across the four changed files resolve, and no file anywhere links to a `README.md#anchor`, so removing the "About this project (please read first)" heading breaks no inbound link.

### The one thing to fix

The "See it work" quickstart needs `make install`, not `make install-core`, and doesn't say so — detail inline. It's a one-line fix but it lands on exactly the reader this PR is written for.

### Second: the PyPI cleanup stops one file short, and now contradicts itself

The new `packages/sbir-graph/README.md:13-14` states the packages "are not published to PyPI." `docs/migrations.md:45` still instructs `pip install sbir-graph`. A reader following the Neo4j migration runbook runs a command that cannot work, after the package's own README told them so.

Same string also appears as a *runtime* error message at `packages/sbir-graph/sbir_graph/loaders/neo4j/__init__.py:92` and `.../client.py:17` — arguably out of scope, but it's the same invalid instruction reaching users through a third channel. (`docs/decisions/ADR-002-*.md` is correctly left alone; the hygiene checker excludes it as a historical record.)

### Smaller

- `packages/sbir-analytics/README.md:12-18` omits "From the repository root," which both sibling READMEs gained in this same PR. There's no `Makefile` under `packages/`, so a reader sitting in that directory gets "No targets specified and no makefile found."
- `README.md:154` says the project "supports Python 3.11 and 3.12." True against `pyproject.toml:8` (`>=3.11,<3.13`), but untested — `ci.yml:21` pins `PYTHON_VERSION: "3.11"` with no 3.12 leg, ruff targets `py311`, mypy `python_version = "3.11"`. Also not covered by any of the four stated claims.
- The north-star framing for `docs/research-questions.md` got demoted. Main called it "the heart of the project, and the code is the work to validate it," matching CLAUDE.md:7; it now survives as a link inside a bullet about authorship and as item 1 of the reading path. Still reachable, but for a repo whose stated organizing principle is that inventory, the emphasis seems worth keeping. "PRs welcome!" also disappeared, and `CONTRIBUTING.md` exists but still isn't linked from the README.

**One claim I could not verify:** `make docs-check`. Running it needs the working tree at this branch, which I avoided. I checked the two things this diff could plausibly break — all relative links resolve, and no new content matches the stale-command patterns in `check_removed_src_references.py:30-51` (`src/`, `src.`, `--cov=src`, `poetry run`, `black`). It very likely passes; I just didn't run it.

---

hollomancer:
Addressed the review in 86c24ff6: the marquee example now requires `make install`, its exploratory tier is explicit, the capability-only row is labeled as such, repository-only install guidance replaces invalid PyPI commands, the supported runtime claim is narrowed to Python 3.11, and the north-star/contributing links are restored. Verification: `make docs-check`, Ruff on the touched Python modules, and `git diff --check` all pass.

README.md
hollomancer: Verified addressed at `86c24ff6` — resolving. Every point, including the ones I marked as nits.

- **Install prerequisite** — `README.md:29-30`, above the fence rather than below it, and it names the specific reason (`make install-core` omits `sbir_ml`) instead of just asserting a requirement.
- **Capability row 1** — now "Exploratory; runnable synthetic demonstration", which matches `sbir_etl/reporting/__init__.py:3-7`'s `EPISTEMIC_TIER = "exploratory"`. All four rows now carry a maturity statement.
- **Row 2's boundary cell** — "Operational capability, not an evidence claim" is a better answer than the one I suggested. I'd flagged that the cell contained two how-to links and no boundary; naming the *absence* of an evidence claim is the honest boundary, rather than inventing one.
- **`docs/migrations.md:41`** — `pip install sbir-graph` → `make install`.
- **`packages/sbir-analytics/README.md:15`** — "From the repository root:" added, matching its two siblings.
- **Python version** — "targets **Python 3.11**" instead of claiming 3.12 support that no CI leg exercises.
- **North-star framing** — restored, and stronger than main's wording: `research-questions.md` is "the heart of the project: the code and studies exist to investigate and validate those questions."
- **`CONTRIBUTING.md`** — now linked.

You also went past what I raised and fixed the three runtime `ImportError` messages in `sbir_graph/loaders/` that told users to `pip install sbir-graph`. I'd called those arguably out of scope; they were the same invalid instruction reaching users through a third channel, so I think fixing them was right.

I re-ran the check that mattered: `git grep "pip install sbir-"` on the branch now returns hits only in `docs/decisions/ADR-002-etl-library-extraction.md`, which is correct to leave — the hygiene checker excludes it as a historical decision record, and rewriting an ADR to match present reality would defeat its purpose.

One non-blocking note: the branch is now 12 commits behind main, though it still merges clean (`git merge-tree` exits 0). Worth a rebase before merge so CI runs against what would actually land.

---

PR #566, opened 2026-08-09
Make CI coverage reporting truthful

hollomancer:
## Summary

- remove the Codecov configuration and upload step that did not produce a working coverage gate
- preserve the full-suite terminal report and retain `coverage.xml` as a downloadable GitHub Actions artifact
- add `sbir-ml` to the blocking MyPy gate and its incremental-cache key
- remove inaccurate and duplicated workflow commentary

## Why

The latest main-branch run generated coverage successfully, but the Codecov upload was rejected because it had no valid token. The action defaulted to `fail_ci_if_error: false`, so the workflow stayed green. Separately, `codecov.yml` waited for five uploads even though CI produced only one, and pull requests produced none.

That combination advertised a coverage regression gate that did not exist. This PR removes the misleading integration while preserving an inspectable coverage artifact. A future Codecov integration can be added when authentication and PR report generation are configured and tested together.

The ML package already passes the repository's MyPy configuration, so it is now included in the existing blocking type-check without suppressions or production-code changes. Analytics typing remains a separate bounded follow-up.

## Verification

- `.venv/bin/python -m mypy sbir_etl packages/sbir-graph/sbir_graph packages/sbir-ml/sbir_ml`
- `actionlint -oneline -shellcheck= -pyflakes=`
- `.venv/bin/python scripts/ci/check_removed_src_references.py`
- `git diff --check`
- repository search confirms no remaining Codecov configuration or action reference


.github/workflows/ci.yml
```diff
@@ -228,13 +220,14 @@ jobs:
...
-        uses: codecov/codecov-action@v7
+        uses: actions/upload-artifact@v7
         with:
-          files: ./coverage.xml
-          flags: full
-          name: full-coverage
+          name: coverage-report
+          path: coverage.xml
+          if-no-files-found: error
+          retention-days: 14
```
hollomancer, line 230: I traced this end to end and the artifact is genuinely reachable — not an orphaned path, which is the usual way this refactor goes wrong:

- generation is untouched at `:207-215` (`--cov-report=xml --cov-report=term`);
- there is no `[tool.coverage]` in `pyproject.toml` and no `.coveragerc`/`setup.cfg`/`tox.ini`, so the xml lands at the default workspace-root `coverage.xml`, which is what `path:` names;
- `coverage.xml` is gitignored (`.gitignore:52`), confirming it's a real runtime artifact rather than a committed file;
- the step sits after `Stop Neo4j`, which is `if: always()`, so it's still reached when tests fail — which is when you most want the report;
- `actions/upload-artifact@v7` does exist upstream (confirmed against the tag list).

**One process point, and it's the reason I'd not merge blind.** This step is never exercised by this PR's own CI. `test-full` is `if: github.event_name != 'pull_request'` (`:170`), and PR #566's checks show `Full Tests` = **skipped**. So every green check here was produced without the new step ever running, and the first evidence of a typo would be a red `Full Tests` on `main` after merge.

That's cheap to close: `workflow_dispatch` also satisfies the `!= 'pull_request'` condition, so dispatching one run from this branch exercises the step for real before it lands. Worth doing given the whole point of the PR is that CI should not claim things it hasn't demonstrated.

Minor, on line 229: `if-no-files-found: error` can't fire — the step `if:` on line 224 already requires `hashFiles('coverage.xml') != ''`, so the step is skipped in exactly the case the guard exists to catch. Two guards for one condition. Harmless, but it reads as protection that isn't doing anything; either drop `if-no-files-found` or drop the `hashFiles` clause and let the upload fail loudly when coverage genuinely didn't generate. I'd keep the latter, since a silent skip is closer to the failure mode this PR is fixing.

---

hollomancer (commented):
## Review: all four claims verified true — the gap is that the PR's own thesis isn't applied to the docs

I checked the premises against live CI logs rather than taking the PR body's word for them, and everything holds:

**The Codecov removal is complete and the diagnosis was right.** `git grep -ril codecov` over the whole branch tree returns **zero** files (main had exactly two). No orphaned README badge — there never was one. The stated cause is confirmed from main's own run log (`31330858727`, job `93288801542`, step "Upload coverage"): `Token length: 0` … `Upload queued for processing failed: {"message":"Token required - not valid tokenless upload"}` — and the step still closed `outcome=success`. The `after_n_builds: 5` observation checks out too: only `test-full` passes `--cov`, and the log confirms "Found 1 coverage files to report." So the gate really was advertised-but-absent.

**The MyPy addition is stronger than the PR claims.** `uv run python -m mypy sbir_etl packages/sbir-graph/sbir_graph packages/sbir-ml/sbir_ml` → `Success: no issues found in 306 source files`. Not vacuous: 265 files without `sbir-ml`, so **41 files are genuinely newly checked**, with no `ignore_errors`, no per-module override, and no exclusion (`[tool.mypy] exclude` covers only `scripts/`, `tests/`, `examples/`). I probed the "passes trivially on a weak config" concern directly: `sbir-ml` still passes under `--check-untyped-defs`, while `sbir_etl` *fails* that same flag with 5 errors — so it isn't sneaking through on `check_untyped_defs = false`. The cache key hashes exactly the three checked roots plus `pyproject.toml`, so there's no path/scope mismatch.

**No required check was silently dropped.** All eight job names are byte-identical to main; only two *step* names changed, and steps aren't check contexts. This PR's 16 check runs contain no `codecov/*` context at all, which independently confirms the Codecov gate never posted anything.

**Merge-safety non-finding, worth stating so nobody re-derives it:** the branch's merge base is `dc3370f6`, and main has since added `check_file_sizes.py` to the guards step — so a two-dot `git diff origin/main..branch` *looks* like it deletes that guard. It doesn't. A `git merge-tree` produces a clean merge where `check_file_sizes.py` survives, MyPy carries `sbir-ml`, and the artifact upload lands intact.

### What I'd fix

**The one real gap is that this PR removes inaccurate commentary from the workflow but leaves the repo's designated source of truth newly inaccurate about the very thing it changed.** `docs/development/pre-commit-ci-consistency.md` — which self-declares as "the **single source of truth**" — still states the CI MyPy scope as `sbir_etl` + `sbir-graph` at `:29` and `:256`. CI now checks `sbir-ml` too. `.pre-commit-config.yaml:68` has the same drift (`MyPy (type check): sbir_etl/ [CI also checks sbir-graph]`).

Concrete failure: a contributor adds untyped code under `packages/sbir-ml/sbir_ml`, consults the consistency doc, sees ML is out of CI scope, and gets a surprise red `Lint, Types, and Guards`. That lands squarely inside this PR's own claim #4, and it's a two-line fix in files the PR is already conceptually about.

I've left one inline comment on the artifact upload step — the main point there is that **this PR's CI never exercises the step it adds** (`Full Tests` is skipped on PRs), which is worth closing with a `workflow_dispatch` run before merge given the subject matter.

### Smaller things

- The deleted block after `:77` bundled one genuinely false sentence ("the only workflow in the repository" — `versioning.yml` exists and its check runs on this very PR) with an accurate, load-bearing rationale: the 2026-08-02 empty-`env:` incident that silently disabled a workflow. The actionlint step now has no statement of why it exists, which invites a future cleanup to delete it and reintroduce silent-CI-outage. "Surgical Changes" argues for correcting the one wrong clause rather than dropping the incident record.
- The PR body's "retains `coverage.xml` as a downloadable artifact" is true only for push-to-main and `workflow_dispatch`; pull requests produce no coverage at all. Not a regression — main behaved the same — but the framing reads as PR-level availability.
- Repo-side Codecov config is gone, but the Codecov **GitHub App** may still be installed on the repo. Inert now that nothing uploads, though "remove the misleading integration" isn't quite finished until it's uninstalled. I can't read app installations from here, so flagging rather than asserting.

No blocking defect. Nice PR — the diagnosis was precise and the fix is smaller than the problem, which is the right shape.

---

PR #567, opened 2026-08-09
Remove phantom ML vectorizer API

hollomancer:
## Summary

- delete the zero-byte `ml.features.vectorizers` module
- remove its swallowed, unused `create_feature_matrix_builder` import
- remove documentation for five vectorizer classes that were never implemented
- document the actual extension point: callers may inject their own numeric feature builder

## Why

The patent-classifier guide presented a structured-feature vectorizer API as implemented, but the referenced module was empty. The classifier imported a nonexistent factory inside a broad exception handler, assigned `None`, and never used the name, so the defect remained invisible at runtime and in CI.

No current caller, test, or milestone requires the proposed built-in classes. Removing the phantom API is smaller and more honest than inventing an unvalidated parallel feature path. The existing `feature_matrix_builder` injection point remains unchanged.

## Verification

- 21 focused patent-classifier, ML pipeline-smoke, and multi-source-vectorizer tests passed
- `ruff check packages/sbir-ml/sbir_ml`
- `ruff format --check packages/sbir-ml/sbir_ml`
- `mypy packages/sbir-ml/sbir_ml`
- repository documentation/hygiene guard
- repository search confirms no remaining reference to the deleted module, factory, or nonexistent classes

### DIFF
--- a/docs/ml/cet-classifier.md
+++ b/docs/ml/cet-classifier.md
@@ -1,6 +1,7 @@
 # CET Patent Classifier — Training and Feature Flow
 
-This document covers the **patent** CET classifier: feature extraction, vectorizers, training/inference flow, evaluation, and Dagster assets.
+This document covers the **patent** CET classifier: feature extraction,
+training and inference, evaluation, and Dagster assets.
 
 For the **award** CET classifier (`ApplicabilityModel`) — including architecture, hyperparameters, Neo4j schema, quality checks, and integration scenarios — see [cet-integration.md](cet-integration.md).
 
@@ -16,14 +17,6 @@ For the **award** CET classifier (`ApplicabilityModel`) — including architectu
     - `extract_ipc_cpc`, `guess_assignee_type`, `bag_of_keywords_features`
     - `extract_features(record) -> PatentFeatureVector` with `as_dict()`
     - Small default keyword map and stopword set for deterministic testing.
-
-- Vectorizers (convert features → numeric matrices)
-  - `packages/sbir-ml/sbir_ml/ml/features/vectorizers.py`
-    - `KeywordVectorizer`: count and presence features by keyword group
-    - `TokenCounterVectorizer`: token counts with/without stopwords
-    - `AssigneeTypeVectorizer`: one-hot encoding of assignee type
-    - `IPCPresenceVectorizer`: IPC/CPC section presence flags (A–H)
-    - `FeatureMatrixBuilder`: combine vectorizers to produce a single 2D matrix
 
 - Classifier wrapper and extractor
   - `packages/sbir-ml/sbir_ml/ml/models/patent_classifier.py`
@@ -65,26 +58,6 @@ This module avoids heavy NLP dependencies and only uses Python and `re`.
 
 ---
 
-## Vectorizers
-
-When training real sklearn-like pipelines, you will typically need numeric feature matrices. `packages/sbir-ml/sbir_ml/ml/features/vectorizers.py` provides a minimal set:
-
-- `KeywordVectorizer(keywords_map)`:
-  - For each group in `keywords_map`, creates `<group>__count` and `<group>__presence`.
-- `TokenCounterVectorizer()`:
-  - Two columns: `n_tokens`, `n_tokens_no_stopwords`.
-- `AssigneeTypeVectorizer()`:
-  - One-hot across: `academic`, `company`, `government`, `individual`, `unknown`.
-- `IPCPresenceVectorizer()`:
-  - Binary presence for IPC and CPC across sections `A..H`.
-- `FeatureMatrixBuilder([vec1, vec2, ...])`:
-  - Horizontal concatenation of individual matrices
-  - `get_feature_names()` returns concatenated feature names across all vectorizers
-
-These are designed to be import-safe and deterministic.
-
----
-
 ## Classifier
 
 `packages/sbir-ml/sbir_ml/ml/models/patent_classifier.py`
@@ -99,9 +72,13 @@ These are designed to be import-safe and deterministic.
 
 Notes:
 
-- Training can proceed using normalized title-only text, or using extracted features transformed into numeric matrices via `FeatureMatrixBuilder`.
+- Training can proceed using normalized title-only text. Callers may instead
+  inject a numeric feature builder through `feature_matrix_builder`; this
+  repository does not currently provide a built-in structured-feature matrix
+  builder.
 - For lightweight pipelines (like `DummyPipeline` used in tests), text inputs are sufficient.
-- For production pipelines (sklearn), pass a matrix via `feature_matrix_builder` in `train_from_dataframe`.
+- For sklearn pipelines that require numeric input, pass a matrix builder through
+  `feature_matrix_builder` in `train_from_dataframe`.
 
 ---
 
@@ -186,9 +163,10 @@ Store evaluation metrics in your artifact metadata by merging them into `config`
 
 ## CI & Import Safety
 
-- Feature extraction and vectorizers avoid heavy dependencies.
+- Feature extraction avoids heavy dependencies.
 - Many components perform defensive imports and fallback to minimal behavior to keep tests green in lean environments.
-- ML unit tests live under `tests/unit/ml/` and are included in the fast `ml-unit-tests` CI job.
+- ML unit tests live under `tests/unit/ml/` and run in the pull-request unit
+  shards and the post-merge full suite.
 
 ---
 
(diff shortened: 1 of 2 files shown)


docs/ml/cet-classifier.md
```diff
@@ -205,7 +183,9 @@ Store evaluation metrics in your artifact metadata by merging them into `config`
 ## Practical Tips
 
 - Start simple: use `use_feature_extraction=True` to feed normalized text to text-based pipelines (or `DummyPipeline` for CI).
-- Move to numeric features for sklearn: construct a `FeatureMatrixBuilder` and pass it to `train_from_dataframe` as `feature_matrix_builder`.
+- To use numeric features, supply a caller-owned builder with `fit_transform`
+  and, optionally, `get_feature_names`, then pass it to `train_from_dataframe`
+  as `feature_matrix_builder`.
```
hollomancer, line 188: **This is the right call, but the replacement contract omits the one thing a caller needs to implement it correctly — and getting it wrong fails silently.**

Deleting the phantom API instead of inventing five classes to justify the docs is clearly correct, and I verified the removal is clean: the deleted module really was the zero-byte blob (`e69de29b`), the `try` block you removed contained exactly one import so no sibling import lost its guard, and `create_feature_matrix_builder` has zero remaining readers anywhere in the repo.

The gap is in what replaces it. "Supply a caller-owned builder with `fit_transform`" doesn't say what `fit_transform` receives, and it is not what an sklearn-shaped guess would assume:

```python
# patent_classifier.py:270
X_matrix = feature_matrix_builder.fit_transform(feature_vectors)
```

`feature_vectors` is `list[PatentFeatureVector]`, built at `:261-264` — not text, not a DataFrame, not an array. A reader following this paragraph writes a text vectorizer, and it does not raise: `:273-275` catches bare `Exception`, sets `X_matrix = None`, and training falls back to `texts` at `:307` with no log line. The run succeeds, the numeric features are silently absent, and `feature_names` is never persisted. That's the same defect class this PR exists to remove — a documented signature that doesn't match the code — just relocated from "class that doesn't exist" to "input type that isn't stated."

Two more things about this paragraph, both cheap to fix here:

**The extension point is unreachable from the entry point this document recommends.** Line 99 and line 202 both point readers at `train_patent_classifier`, and that is also what production uses (`assets/cet/training.py:118`). Its signature (`patent_training.py:129-141`) has no `feature_matrix_builder` parameter and doesn't forward one to `train_from_dataframe` at `:209-217`. The only way to reach what this paragraph documents is to construct `PatentCETClassifier` by hand. Either forward the kwarg or say so.

**It now actively recommends a path the repo's own steering doc flags as critical.** `docs/steering/ml-methodology-review.md:218` records `ml-007, critical` against exactly this call: the builder is fit over all rows before any train/test split. Before this PR that hazard sat behind an API nobody could use; promoting it to a recommended workflow without the caveat makes it reachable. Worth a sentence. (Incidentally your diff fixes that citation's line numbers — it was off by 5 against main and now lands exactly on `:268-275`.)

Suggested minimal fix: state the input type (`list[PatentFeatureVector]`), note that a builder that raises is swallowed and training silently falls back to text, and add the leakage caveat with a pointer to ml-007.

---

docs/ml/cet-classifier.md
```diff
@@ -205,7 +183,9 @@ Store evaluation metrics in your artifact metadata by merging them into `config`
 ## Practical Tips
 
 - Start simple: use `use_feature_extraction=True` to feed normalized text to text-based pipelines (or `DummyPipeline` for CI).
```
hollomancer, line 185: **There is a second phantom API in this file, and it survived the rewrite — this line documents it.**

`use_feature_extraction` is declared at `patent_classifier.py:219` as `use_feature_extraction: bool = False` and is **never read**. One grep hit in the entire 541-line module: the parameter declaration itself. `train_from_dataframe` runs `PatentFeatureExtractor` unconditionally at `:261-264` regardless of the flag's value.

It propagates. `patent_training.py:215` forwards it straight into that dead parameter, and the production Dagster asset at `assets/cet/training.py:125` passes `use_feature_extraction=True` — believing it selects a behavior. It selects nothing. The only function that actually honors the flag, `build_training_inputs` (`patent_training.py:81-126`), has zero callers repo-wide.

This line also states the default backwards: it presents `use_feature_extraction=True` as the "start simple" setting, while the real default is `False`.

I'm raising it here rather than as separate work because it's the same defect this PR is fixing — an API documented as meaningful that the code ignores — in the same file, one line above the paragraph you rewrote for accuracy. Leaving it means the file still misdescribes the code after a PR whose entire thesis is that it shouldn't.

Fixing it is a judgment call I'd rather you make than assume: either delete the dead parameter and this line (smallest change, matches the PR's existing logic), or wire it up so `build_training_inputs` is reachable and the production asset's `True` means something. If you'd rather keep this PR narrow, that's defensible — but then it's worth a one-line note here so the next reader isn't misled, because a live caller is currently passing `True` into it.

---

hollomancer (commented):
## Review: the deletion is correct and complete; the replacement documentation is where the work is left

I went looking specifically for the two ways this kind of change breaks, and **neither is present**:

- **No `NameError`.** On main, `create_feature_matrix_builder` appears at exactly two lines — the import at `patent_classifier.py:74` and the `= None` fallback at `:76`. Zero other references in the module, zero repo-wide. The PR removed both together. `py_compile` succeeds on the branch file and a `ruff --select F` pass reports no `F821`. No star-imports or `getattr` access anywhere under `packages/`, `tests/`, `scripts/` that could reach it dynamically.
- **No newly-unguarded sibling import.** The deleted `try` block imported exactly one name. The other defensive block at `:60-71` (importing `PatentFeatureVector` / `extract_features`) is untouched context in the hunk, still wrapped in its own `try/except`, with the module-level `None` placeholders at `:54-55` preserved.

I also reproduced the repo-search claim independently rather than taking it: all five class names, the factory, and both `features/vectorizers` and `features.vectorizers` return **zero** case-insensitive hits on the branch, including docs, specs, notebooks, and tests. `ml/features/__init__.py` and `ml/__init__.py` both carry `__all__ = []` with no imports, so there's no stale export. The four inbound doc links target the file rather than the removed `#vectorizers` anchor, so nothing dangles. `pyproject.toml` uses whole-tree `packages = ["sbir_ml"]`, so there's no packaging manifest to update.

Branch is 1 ahead / 8 behind main; none of those 8 commits touch these three files, so it merges clean.

### What I'd fix before merging

Two inline comments, both on `docs/ml/cet-classifier.md`, and they're the same defect class the PR exists to eliminate:

1. **`use_feature_extraction` is a second phantom API still documented here** — dead parameter at `patent_classifier.py:219`, never read, and the production Dagster asset passes `True` into it believing it does something.
2. **The new extension-point contract omits the input type**, and a caller who guesses sklearn semantics fails *silently* — the exception is swallowed at `:273-275` and training falls back to text with no log line. Also unreachable from the entry point this same document recommends, and it now promotes a path the repo's own steering doc flags as `ml-007, critical` without the caveat.

Neither breaks anything at runtime, so I wouldn't block on them — but they're cheap here and expensive later, and #2 in particular converts "documentation nobody could act on" into "documentation someone can act on and get quietly wrong."

### Minor, not worth blocking

- Line 190 says feature names are stored "when a matrix builder is used"; the code at `:330-332` stores them only `if feature_names:` — i.e. only when the builder *also* has `get_feature_names`, it returns something truthy, and `fit_transform` didn't raise. Slight overstatement, now sitting two lines below new text that correctly calls `get_feature_names` optional.
- Line 212 ("Add richer vectorizers and calibration") implies existing vectorizers in a file that no longer documents any. Same for `config/cet/patent_keywords.yaml:6,155`, which still describe the keyword map as feeding "the patent feature extractor ... and/or vectorizers." These don't name the module or classes, so your claim survives as worded — it's residual drift, not a false statement. (Unrelated and pre-existing: `patent_keywords.yaml:11` shows `from sbir_etl.ml.features.patent_features import ...` — wrong package, should be `sbir_ml`.)
- The `ml-unit-tests` → PR-shards CI correction at lines 168-169 is **accurate** — that job genuinely doesn't exist, and `ci.yml:138-174` does shard `tests/unit/ -m "not slow"` on `pull_request`. Flagging only because it isn't in the PR body's list of changes, and CLAUDE.md asks that every changed line trace to the request.

No lint or type regression: `E501` is globally ignored (`pyproject.toml:122`), `warn_unused_ignores = false` (`:159`) makes the dropped `# type: ignore[attr-defined]` neutral, `check_removed_src_references.py` only scans for `src/` paths, and `check_epistemic_tiers.py` only validates `specs/`.

---

PR #568, opened 2026-08-09
Align environment configuration contract

hollomancer:
## Summary

- document the direct Neo4j variables used by host and Compose workflows, including the database selector
- distinguish the Compose `ENVIRONMENT` selector from the canonical loader environment variable
- omit nested `SBIR_ETL__NEO4J__USERNAME/PASSWORD` compatibility overrides because they shadow the direct credentials and `env_file` does not expand their old ${...} values
- retain the legacy Bolt URL only where the host E2E scripts still consume it
- forward every documented direct Neo4j setting through both Compose environments
- add contract tests for the public template, Compose mapping, and override precedence

## Why

The public environment template mixed three different contracts: direct loader settings, nested compatibility overrides, and script-only E2E names. In particular, the nested username/password settings are accepted by the loader and take precedence over the direct values, while their old ${NEO4J_USER} syntax remained literal when read through Compose `env_file`.

This PR makes those boundaries explicit without changing the loader's compatibility behavior. It intentionally does not enable global rejection of extra configuration keys because environment profiles still contain component-owned top-level sections.

## Verification

- `uv run pytest -q tests/unit/config/test_loader.py tests/unit/config/test_env_example.py tests/unit/test_server_runtime_contract.py tests/unit/test_server_ops.py` — 99 passed
- `uv run ruff check tests/unit/config/test_loader.py tests/unit/config/test_env_example.py`
- development and server Compose configurations parse with their example environment files
- `git diff --check`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,9 +5,11 @@
 # -----------------------
 # Neo4j configuration
 # -----------------------
-# Neo4j authentication (username/password). Use a strong password for real deployments.
+# These four direct variables are read by the application configuration loader.
+# Use a strong password for real deployments.
 NEO4J_USER=neo4j
 NEO4J_PASSWORD=test
+NEO4J_DATABASE=neo4j
 
 # Host-local processes connect through the published Docker port. Compose
 # overrides this with the internal service hostname for container processes.
@@ -22,8 +24,13 @@ NEO4J_server_memory_heap_max__size=1G
 # -----------------------
 # Application / runtime
 # -----------------------
-# Environment mode: dev | test | prod
+# Compose profile selector: dev | test | prod. Compose forwards this value to
+# the application under its canonical selector.
 ENVIRONMENT=dev
+
+# Direct host processes use this selector. Long aliases such as development
+# and production are also accepted.
+SBIR_ETL__PIPELINE__ENVIRONMENT=dev
 
 # Keep Python output unbuffered for container logs
 PYTHONUNBUFFERED=1
@@ -34,23 +41,8 @@ DOCKER_IMAGE_NAME=sbir-etl:latest
 # -----------------------
 # SBIR ETL application-specific variables
 # -----------------------
-# Canonical Neo4j Bolt URL used by host-local application processes
-SBIR_ETL__NEO4J__BOLT_URL=bolt://localhost:7687
-
-# Individual connection pieces (application may prefer either URL or these pieces)
-SBIR_ETL__NEO4J__HOST=localhost
-SBIR_ETL__NEO4J__PORT=7687
-SBIR_ETL__NEO4J__USERNAME=${NEO4J_USER}
-SBIR_ETL__NEO4J__PASSWORD=${NEO4J_PASSWORD}
-
 # DuckDB defaults (path inside container or host bind mount)
 SBIR_ETL__DUCKDB__DATABASE_PATH=data/processed/sbir.duckdb
-
-# Data directories (when using bind mounts in dev)
-SBIR_ETL__DATA_DIR=/app/data
-SBIR_ETL__CONFIG_DIR=/app/config
-SBIR_ETL__LOG_DIR=/app/logs
-SBIR_ETL__METRICS_DIR=/app/metrics
 
 # -----------------------
 # Optional service toggles / tuning
--- a/.env.example
+++ b/.env.example
@@ -5,7 +5,11 @@
 # -----------------------
 # Neo4j configuration
 # -----------------------
-# These four direct variables are read by the application configuration loader.
+# These direct variables are the documented application contract. Nested
+# SBIR_ETL__NEO4J__USERNAME/PASSWORD overrides are intentionally omitted: the
+# loader accepts them for compatibility, but they shadow these direct values,
+# and references such as ${NEO4J_USER} are not expanded inside Compose env_file
+# files. Unset stale nested overrides when migrating an existing environment.
 # Use a strong password for real deployments.
 NEO4J_USER=neo4j
 NEO4J_PASSWORD=test
@@ -80,6 +84,10 @@ CPU_LIMIT=2.0
 # E2E Neo4j ports (to avoid conflicts with dev instance)
 NEO4J_E2E_HTTP_PORT=7475
 NEO4J_E2E_BOLT_PORT=7688
+
+# Direct host E2E scripts retain these compatibility names.
+NEO4J_USERNAME=neo4j
+SBIR_ETL__NEO4J__BOLT_URL=bolt://localhost:7687
 
 # -----------------------
 # Notes
--- a/tests/unit/config/test_env_example.py
+++ b/tests/unit/config/test_env_example.py
@@ -1,0 +1,35 @@
+from pathlib import Path
+
+
+ENV_EXAMPLE = Path(__file__).parents[3] / ".env.example"
+
+
+def _documented_keys() -> set[str]:
+    return {
+        line.partition("=")[0]
+        for raw_line in ENV_EXAMPLE.read_text(encoding="utf-8").splitlines()
+        if (line := raw_line.strip()) and not line.startswith("#") and "=" in line
+    }
+
+
+def test_env_example_documents_the_supported_runtime_selectors() -> None:
+    keys = _documented_keys()
+
+    assert {
+        "ENVIRONMENT",
+        "SBIR_ETL__PIPELINE__ENVIRONMENT",
+        "NEO4J_URI",
+        "NEO4J_USER",
+        "NEO4J_PASSWORD",
+        "NEO4J_DATABASE",
+    } <= keys
+
+    assert {
+        "SBIR_ETL__NEO4J__BOLT_URL",
+        "SBIR_ETL__NEO4J__HOST",
+        "SBIR_ETL__NEO4J__PORT",
+        "SBIR_ETL__DATA_DIR",
+        "SBIR_ETL__CONFIG_DIR",
+        "SBIR_ETL__LOG_DIR",
+        "SBIR_ETL__METRICS_DIR",
+    }.isdisjoint(keys)
--- a/tests/unit/config/test_env_example.py
+++ b/tests/unit/config/test_env_example.py
@@ -2,6 +2,7 @@ from pathlib import Path
 
 
 ENV_EXAMPLE = Path(__file__).parents[3] / ".env.example"
+COMPOSE_FILE = Path(__file__).parents[3] / "docker-compose.yml"
 
 
 def _documented_keys() -> set[str]:
@@ -22,14 +23,19 @@ def test_env_example_documents_the_supported_runtime_selectors() -> None:
         "NEO4J_USER",
         "NEO4J_PASSWORD",
         "NEO4J_DATABASE",
+        "NEO4J_USERNAME",
+        "SBIR_ETL__NEO4J__BOLT_URL",
     } <= keys
 
-    assert {
-        "SBIR_ETL__NEO4J__BOLT_URL",
-        "SBIR_ETL__NEO4J__HOST",
-        "SBIR_ETL__NEO4J__PORT",
-        "SBIR_ETL__DATA_DIR",
-        "SBIR_ETL__CONFIG_DIR",
-        "SBIR_ETL__LOG_DIR",
-        "SBIR_ETL__METRICS_DIR",
-    }.isdisjoint(keys)
+    nested_neo4j_keys = {key for key in keys if key.startswith("SBIR_ETL__NEO4J__")}
+    assert nested_neo4j_keys == {"SBIR_ETL__NEO4J__BOLT_URL"}
+
+
+def test_compose_forwards_every_documented_direct_neo4j_setting() -> None:
+    compose = COMPOSE_FILE.read_text(encoding="utf-8")
+    common_environment = compose.split("x-common-environment:", 1)[1].split(
+        "x-dev-environment:", 1
+    )[0]
+
+    for key in ("NEO4J_URI", "NEO4J_USER", "NEO4J_PASSWORD", "NEO4J_DATABASE"):
+        assert f"  {key}:" in common_environment
--- a/tests/unit/config/test_loader.py
+++ b/tests/unit/config/test_loader.py
@@ -584,7 +584,7 @@ class TestGetConfig:
                 "NEO4J_PASSWORD": "custom-password",
                 "NEO4J_DATABASE": "analytics",
             }
-            with patch.dict(os.environ, neo4j_environment):
+            with patch.dict(os.environ, neo4j_environment, clear=True):
                 reload_config()  # Clear cache
                 config = get_config(environment="development", config_dir=config_dir)
 
@@ -592,6 +592,22 @@ class TestGetConfig:
             assert config.neo4j.username == "custom-user"
             assert config.neo4j.password == "custom-password"
             assert config.neo4j.database == "analytics"
+
+    def test_nested_neo4j_environment_overrides_direct_compatibility_values(self):
+        """Nested compatibility values retain their documented precedence."""
+        environment = {
+            "NEO4J_USER": "direct-user",
+            "NEO4J_PASSWORD": "direct-password",
+            "SBIR_ETL__NEO4J__USERNAME": "nested-user",
+            "SBIR_ETL__NEO4J__PASSWORD": "nested-password",
+        }
+
+        with patch.dict(os.environ, environment, clear=True):
+            reload_config()
+            config = get_config(environment="development")
+
+        assert config.neo4j.username == "nested-user"
+        assert config.neo4j.password == "nested-password"
 
     def test_get_config_handles_legacy_loading_neo4j(self):
         """Test get_config handles legacy 'loading.neo4j' structure."""
--- a/tests/unit/config/test_loader.py
+++ b/tests/unit/config/test_loader.py
@@ -567,8 +567,8 @@ class TestGetConfig:
             # Production uses prod-neo4j by default (can be overridden via env vars)
             assert config.neo4j.uri == "bolt://prod-neo4j:7687"
 
-    def test_get_config_respects_explicit_neo4j_uri_env(self):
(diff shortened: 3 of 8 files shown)


.env.example
```diff
@@ -5,9 +5,11 @@
 # -----------------------
 # Neo4j configuration
 # -----------------------
-# Neo4j authentication (username/password). Use a strong password for real deployments.
+# These four direct variables are read by the application configuration loader.
+# Use a strong password for real deployments.
 NEO4J_USER=neo4j
 NEO4J_PASSWORD=test
+NEO4J_DATABASE=neo4j
```
hollomancer: **Two of the variables this removes aren't ignored — they're honored, and they override the ones being promoted here.**

`SBIR_ETL__NEO4J__USERNAME` and `SBIR_ETL__NEO4J__PASSWORD` map to real `Neo4jConfig` fields through `_apply_env_overrides` (`sbir_etl/config/loader.py:93-120`), which splits on `__` and writes `config_dict["neo4j"]["username"]`. I traced the ordering to be sure of the direction:

```python
neo.setdefault("username", os.getenv("NEO4J_USER", "neo4j"))     # loader.py:228
...
if apply_env_overrides_flag:
    config_dict = _apply_env_overrides(config_dict)               # loader.py:258-259
```

The `setdefault` from `NEO4J_USER` runs *first*; the nested override runs after and wins. So the nested form doesn't just work — it takes precedence over the direct variable this file now presents as canonical. The repo asserts this itself at `tests/integration/test_configuration_environments.py:150-167`, `docker-compose.yml:248-249` sets both, and `scripts/docker/entrypoint.sh:22-23` documents both.

The removal is still the right call, and for a better reason than the one given. On main these lines read `SBIR_ETL__NEO4J__USERNAME=${NEO4J_USER}` — and Compose's `env_file:` does **not** interpolate. A verbatim `cp .env.example .env` injected the literal string `${NEO4J_USER}` into the `app` container (`docker-compose.yml:238-239`), where `_apply_env_overrides` would have set `neo4j.username = "${NEO4J_USER}"`. The `environment:` block masked it, but that was a genuine footgun and deleting it is correct.

What I'd change is the rationale, in the PR body and in the header comment above: these are **redundant duplicates that shadow the direct variable**, not variables the loader ignores. The distinction matters operationally — someone with a stale `SBIR_ETL__NEO4J__PASSWORD` export who reads "ignored or incorrect" will not think to unset it, and it will silently beat the `NEO4J_PASSWORD` they just set. `scripts/docker/entrypoint.sh:18-23` also still says "Environment variables used (documented in `.env.example`)" and then lists all four now-removed names.

**Separately, `NEO4J_DATABASE` on line 12 reintroduces the exact defect this PR exists to remove.** It's documented here under "read by the application configuration loader" — and it is, at `loader.py:230` — but neither compose file forwards it into any container. `docker-compose.yml` `x-common-environment` (lines 20-33) and `docker-compose.server.yml` `x-server-environment` (lines 29-88) have no `NEO4J_DATABASE` entry. `docker-compose.server.yml:54-56` states the governing rule out loud: "`--env-file` only supplies Compose interpolation, so every variable the code reads must be mapped explicitly here or it never reaches the container."

So an operator sets `NEO4J_DATABASE=sbir`, runs `make docker-up`, and dagster-webserver, dagster-daemon, and etl-runner all silently keep using `neo4j`. Only the `ci` `app` service picks it up, because it alone has `env_file: - .env`. That's a template advertising an inert knob — the precise thing the PR is fixing elsewhere. One line in each of the two compose anchors closes it:

```yaml
NEO4J_DATABASE: ${NEO4J_DATABASE:-neo4j}
```

Two smaller things on this block: line 8 says "These four direct variables" directly above three assignments (the fourth, `NEO4J_URI`, is at line 16 under its own comment), and `NEO4J_USERNAME` — distinct from `NEO4J_USER`, hard-required by `scripts/e2e_health_check.py:48` and `scripts/run_e2e_tests.py:69`, set by `docker-compose.yml:245` and `ci.yml:23` — is documented in neither template.

---

tests/unit/config/test_loader.py
```diff
@@ -578,11 +578,20 @@ def test_get_config_respects_explicit_neo4j_uri_env(self):
...
+                "NEO4J_DATABASE": "analytics",
+            }
+            with patch.dict(os.environ, neo4j_environment):
                 reload_config()  # Clear cache
                 config = get_config(environment="development", config_dir=config_dir)
 
             assert config.neo4j.uri == "bolt://custom:7687"
+            assert config.neo4j.username == "custom-user"
+            assert config.neo4j.password == "custom-password"
+            assert config.neo4j.database == "analytics"
```
hollomancer: **This test fails inside the repo's own container test run.** `patch.dict` without `clear=True` merges into the ambient environment rather than replacing it, so an inherited `SBIR_ETL__NEO4J__USERNAME` survives — and per the precedence I traced on `.env.example`, it then beats the `NEO4J_USER` this test is asserting.

Reproduced on this branch:

```
$ env -u SBIR_ETL__NEO4J__USERNAME -u SBIR_ETL__NEO4J__PASSWORD pytest ...::test_get_config_respects_direct_neo4j_environment
1 passed

$ SBIR_ETL__NEO4J__USERNAME=neo4j SBIR_ETL__NEO4J__PASSWORD=password pytest ...::test_get_config_respects_direct_neo4j_environment
>       assert config.neo4j.username == "custom-user"
E       AssertionError: assert 'neo4j' == 'custom-user'
1 failed
```

Those two variables aren't hypothetical — they're exactly what the `ci`-profile `app` service sets:

```yaml
- SBIR_ETL__NEO4J__USERNAME=${NEO4J_USER}
- SBIR_ETL__NEO4J__PASSWORD=${NEO4J_PASSWORD}
...
command: ... pytest -m fast -q;
```

and `test_loader.py:26` is `pytestmark = pytest.mark.fast`. So `make docker-test` / `make docker-e2e` go red on this test. It also breaks plain `make test-unit` for any developer whose `.env` predates this PR and who uses direnv (`.envrc:17 dotenv_if_exists`) — which is precisely the population the PR is aimed at. GitHub Actions is unaffected, since those jobs don't set the nested variables, which is why CI here is green.

Fix is one argument:

```python
with patch.dict(os.environ, neo4j_environment, clear=True):
```

Worth noting the failure is *informative* rather than spurious: the test is implicitly asserting "the direct variables win," and the environment where it fails is one where they genuinely don't. Adding `clear=True` makes the test pin the intended contract; it would be worth a second test that sets both forms and asserts the nested one wins, since that's the real precedence rule and nothing currently pins it at unit level.

---

tests/unit/config/test_env_example.py
```diff
@@ -0,0 +1,35 @@
...
+    } <= keys
+
+    assert {
+        "SBIR_ETL__NEO4J__BOLT_URL",
+        "SBIR_ETL__NEO4J__HOST",
+        "SBIR_ETL__NEO4J__PORT",
+        "SBIR_ETL__DATA_DIR",
+        "SBIR_ETL__CONFIG_DIR",
+        "SBIR_ETL__LOG_DIR",
+        "SBIR_ETL__METRICS_DIR",
```
hollomancer: This test can't enforce the property the PR is actually about. It checks required keys with a subset relation and then blacklists seven literal names — so it pins "these must be present" and "these seven must be absent," but not "no phantom variable may be present," which is the contract.

Demonstrated: appending `SBIR_ETL__NEO4J__USERNAME=neo4j`, `SBIR_ETL__TOTALLY__PHANTOM=1`, and `SBIR_ETL__NEO4J__URL=typo` to this branch's `.env.example` leaves the test **passing**. Note the first of those is one of the two variables this PR removes most consequentially, and the blacklist doesn't cover it.

The repo already has the shape that would work, at `tests/unit/test_server_runtime_contract.py:87-107` — it cross-checks `.env.server.example` against the compose file rather than against a hardcoded list. An equivalent `.env.example` ↔ `docker-compose.yml` mapping check would have caught the `NEO4J_DATABASE` gap I flagged on the template, instead of leaving it to review.

**One entry in the blacklist is actively wrong.** Line 28 forbids `SBIR_ETL__NEO4J__BOLT_URL` from `.env.example` permanently, but two scripts hard-require it and abort without it:

- `scripts/run_e2e_tests.py:71` and `scripts/e2e_health_check.py:50` both fail `check_environment()` if it's unset
- `docs/testing/e2e-testing.md:49` instructs developers to export it

So a developer regenerating `.env` from the new template and running the documented host command gets `❌ Missing required environment variables: SBIR_ETL__NEO4J__BOLT_URL`. The container path is fine (`docker-compose.yml:247` sets it explicitly) — it's the documented host workflow that breaks. Either keep the variable in the template, or drop it from the blacklist and update those two scripts and the doc; the current combination encodes "never document it" while the tooling requires it.

Minor: this file has no `pytestmark`, breaking the convention in its two closest neighbours (`test_server_env_defaults.py:19`, `test_server_runtime_contract.py:10`, both `[pytest.mark.fast, pytest.mark.unit]`). Today that omission is load-bearing in an unfortunate way — `.env.example` is neither `COPY`'d into the image nor bind-mounted, unlike `.env.server.example` (`docker-compose.yml:300`), so adding the `fast` marker without also adding the mount would make the container run die on `FileNotFoundError: /app/.env.example`. Worth doing both together.

---

hollomancer (commented):
## Right direction, but the central premise is partly false and one new test breaks the container suite

The goal here is good and most of it lands. Three of the five claims hold exactly, no live deployment breaks, and `.env.server` is untouched so the server host is safe. Two things need fixing before this comes out of draft.

**1. The premise is wrong for two of the nine removed variables.** `SBIR_ETL__NEO4J__USERNAME` and `SBIR_ETL__NEO4J__PASSWORD` are not ignored — they map to real `Neo4jConfig` fields and, because `_apply_env_overrides` runs *after* the `NEO4J_USER`/`NEO4J_PASSWORD` defaults, they **override** the direct variables this PR promotes as canonical. Detail on the `.env.example` comment. The removal is still correct, but for a different reason (they were `${NEO4J_USER}` literals that Compose's `env_file:` never interpolates), and the stated reason will mislead an operator with a stale export.

**2. `test_get_config_respects_direct_neo4j_environment` fails inside the repo's own container test run.** `patch.dict` without `clear=True` lets the ambient nested variables through, and those are exactly what the `ci`-profile `app` service sets before running `pytest -m fast`. I reproduced both directions — passes clean, fails with `AssertionError: assert 'neo4j' == 'custom-user'` under the container's environment. GitHub Actions doesn't set those variables, which is why CI here is green and this wouldn't have been caught by the checks on this PR.

**3. `NEO4J_DATABASE` reintroduces the defect the PR exists to remove.** It's newly documented as loader-read (true — `loader.py:230`), but neither compose file forwards it, so setting it changes nothing for dagster-webserver, dagster-daemon, or etl-runner. `docker-compose.server.yml:54-56` states the rule this violates in so many words.

### Every removed variable, and whether a consumer survives

| Removed from `.env.example` | Ignored by the loader? | Live consumer still on the branch? |
|---|---|---|
| `SBIR_ETL__NEO4J__BOLT_URL` | Yes (no field, dropped by `extra="ignore"`) | **Yes** — `e2e_health_check.py:23,50`; `run_e2e_tests.py:71`; `docker-compose.yml:247`; `docs/testing/e2e-testing.md:49` |
| `SBIR_ETL__NEO4J__HOST` | Yes | **Yes** — `docker-compose.server.yml:36`; `docker-compose.yml:29`; four `scripts/docker/*.sh` |
| `SBIR_ETL__NEO4J__PORT` | Yes | **Yes** — `docker-compose.server.yml:37`; `docker-compose.yml:30`; same four scripts |
| `SBIR_ETL__NEO4J__USERNAME` | **No — honored, and wins** | **Yes** — `docker-compose.yml:248`; `test_configuration_environments.py:154,167`; `entrypoint.sh:22` |
| `SBIR_ETL__NEO4J__PASSWORD` | **No — honored, and wins** | **Yes** — `docker-compose.yml:249`; `entrypoint.sh:23` |
| `SBIR_ETL__DATA_DIR` / `CONFIG_DIR` / `LOG_DIR` / `METRICS_DIR` | Partly (land as `extra="allow"` attrs on `PipelineConfig`; nothing reads them) | No |

**No removal breaks a live deployment** — every compose/shell consumer supplies its own `${…:-default}`, and the container `environment:` blocks set values explicitly. The one user-visible regression is the *documented host* e2e workflow, via the `BOLT_URL` blacklist entry (detail on the test file).

### Also worth folding in

- **`.env.server.example` is untouched**, so the "aligned contract" stops at the dev template. It documents neither `NEO4J_DATABASE` (read by `loader.py:230` and by `scripts/server/backup.sh:16`), nor `NEO4J_URI` (interpolated at `docker-compose.server.yml:38`), nor `SBIR_ETL__PIPELINE__ENVIRONMENT`. Given CLAUDE.md says the live host runs on `.env.server`, that's the template where drift costs most. Related and pre-existing: `backup.sh:16` reads `NEO4J_DATABASE` from the process environment and never calls `load_env_key`, so it can't pick it up from `.env.server` even once documented.
- `scripts/docker/entrypoint.sh:18-23` still says "Environment variables used (documented in `.env.example`)" and lists all four now-removed `SBIR_ETL__NEO4J__*` names.
- `.env.example` ships both `ENVIRONMENT=dev` and `SBIR_ETL__PIPELINE__ENVIRONMENT=dev` with nothing enforcing agreement. The new comments explain the distinction well, but a host developer who edits only `ENVIRONMENT=prod` — which every previous version of this file told them to do — silently keeps loading `dev.yaml`, since `loader.py:36,57` reads only the latter.

### Credit where it's due

Removing the `SBIR_ETL__NEO4J__USERNAME=${NEO4J_USER}` lines fixes a real latent footgun that nobody had noticed: `env_file:` does not interpolate, so a verbatim `cp .env.example .env` was injecting the literal string `${NEO4J_USER}` into the `app` container, where `_apply_env_overrides` would have set `neo4j.username = "${NEO4J_USER}"`. The `environment:` block masked it. Right call regardless of the stated rationale.

**Minimal fix set:** add `clear=True` at `test_loader.py:587`; add `NEO4J_DATABASE: ${NEO4J_DATABASE:-neo4j}` to both compose anchors; restate the USERNAME/PASSWORD rationale as "redundant duplicate that shadows the direct variable" and update `entrypoint.sh:18-23`; resolve the `BOLT_URL` contradiction one way or the other; mirror `NEO4J_URI`/`NEO4J_DATABASE` into `.env.server.example`.

---

hollomancer:
Addressed all three review threads in a2bdacb5. The template now explains that nested username/password overrides are supported but shadow direct values; both Compose anchors forward `NEO4J_DATABASE`; the CI app no longer injects the redundant nested credentials; host E2E compatibility variables remain documented; `.env.server.example` and backup loading cover the database selector; and tests pin both mapping and precedence. Verification: 99 focused tests passed, Ruff passed, and both Compose configurations parse.

hollomancer:
## Verified fixed at `a2bdacb5` — all findings addressed, and the contract test is now the enforcing shape

I re-checked each finding against the new head rather than taking the resolutions at face value.

**The container test failure is gone, confirmed in the environment that caused it.** I ran the config suite twice from a clean worktree at this head:

```
$ env -u SBIR_ETL__NEO4J__USERNAME -u SBIR_ETL__NEO4J__PASSWORD pytest tests/unit/config/
162 passed

$ SBIR_ETL__NEO4J__USERNAME=neo4j SBIR_ETL__NEO4J__PASSWORD=password pytest tests/unit/config/test_loader.py
62 passed
```

That second invocation is exactly what the `ci`-profile `app` service sets before `pytest -m fast`, and it was the failing case. `make docker-test` is unblocked.

**The precedence rule is now pinned, which is the better half of that fix.** `test_nested_neo4j_environment_overrides_direct_compatibility_values` sets both forms and asserts the nested one wins. That was the real contract and nothing pinned it before — the old test only asserted the direct variables work when nothing shadows them.

**`NEO4J_DATABASE` now reaches containers** — `docker-compose.yml:34` and `docker-compose.server.yml:41`, both as `${NEO4J_DATABASE:-neo4j}` on the shared anchors, so all services pick it up rather than just the `ci` `app`.

**The template test became an enforcing check.** This is the change I'd single out:

```python
nested_neo4j_keys = {key for key in keys if key.startswith("SBIR_ETL__NEO4J__")}
assert nested_neo4j_keys == {"SBIR_ETL__NEO4J__BOLT_URL"}
```

Exact equality on the nested namespace, so a future stray `SBIR_ETL__NEO4J__*` fails rather than slipping past a hardcoded blacklist. And `test_compose_forwards_every_documented_direct_neo4j_setting` cross-checks `.env.example` against the compose anchor — that's the mapping check I said would have caught the `NEO4J_DATABASE` gap without needing review to find it. Both are modeled on `test_server_runtime_contract.py`, which was the right thing to copy.

**The `BOLT_URL` contradiction resolved the correct direction.** It's kept in `.env.example:90` and now *required* by the test, so `scripts/run_e2e_tests.py` and `scripts/e2e_health_check.py` still work from the documented host workflow. `NEO4J_USERNAME` is now required too, closing the gap where it was set by compose and CI but documented nowhere.

**The rationale is fixed, and more precisely than I suggested.** `.env.example:8-12` now says the nested overrides "shadow these direct values" and adds the operational instruction I was actually after — "Unset stale nested overrides when migrating an existing environment" — plus the `${NEO4J_USER}`-isn't-expanded-in-`env_file` explanation. `scripts/docker/entrypoint.sh:18-25` no longer claims the removed variables are documented in `.env.example`, and correctly notes that Compose supplies `HOST`/`PORT` internally for dependency probes.

**`.env.server.example`** gained `NEO4J_URI` and `NEO4J_DATABASE` (lines 45-46) and the `ENVIRONMENT` forwarding note, and `scripts/server/backup.sh` was touched as well — so the server template no longer lags the dev one.

One deliberate non-change worth confirming as correct: `test_env_example.py` still has no `pytestmark`. That's the right call — adding `fast` without also bind-mounting `.env.example` into the image would make the container run die on `FileNotFoundError`, as I noted. Leaving it host-only avoids that.

Nothing outstanding. The branch merges clean into current main.

---

PR #531, opened 2026-08-04
Specify the Phase III survivor substantiation audit

hollomancer:
## What changed

- adds a prospective Phase 0 design for hand-reviewing a probability sample of frozen R15 census survivors
- defines the public-evidence hierarchy, element-level findings, deterministic adjudication categories, outcome masking, immutable review records, and weighted reporting contract
- adds explicit approval questions, a pending decision record, and an append-only amendment-log template

## Why

The census, negative controls, and placebo are reproducible, but they do not establish technical lineage to a particular prior SBIR/STTR effort. This specification defines the next bounded validation step without reopening or tuning the frozen census criteria.

The design deliberately calls the work a **survivor substantiation audit**. A survivor-only public-evidence sample can estimate substantiation and credible-support yield; it cannot identify recall, sensitivity, statutory prevalence, or total Phase III undercount.

## Process gate

This PR is specification-only and remains draft. No pilot or confirmatory sample may be drawn, no packet may be opened, and no numeric gate may be selected until the repository owner resolves and freezes every question in `questions.md`.

## Validation

- scope-guard review completed
- independent quality sweep completed
- Markdown whitespace and 100-column checks passed
- `git diff --check` passed
- no scoring imports, weights, thresholds, ML, census criteria changes, sampling, labels, or outcomes are included

PR #569, opened 2026-08-09
Address remaining epistemic-tier review gaps

hollomancer:
## What changed

- ignore tier declarations inside fenced Markdown code blocks so quoted examples cannot create false duplicate declarations
- document the rule used to assign specification target tiers, including the lifecycle gate between deterministic preparation and contestable analysis
- document the subprocess guard's literal/static limits and synchronize the architecture guide with the now-empty migration allowlists

## Why

This is the follow-up to #538, which merged before its review thread was resolved. Later merges already fixed the subprocess false-positive bug, local-command resolution, capitalized-tier diagnostics, the gated/deferred tier assignments, and the #531 compatibility file. This PR closes the remaining review gaps without replaying those already-landed changes.

The fenced-block issue came from applying the tier-declaration regular expression to raw Markdown, so a declaration quoted in a code example counted as a real header.

## Impact

No production runtime behavior changes. Spec validation becomes more precise, and future implementers get an explicit rule for the maintenance cost implied by a target-tier declaration.

## Validation

- `pytest -q tests/unit/scripts/test_architecture_boundaries.py tests/unit/scripts/test_epistemic_tiers.py` — 18 passed
- Ruff check and format check on changed Python files
- `scripts/ci/check_architecture_boundaries.py`
- `scripts/ci/check_epistemic_tiers.py`
- `scripts/ci/check_tier_boundaries.py`
- `scripts/ci/check_removed_src_references.py`
- `git diff --check`

### DIFF
--- a/scripts/ci/check_epistemic_tiers.py
+++ b/scripts/ci/check_epistemic_tiers.py
@@ -13,6 +13,7 @@ TIER_DECLARATION = re.compile(
     r"^\*\*Target epistemic tier:\*\*\s*`?([A-Za-z]+)`?\s*$",
     flags=re.MULTILINE,
 )
+FENCE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
 EXCLUDED_SPEC_DIRECTORIES = frozenset({"archive"})
 
 
@@ -25,6 +26,25 @@ class TierDeclarationViolation:
 
     def format(self) -> str:
         return f"{self.path}: {self.message}"
+
+
+def _outside_fenced_code(markdown: str) -> str:
+    """Return Markdown content outside fenced code blocks."""
+
+    retained: list[str] = []
+    marker: tuple[str, int] | None = None
+    for line in markdown.splitlines():
+        match = FENCE.match(line)
+        if marker is None:
+            if match:
+                fence = match.group(1)
+                marker = fence[0], len(fence)
+            else:
+                retained.append(line)
+            continue
+        if match and match.group(1)[0] == marker[0] and len(match.group(1)) >= marker[1]:
+            marker = None
+    return "\n".join(retained)
 
 
 def validate_spec_directory(
@@ -44,7 +64,8 @@ def validate_spec_directory(
             )
         ]
 
-    declarations = TIER_DECLARATION.findall(requirements.read_text(encoding="utf-8"))
+    markdown = _outside_fenced_code(requirements.read_text(encoding="utf-8"))
+    declarations = TIER_DECLARATION.findall(markdown)
     if not declarations:
         return [
             TierDeclarationViolation(
--- a/scripts/ci/check_epistemic_tiers.py
+++ b/scripts/ci/check_epistemic_tiers.py
@@ -13,7 +13,7 @@ TIER_DECLARATION = re.compile(
     r"^\*\*Target epistemic tier:\*\*\s*`?([A-Za-z]+)`?\s*$",
     flags=re.MULTILINE,
 )
-FENCE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
+FENCE = re.compile(r"^[ \t]*(?P<fence>`{3,}|~{3,})(?P<suffix>.*)$")
 EXCLUDED_SPEC_DIRECTORIES = frozenset({"archive"})
 
 
@@ -29,21 +29,37 @@ class TierDeclarationViolation:
 
 
 def _outside_fenced_code(markdown: str) -> str:
-    """Return Markdown content outside fenced code blocks."""
+    """Return Markdown content outside complete fenced code blocks.
+
+    Unterminated candidate fences are retained. A malformed example must not
+    hide a real declaration later in the specification.
+    """
 
     retained: list[str] = []
+    pending: list[str] = []
     marker: tuple[str, int] | None = None
     for line in markdown.splitlines():
         match = FENCE.match(line)
         if marker is None:
-            if match:
-                fence = match.group(1)
+            if match and not (
+                match.group("fence").startswith("`") and "`" in match.group("suffix")
+            ):
+                fence = match.group("fence")
                 marker = fence[0], len(fence)
+                pending = [line]
             else:
                 retained.append(line)
             continue
-        if match and match.group(1)[0] == marker[0] and len(match.group(1)) >= marker[1]:
+        pending.append(line)
+        if (
+            match
+            and match.group("fence")[0] == marker[0]
+            and len(match.group("fence")) >= marker[1]
+            and not match.group("suffix").strip()
+        ):
             marker = None
+            pending = []
+    retained.extend(pending)
     return "\n".join(retained)
 
 
--- a/tests/unit/scripts/test_epistemic_tiers.py
+++ b/tests/unit/scripts/test_epistemic_tiers.py
@@ -86,3 +86,18 @@ def test_capitalized_tier_is_reported_as_invalid_not_missing(tmp_path: Path) -> 
 
     assert len(violations) == 1
     assert "invalid target tier 'Evidence'" in violations[0].message
+
+
+def test_declaration_inside_fenced_code_is_ignored(tmp_path: Path) -> None:
+    spec = tmp_path / "specs" / "example"
+    _write(
+        tmp_path,
+        "specs/example/requirements.md",
+        "# Example\n\n"
+        "**Target epistemic tier:** pipelines\n\n"
+        "```markdown\n"
+        "**Target epistemic tier:** evidence\n"
+        "```\n",
+    )
+
+    assert not tiers.validate_spec_directory(spec, repository_root=tmp_path)
--- a/tests/unit/scripts/test_epistemic_tiers.py
+++ b/tests/unit/scripts/test_epistemic_tiers.py
@@ -101,3 +101,89 @@ def test_declaration_inside_fenced_code_is_ignored(tmp_path: Path) -> None:
     )
 
     assert not tiers.validate_spec_directory(spec, repository_root=tmp_path)
+
+
+def test_unterminated_fence_does_not_swallow_a_real_declaration(tmp_path: Path) -> None:
+    spec = tmp_path / "specs" / "example"
+    _write(
+        tmp_path,
+        "specs/example/requirements.md",
+        "# Example\n\n```markdown\nunterminated example\n\n"
+        "**Target epistemic tier:** pipelines\n",
+    )
+
+    assert not tiers.validate_spec_directory(spec, repository_root=tmp_path)
+
+
+def test_unterminated_fence_still_reports_a_genuine_duplicate(tmp_path: Path) -> None:
+    spec = tmp_path / "specs" / "example"
+    _write(
+        tmp_path,
+        "specs/example/requirements.md",
+        "**Target epistemic tier:** pipelines\n\n```markdown\nunterminated example\n\n"
+        "**Target epistemic tier:** evidence\n",
+    )
+
+    violations = tiers.validate_spec_directory(spec, repository_root=tmp_path)
+
+    assert len(violations) == 1
+    assert "multiple target tier" in violations[0].message
+
+
+def test_indented_fence_inside_list_hides_example_declaration(tmp_path: Path) -> None:
+    spec = tmp_path / "specs" / "example"
+    _write(
+        tmp_path,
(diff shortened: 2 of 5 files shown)


scripts/ci/check_epistemic_tiers.py
```diff
@@ -27,6 +28,25 @@ def format(self) -> str:
...
+        if marker is None:
+            if match:
+                fence = match.group(1)
+                marker = fence[0], len(fence)
+            else:
+                retained.append(line)
+            continue
+        if match and match.group(1)[0] == marker[0] and len(match.group(1)) >= marker[1]:
+            marker = None
+    return "\n".join(retained)
```
hollomancer, line 63: **A fence that is never closed silently discards the rest of the file — which means a stray backtick can turn a genuine duplicate declaration into a clean pass.**

There is no end-of-input handling here. `marker` is set on an opener at line 41 and cleared only by a matching closer at line 45-46. If the loop reaches EOF with `marker` still set, every line consumed since the opener was never appended to `retained` and is simply gone from what the caller scans.

I reproduced this directly against both parsers rather than reasoning about it. Input: a spec with two *real* tier declarations, with an unclosed fence between them.

```
# Requirements

**Target epistemic tier:** pipelines

Example of a bad spec:

```
(someone forgot the closing fence)

**Target epistemic tier:** evidence
```

Result:

```
raw file contains declarations: ['pipelines', 'evidence']
main   parser sees: ['pipelines', 'evidence'] ->  2 declaration(s)
branch parser sees: ['pipelines']             ->  1 declaration(s)

main verdict  : DUPLICATE (correctly rejected)
branch verdict: CLEAN
```

`main` catches the duplicate. This branch passes it. That is the guard getting weaker, not stronger, and duplicate-declaration detection is one of the three things it exists to do.

The same missing EOF case produces three *false* rejections too — a file that visibly has a declaration is reported as missing one — whenever an unclosed fence appears before the real declaration, when a line starts with an inline code span (` ```markdown``` is inline code ` — CommonMark doesn't treat that as a fence opener because a backtick info string may not contain backticks, but `FENCE` at line 16 matches it), or when a bare `~~~` is left in the file as a rule or an editing leftover. That is exactly the experience `test_epistemic_tiers.py:70-88` was written to prevent — its own docstring says the old behavior "sent the author looking for a line that was already there." This reintroduces it through a different door.

**The fix is small.** Buffer what an open fence consumes, and flush it back if the fence never closes:

```python
pending.append(line)
if match and match.group(1)[0] == marker[0] and len(match.group(1)) >= marker[1]:
    marker = None
    pending = []
retained.extend(pending)   # an unclosed fence never opened a block
```

That cures all four unterminated-fence cases and changes nothing about the cases you already handle correctly — and you do handle more than the tests suggest: tilde fences, 4+-backtick fences, tab indentation, and 3-space indentation all behave correctly on this branch. None of that is pinned by a test, so it's untested-correct rather than protected.

Two narrower gaps remain, both **also present on main**, so they're incomplete-fix rather than regression — but the PR's headline claim is that quoted examples can't create false duplicates, and in these two shapes they still can:

- **Fences indented 4+ spaces are not recognized** (line 16 allows only `[ \t]{0,3}`). A code block inside a list item is conventionally indented four spaces, so a declaration quoted there leaks through and reports a false duplicate.
- **A closing fence carrying an info string wrongly closes the block** (line 45 checks only the fence character and length). CommonMark forbids an info string on a closer, so in a nested ` ```markdown ` / ` ```python ` example the ` ```python ` line is treated as the closer and the declaration leaks.

**Mitigating context, and the reason I'd call this fix-before-merge rather than urgent:** no `specs/*/requirements.md` in the repo currently contains a fenced block at all, so there is zero behavior difference on the real tree today. I confirmed that — the branch parser and main parser produce identical results across every non-archived spec, and `scan_specs()` returns `[]` on both. Every defect above is latent. But `epistemic-tiers.md:30` is unusually direct about the stakes here: "If a tier's contract is not enforced, the tier is decoration." A guard that a stray backtick can silence is worth getting right while it's cheap.

---

tests/unit/scripts/test_epistemic_tiers.py
```diff
@@ -86,3 +86,18 @@ def test_capitalized_tier_is_reported_as_invalid_not_missing(tmp_path: Path) ->
...
+        tmp_path,
+        "specs/example/requirements.md",
+        "# Example\n\n"
+        "**Target epistemic tier:** pipelines\n\n"
+        "```markdown\n"
+        "**Target epistemic tier:** evidence\n"
+        "```\n",
+    )
+
+    assert not tiers.validate_spec_directory(spec, repository_root=tmp_path)
```
hollomancer, line 103: This test covers the single easiest input the new parser can receive: column-0, properly closed, three backticks, with a `markdown` info string. It passes, and it would still pass with every defect I described on `check_epistemic_tiers.py` present.

Specifically uncovered: unterminated fences (the regression), fences indented 4+ spaces, closing fences carrying an info string, tilde fences, 4+-backtick fences, tab-indented fences, and inline code spans at line start.

Four of those the branch already handles correctly — tilde, 4-backtick, tab-indented, and 3-space-indented fences all behave right. That's worth pinning while you're here, because nothing currently stops a future simplification of `_outside_fenced_code` from silently dropping them.

I wrote five adversarial tests against the branch parser and all five fail:

```
FAILED test_unterminated_fence_does_not_swallow_a_real_declaration
FAILED test_unterminated_fence_still_reports_a_genuine_duplicate
FAILED test_fence_indented_four_spaces_inside_a_list_item_is_a_code_block
FAILED test_closing_fence_with_info_string_does_not_close
FAILED test_inline_code_span_at_line_start_is_not_a_fence
```

All six of the PR's own tests pass. The suite isn't wrong, it's just aimed at the happy path — and this is a guard whose whole value is behavior on malformed input.

---

hollomancer (commented):
## Fix before merge: the parser makes the guard weaker on the inputs it was written to handle

The rebase landed cleanly — `git merge-tree --write-tree origin/main HEAD` exits 0, so the earlier `docs/steering/structure.md` conflict is gone, and the "allowlists are now empty" rewrite that used to be in this diff is correctly no longer duplicated (it arrived on main via #558). The scope claim holds exactly: five files, all docs / CI guards / tests, zero lines under `sbir_etl/`, `packages/`, `studies/`, or `config/`.

**But `_outside_fenced_code` has no end-of-input handling, and the consequence is a guard that a stray backtick can silence.** I reproduced it against both parsers rather than reasoning about it:

```
raw file contains declarations: ['pipelines', 'evidence']
main   parser sees: ['pipelines', 'evidence'] -> DUPLICATE (correctly rejected)
branch parser sees: ['pipelines']             -> CLEAN
```

A spec with two genuine tier declarations separated by an unclosed fence is rejected by `main` and **passes on this branch**. Three further cases turn a clean file into a false "missing declaration." Details and a ~5-line fix are in the inline comment; I verified the fix cures all four cases and changes none of the cases you already handle.

I'd hold this rather than land it, because of what this specific guard is for. `epistemic-tiers.md:30` — a file this PR edits — says "If a tier's contract is not enforced, the tier is decoration." A change that makes the enforcement bypassable is worth fixing before it lands, and the fix is mechanical.

**Latent, not live.** No `specs/*/requirements.md` in the repo contains a fenced block today, so the branch parser and main parser produce identical results across every non-archived spec and `scan_specs()` returns `[]` on both. Nothing is broken right now. That's why this is fix-before-merge rather than urgent.

### The rest of the PR is accurate

- **Spec-target-tier rule** — TRUE. `docs/steering/epistemic-tiers.md:163-186`. The lifecycle gate is explicit at `:172-175` ("A gated or deferred spec may use this target to permit data preparation, but it does not authorize any contestable analysis in that spec; that phase requires an explicit retier"), consistent with the ladder at `:147-152`, and the `specs/status.md` link resolves.
- **Subprocess guard limits** — TRUE, and the claim is precise rather than hand-wavy. `structure.md:115-118` says the guard follows "simple local assignments," and `_literal_assignments` / `_path_literal_parts` (`check_architecture_boundaries.py:106-149`) do exactly that, including dropping re-assigned names rather than guessing at them.
- **Empty allowlists** — TRUE. Both are genuinely empty (`check_architecture_boundaries.py:35` and `:42`).
- **No production runtime change** — TRUE. Though worth stating precisely: CI *gate* behavior does change, and per the above it changes for the worse on inputs the tests don't cover.

### Two small things while you're in here

- `docs/steering/epistemic-tiers.md:114-116` still reads "See [structure.md](structure.md#transitional-script-dependencies) for **the current bridge**" — singular, asserting one exists — while both allowlists are empty and `structure.md:108-109` now says "All transitional bridges are retired." This PR edits that file and claims to synchronize the docs with the empty allowlists; this is the one sentence in it that still says otherwise. One-line fix.
- Minor mismatch worth a glance: `structure.md:109` attributes the retired bridges to "report and source-download" (reads as two), while the unchanged code comment at `check_architecture_boundaries.py:37-41` says "Package code also reached **three** scripts... All three are retired." Cosmetic, but one of them is wrong.

Non-findings I checked and cleared: `ruff check` passes with the repo config on the new file (an `--isolated` import-order complaint reproduces on main's version, so it's pre-existing), no line exceeds 100 characters, and the new `../../specs/status.md` link resolves.

---

hollomancer:
Addressed the guard review in fc6a2618. Unterminated candidate fences are retained so they cannot hide declarations; closers must be bare; inline backtick spans are not treated as openers; indented/list, tilde, long-backtick, tab, malformed-closer, and EOF cases are covered. The bridge documentation now states that no bridges are active. Verification: all 13 guard tests, the repository tier scan, Ruff, and `git diff --check` pass.

hollomancer:
## Verified fixed at `fc6a2618` — and the fix goes further than the finding

I re-ran the same 12-case matrix against three parsers (main, the old head, this head) rather than reading the new code and calling it good:

```
CASE                                  main       old head   NEW head   want
----------------------------------------------------------------------------------
A backtick fence + info string        DUPLICATE  CLEAN      CLEAN      CLEAN      ok
B tilde fence                         DUPLICATE  CLEAN      CLEAN      CLEAN      ok
C four-backtick fence                 DUPLICATE  CLEAN      CLEAN      CLEAN      ok
D tab-indented fence                  DUPLICATE  CLEAN      CLEAN      CLEAN      ok
E 3-space-indented fence              DUPLICATE  CLEAN      CLEAN      CLEAN      ok
F 4-space fence in list item          DUPLICATE  DUPLICATE  CLEAN      CLEAN      ok
G closing fence w/ info string        DUPLICATE  DUPLICATE  CLEAN      CLEAN      ok
H inline code span at line start      CLEAN      MISSING    CLEAN      CLEAN      ok
I unterminated fence BEFORE decl      CLEAN      MISSING    CLEAN      CLEAN      ok
J unterminated between 2 REAL decls   DUPLICATE  CLEAN      DUPLICATE  DUPLICATE  ok
K lone ~~~ then real decl             CLEAN      MISSING    CLEAN      CLEAN      ok
L no fences (control)                 CLEAN      CLEAN      CLEAN      CLEAN      ok
```

**Case J is the one that mattered** — two genuine declarations separated by an unclosed fence. It now reports `DUPLICATE` again, matching main. The guard is no longer silenceable by a stray backtick.

**F and G are the part I didn't ask for.** Both were wrong on `main` too — I'd classified them as incomplete-fix rather than regression and said so. They're now correct, so this branch leaves the guard strictly stronger than what it started from, not merely restored. A 4-space-indented fence inside a list item and a closing fence carrying an info string both hide their example declarations properly.

The `pending`/flush approach is what I sketched, and the inline-code-span guard (`match.group("fence").startswith("`") and "`" in match.group("suffix")`) is a cleaner fix for case H than anything I proposed — it encodes the actual CommonMark rule that a backtick info string may not contain backticks, rather than special-casing.

**Tests:** all six cases I named as uncovered now have one, plus a variants test:

```
test_unterminated_fence_does_not_swallow_a_real_declaration
test_unterminated_fence_still_reports_a_genuine_duplicate
test_indented_fence_inside_list_hides_example_declaration
test_closing_fence_with_info_string_does_not_close
test_inline_code_span_at_line_start_is_not_a_fence
test_supported_fence_variants_hide_example_declarations
```

13 passed, and `check_epistemic_tiers.py` still passes against the real spec tree.

`docs/steering/epistemic-tiers.md:114-116` — the "see structure.md for **the current bridge**" sentence — is also corrected.

Nothing outstanding from my review. The branch merges clean into current main.

---

PR #570, opened 2026-08-09
Consolidate YAML configuration loading

hollomancer:
## What changed

- routes the seven production `yaml.safe_load` call sites through `read_yaml_mapping`
- adds an explicit `allow_empty=True` policy for callers that intentionally accept empty documents
- adds an AST boundary guard that rejects new direct production `yaml.safe_load` calls
- wires the guard into CI and `make lint-boundaries`
- updates the epistemic-tier steering readout to reflect the enforced contract

## Why

Configuration parsing had seven local implementations with subtly different failure and empty-file behavior. That was invisible divergence outside the existing configuration primitive.

## Impact

Production YAML mapping reads now share one error and type-validation boundary. Existing permissive callers retain their behavior through an explicit option, and future bypasses fail CI.

## Validation

- 99 focused unit tests passed
- baseline-registry and patent-keyword smoke checks passed
- configuration, repository-hygiene, identity, epistemic declaration, architecture, and tier boundary guards passed
- study manifest validation passed

### DIFF
--- a/sbir_etl/reporting/procurement_transition/cet_vocabulary.py
+++ b/sbir_etl/reporting/procurement_transition/cet_vocabulary.py
@@ -6,8 +6,8 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Any
 
-import yaml
-
+from sbir_etl.config.yaml_io import read_yaml_mapping
+from sbir_etl.exceptions import ConfigurationError
 from sbir_etl.utils.procurement_text import find_lineage_phrases
 
 
@@ -24,11 +24,15 @@ def load_cet_vocabulary(path: str | None = None) -> dict[str, tuple[str, ...]]:
 
     taxonomy_path = Path(path) if path is not None else DEFAULT_TAXONOMY_PATH
     try:
-        data = yaml.safe_load(taxonomy_path.read_text(encoding="utf-8"))
-    except (OSError, yaml.YAMLError):
+        data = read_yaml_mapping(
+            taxonomy_path,
+            description="CET taxonomy",
+            allow_empty=True,
+        )
+    except ConfigurationError:
         return {}
     vocabulary: dict[str, tuple[str, ...]] = {}
-    for area in (data or {}).get("cet_areas", []):
+    for area in data.get("cet_areas", []):
         name = str(area.get("name", "")).strip()
         keywords = tuple(str(keyword).strip() for keyword in area.get("keywords", []) if keyword)
         if name and keywords:
--- a/scripts/ci/check_config_boundaries.py
+++ b/scripts/ci/check_config_boundaries.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+"""Prevent production code from bypassing the shared YAML configuration reader."""
+
+from __future__ import annotations
+
+import ast
+import subprocess
+from dataclasses import dataclass
+from pathlib import Path
+
+
+REPOSITORY_ROOT = Path(__file__).resolve().parents[2]
+CANONICAL_YAML_READERS = frozenset(
+    {
+        "sbir_etl/config/loader.py",
+        "sbir_etl/config/yaml_io.py",
+    }
+)
+IGNORED_PREFIXES = ("scripts/archive/", "tests/")
+
+
+@dataclass(frozen=True)
+class ConfigBoundaryViolation:
+    """One direct PyYAML read outside the configuration primitive."""
+
+    path: str
+    line_number: int
+
+    def format(self) -> str:
+        return (
+            f"{self.path}:{self.line_number}: direct yaml.safe_load bypasses the "
+            "configuration primitive; use sbir_etl.config.yaml_io.read_yaml_mapping "
+            "or sbir_etl.config.get_config"
+        )
+
+
+def _safe_load_aliases(tree: ast.AST) -> tuple[set[str], set[str]]:
+    """Return aliases for the yaml module and directly imported safe_load function."""
+
+    module_aliases: set[str] = set()
+    function_aliases: set[str] = set()
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Import):
+            for alias in node.names:
+                if alias.name == "yaml":
+                    module_aliases.add(alias.asname or alias.name)
+        elif isinstance(node, ast.ImportFrom) and node.module == "yaml":
+            for alias in node.names:
+                if alias.name == "safe_load":
+                    function_aliases.add(alias.asname or alias.name)
+    return module_aliases, function_aliases
+
+
+def _is_safe_load_call(
+    node: ast.Call,
+    *,
+    module_aliases: set[str],
+    function_aliases: set[str],
+) -> bool:
+    function = node.func
+    if isinstance(function, ast.Name):
+        return function.id in function_aliases
+    return (
+        isinstance(function, ast.Attribute)
+        and function.attr == "safe_load"
+        and isinstance(function.value, ast.Name)
+        and function.value.id in module_aliases
+    )
+
+
+def scan_file(
+    path: Path, *, repository_root: Path = REPOSITORY_ROOT
+) -> list[ConfigBoundaryViolation]:
+    """Find direct ``yaml.safe_load`` calls in one production Python file."""
+
+    relative = path.resolve().relative_to(repository_root.resolve()).as_posix()
+    if relative in CANONICAL_YAML_READERS or relative.startswith(IGNORED_PREFIXES):
+        return []
+
+    tree = ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
+    module_aliases, function_aliases = _safe_load_aliases(tree)
+    return [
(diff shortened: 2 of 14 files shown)


scripts/ci/check_config_boundaries.py
```diff
@@ -0,0 +1,123 @@
...
+) -> bool:
+    function = node.func
+    if isinstance(function, ast.Name):
+        return function.id in function_aliases
+    return (
+        isinstance(function, ast.Attribute)
+        and function.attr == "safe_load"
+        and isinstance(function.value, ast.Name)
+        and function.value.id in module_aliases
+    )
```
hollomancer, line 68: **The guard matches the literal name `safe_load` and nothing else, so the most idiomatic bypass sails straight through.**

Line 65 tests `function.attr == "safe_load"`, and the `ImportFrom` branch at line 49 only tracks `safe_load` — so all of these are invisible to it:

```python
yaml.load(f, Loader=yaml.SafeLoader)   # the standard pre-safe_load idiom
yaml.full_load(f)
yaml.unsafe_load(f)
yaml.safe_load_all(f)
from yaml import load
```

I ran 17 adversarial samples through it. It correctly catches all four `safe_load` spellings — `yaml.safe_load`, `import yaml as y`, `from yaml import safe_load`, and `from yaml import safe_load as sl` — and correctly exempts the canonical readers and `tests/`. It misses every form above.

The realistic failure isn't malice, it's habit: someone writes `yaml.load(path.read_text(), Loader=yaml.SafeLoader)` because that's the muscle memory from before `safe_load` existed, CI stays green, and the eighth divergent reader lands — with the guard in place and reporting success. `yaml.unsafe_load` is the uncomfortable one, since the guard's name and its position next to the security scan imply a coverage it doesn't have.

One-line fix:

```python
and function.attr in {"safe_load", "load", "full_load", "unsafe_load", "safe_load_all", "load_all"}
```

with the matching widening at line 49.

**Second scoping gap: `git ls-files "*.py"` at line 96 means notebooks aren't scanned, and there's a live reader in one.** `notebooks/examples/technology_area_cohort_review.ipynb`, cell 2:

```python
config = yaml.safe_load(CONFIG_PATH.read_text())
```

reading `config/transition_reports/nanotechnology.yaml` — the same file `ReportPaths.load_config()` reads, which is one of the seven sites this PR converts. So the guard passes while the duplication it exists to prevent persists. CLAUDE.md's notebook rules name this directly ("Import existing project functions … rather than copying logic"). Notebooks are exploratory tier so I wouldn't block on it, but it's the reason the steering-doc claim in my summary comment needs softening — the contract isn't as enforced as the doc now says.

---

sbir_etl/reporting/procurement_transition/cet_vocabulary.py
```diff
@@ -24,11 +24,15 @@ def load_cet_vocabulary(path: str | None = None) -> dict[str, tuple[str, ...]]:
...
     try:
-        data = yaml.safe_load(taxonomy_path.read_text(encoding="utf-8"))
-    except (OSError, yaml.YAMLError):
+        data = read_yaml_mapping(
+            taxonomy_path,
+            description="CET taxonomy",
+            allow_empty=True,
+        )
+    except ConfigurationError:
         return {}
```
hollomancer, line 33: This is the one site where the conversion changes behavior in a way a caller would notice, and it changes it toward silence.

On main, a top-level list or scalar in `taxonomy.yaml` raised `AttributeError` — a loud crash on first call. Now `read_yaml_mapping` raises `ConfigurationError`, line 32 catches it, and the function returns `{}`. I confirmed the difference by running both trees against a list-shaped document: main crashes, the branch returns `{}` silently.

Concretely: someone restructures `config/cet/taxonomy.yaml` to a top-level list. Every procurement-transition packet from then on quietly carries no CET agreement fact — no exception, no warning, and because this is `@lru_cache`d the empty result is pinned for the process lifetime. The packets still generate and still look fine.

The docstring at lines 21-22 does promise degradation ("degrades to no CET fact rather than failing"), so the direction is defensible and I don't think it should be reverted. But "missing or unreadable" and "structurally wrong" are different situations, and only the first is what that sentence was written for. A malformed taxonomy is an authoring error someone needs to find out about.

Minimal fix — log before swallowing:

```python
except ConfigurationError as error:
    logger.warning("CET taxonomy unusable, degrading to no CET fact: %s", error)
    return {}
```

Which leaves the degradation behavior intact and makes the cause discoverable. Worth extending the docstring to say a structurally invalid document is included in "unreadable."

---

hollomancer (commented):
## The risk this refactor carries didn't materialize — success and empty paths are byte-identical at all seven sites

I checked the thing that usually goes wrong in a consolidation like this by running both trees side by side across seven cases per site (missing / empty / comment-only / list / bare string / invalid YAML / valid mapping) rather than reading for it:

- **Every success path and every empty-file path is identical** between main and the branch.
- `allow_empty=True` is applied at exactly the seven sites that previously wrote `or {}` — no more, no fewer.
- `allow_empty=True` returns `{}`, never `None` (`yaml_io.py:65-67`), so no caller can hit `None["key"]`.

**The highest-value bug class — a caller catching `yaml.YAMLError` that silently stops catching — does not exist here.** `git grep YAMLError` returns three hits repo-wide, all inside `loader.py` and `yaml_io.py` themselves. No caller anywhere catches `yaml.YAMLError`, `OSError`, or `AttributeError` around any converted site.

**The seven-site count is exactly right.** Running the new guard against main reports precisely the seven lines this PR converts, and against the branch it reports none.

**There is no merge conflict.** GitHub's `dirty` was stale — `mergeable_state` now reads `clean`, the branch is a single commit directly on current main, and `git merge-tree --write-tree origin/main <branch>` exits 0 with zero conflicting files.

Error-path *types* changed at six of seven sites (`FileNotFoundError` / `AttributeError` / `ParserError` → `ConfigurationError`). No live caller breaks — I checked each one — but that's a real change to a public classmethod's contract in `PublishedBaselineRegistry.load()`, and it isn't in the PR body.

### What I'd fix before un-drafting

Two inline comments: **the guard only recognizes `safe_load`**, so `yaml.load(f, Loader=...)`, `full_load`, and `unsafe_load` pass it (one-line fix), and **`cet_vocabulary` now degrades silently** where it used to crash, with no log line.

Third, and the one I'd weigh most: **the new steering-doc claim overstates what's enforced.** `docs/steering/epistemic-tiers.md:205-208` presents the single-reader contract as executable, and this PR simultaneously *deletes* the honest "Seven direct `yaml.safe_load` call sites remain" caveat from "What does not." Given the guard covers `.py` files only, `safe_load` only, and mechanism-only, the replacement bullet should carry those three qualifiers. A tier doc that overstates its own enforcement is a costlier error here than in most repos.

### The deeper thing, worth a decision rather than a fix

**The PR consolidates the mechanism and leaves the policy split** — which is what its own "Why" section describes as the problem. After this change, two readers of the same config family disagree about empty files:

- `config/transition_reports/<area>.yaml` — `transition_report_paths.py:131` is now `allow_empty=True` (empty → `{"area_id": ...}`), while `tech_area_cohort.py:53` keeps the strict default (empty → `ConfigurationError`).
- `config/cet/taxonomy.yaml` has **four** readers: `cet_vocabulary.py:27` (permissive, swallows errors) versus `defense_taxonomy_crosswalk.py:50`, `local_cet_classifier.py:132`, and `tech_area_cohort.py:43` (all strict).

Same file, same question, different answers. That's a smaller problem than seven hand-rolled loaders and this PR is a clear improvement — but "invisible divergence outside the existing configuration primitive" is now *visible* divergence inside it, and it'd be worth saying which way each family is supposed to resolve rather than leaving it to whoever next reads two of them.

Related: `yaml_io.py`'s docstring on main said explicitly that empty-as-valid "is a different policy, not a bug this helper fixes." This PR deletes that sentence and moves the policy into the primitive. I think that's the right call and the new text argues it well — but it reverses a stated decision in a `primitives`-tier module, and the PR body doesn't mention it. Worth a line, since 9 pre-existing callers still rely on the fail-closed default.

### Smaller notes

- `compute_tech_census.py:121` still catches `(FileNotFoundError, ValueError)` and not `ConfigurationError`, so a malformed profile hard-crashes the tool path instead of degrading. Not a regression, but unifying the error type was the point and this is the one caller that could now benefit.
- The guard duplicates ~50 lines of scanner harness from `check_identity_boundaries.py` near-verbatim. House style arguably endorses it, and I wouldn't block — but it's worth naming in a PR whose thesis is one implementation per concept.
- `IGNORED_PREFIXES` exempts only top-level `tests/`, so a future `packages/*/tests/` tree would be flagged. Latent — no such tree exists today.
- `fiscal_bea_mapper.py` moves from locale-default encoding to explicit UTF-8. Irrelevant on CI and Docker; strictly an improvement.

**Verification:** 78 tests across every touched module pass; `mypy sbir_etl packages/sbir-graph/sbir_graph` clean at 265 files; ruff check and format clean; all five boundary guards pass. Green CI including "Lint, Types, and Guards".

---

PR #571, opened 2026-08-09
Label tech census outputs non-citable

hollomancer:
## What changed

- adds a prominent exploratory/non-citable banner to the Massachusetts unmanned-systems readout
- emits canonical `_epistemic` metadata from the shared census engine
- preserves that metadata through successful and empty API results
- prints the status in the CLI and writes it into `summary.json`
- bumps the tech-census tool from 2.1.0 to 2.2.0 for the additive output contract

## Why

The census implementation was labeled exploratory in module prose, but generated results and the published Massachusetts readout did not reliably carry that status. The reviewed 238-award figure cannot currently be regenerated because its award-level review ledger was not retained.

## Impact

People reading or consuming a tech-census result now see `tier: exploratory` and `citable: false` at the output boundary, including zero-row/error results. This prevents a deterministic-but-contestable classifier result from being mistaken for evidence-tier output.

## Validation

- 57 focused tech-census core, API, and CLI tests passed
- repository-hygiene, epistemic-declaration, and tier-boundary guards passed
- `git diff --check` passed

### DIFF
--- a/sbir_etl/utils/tech_census.py
+++ b/sbir_etl/utils/tech_census.py
@@ -31,6 +31,19 @@ REPO_ROOT = Path(__file__).resolve().parents[2]
 CONFIG_DIR = REPO_ROOT / "config" / "tech_census"
 
 
+def census_epistemic_metadata() -> dict[str, Any]:
+    """Return the non-citable status that every census output must preserve."""
+
+    return {
+        "tier": EPISTEMIC_TIER,
+        "citable": False,
+        "notice": (
+            "Profile-driven technology classification is exploratory; "
+            "do not cite as an evidence-tier result."
+        ),
+    }
+
+
 class CensusAward(TypedDict, total=False):
     """Canonical award shape used by the census engine."""
 
@@ -536,6 +549,7 @@ def run_census(
             totals["usd"] = round(totals["usd"], 2)
 
     return {
+        "_epistemic": census_epistemic_metadata(),
         "area_id": compiled.area_id,
         "display_name": compiled.display_name,
         "config_version": compiled.version,
--- a/docs/research/massachusetts_unmanned_systems_sbir_readout.md
+++ b/docs/research/massachusetts_unmanned_systems_sbir_readout.md
@@ -1,5 +1,10 @@
 # Massachusetts Physical Unmanned-Systems SBIR/STTR Readout
 
+> **Exploratory / non-citable.** Use this readout for discovery, not as
+> evidence-tier support or an externally cited result. The 238-award reviewed
+> total cannot be regenerated until the original award-level review decisions
+> are reconstructed as a retained, auditable ledger.
+
 - **Data source:** SBIR.gov bulk `award_data.csv` local snapshot
 - **Source timestamp:** 2026-04-01T00:00:06.904233+00:00
 - **Source SHA-256:** `73d646fc6883ed93b36d19518b0d9442a9ebae94c5b49ad5a7fcd6d3c2b872dd`
--- a/packages/sbir-analytics/sbir_analytics/tools/tech_census/compute_tech_census.py
+++ b/packages/sbir-analytics/sbir_analytics/tools/tech_census/compute_tech_census.py
@@ -66,7 +66,7 @@ class ComputeTechCensusTool(BaseTool):
     """Classify awards into a versioned technology profile and aggregate them."""
 
     name = "compute_tech_census"
-    version = "2.1.0"
+    version = "2.2.0"
 
     def execute(
         self,
@@ -181,6 +181,7 @@ class ComputeTechCensusTool(BaseTool):
         results_df = pd.DataFrame(result["classified_awards"])
 
         summary = {
+            "_epistemic": result["_epistemic"],
             "area_id": result["area_id"],
             "display_name": result["display_name"],
             "config_version": result["config_version"],
@@ -235,11 +236,14 @@ class ComputeTechCensusTool(BaseTool):
 
     @staticmethod
     def _empty_result(area_id: str) -> dict[str, Any]:
+        from sbir_etl.utils.tech_census import census_epistemic_metadata
+
         return {
             "award_count": 0,
             "award_dollars": 0.0,
             "results": pd.DataFrame(),
             "summary": {
+                "_epistemic": census_epistemic_metadata(),
                 "area_id": area_id,
                 "grand_total": {"n": 0, "usd": 0.0},
                 "fy_totals": {},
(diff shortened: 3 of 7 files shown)


sbir_etl/utils/tech_census.py
```diff
@@ -31,6 +31,19 @@
...
+    """Return the non-citable status that every census output must preserve."""
+
+    return {
+        "tier": EPISTEMIC_TIER,
+        "citable": False,
+        "notice": (
+            "Profile-driven technology classification is exploratory; "
+            "do not cite as an evidence-tier result."
+        ),
+    }
```
hollomancer, line 44: "every census output must preserve" is the right standard, and the one output most likely to be shared doesn't meet it.

`scripts/data/build_tech_census.py` writes two artifacts side by side. `summary.json` gets `_epistemic` (line 226). `classified_awards.csv` — the award-grain table, written at `:190-221` — gets nothing.

That's the wrong way round for how these get used. The CSV is the file someone opens in a spreadsheet, filters, and pastes a count from; `summary.json` sitting next to it is the file nobody opens. So the artifact that can most easily produce an uncited number is the one carrying no marker, while the marker lives in the file that already had the least risk of being mistaken for evidence.

I don't think this needs to be solved perfectly — a CSV has no natural place for structured metadata. But a leading comment row, or a sibling `classified_awards.NOTICE.txt`, or simply naming the file `classified_awards.exploratory.csv` would each close it. Worth doing in this PR, since the docstring here is what sets the standard the CSV then misses.

Two smaller notes on the same helper:

**"Canonical" is overstated in the PR body.** This helper is census-local. `sbir_etl/reporting/tech_area_cohort.py:488,1027` and `sbir_etl/reporting/phase_transition_analysis.py:153` still hand-roll their own `_epistemic` literals, and `tech_area_cohort` uses a *different* shape — `cohort_profile` where this uses `notice`. Exploratory tier explicitly permits that duplication and CLAUDE.md's Surgical Changes rule argues against unifying it here, so I'd leave the code alone and just soften the wording.

**The key and shape are otherwise well-chosen.** `_epistemic` with `{tier, citable, notice}` exactly matches the pre-existing `phase_transition_analysis.py:153-157`, and `assets/transition_report.py:107-108` already reads `(summary.get("_epistemic") or {}).get("tier")` from a sibling JSON — so this is the right convention for the one existing reader. Worth knowing it's a *de facto* convention only: `docs/steering/epistemic-tiers.md` never names `_epistemic` (it says operated exploratory work "must emit `citable: false` metadata"), the Dagster side uses flat `epistemic_tier` / `citable` keys instead, and no CI guard validates the key on any path. The three unit assertions this PR adds are the entire enforcement.

---

hollomancer (commented):
## Mergeable. The banner makes the strong claim, and the versioning question resolves in your favor

Two things I went in skeptical about, both of which hold up:

**The banner states non-reproducibility, not just "exploratory."** Those are different claims and only the stronger one is warranted here:

> **Exploratory / non-citable.** Use this readout for discovery, not as evidence-tier support or an externally cited result. The 238-award reviewed total cannot be regenerated until the original award-level review decisions are reconstructed as a retained, auditable ledger.

And there's no unlabeled copy elsewhere: `git grep -nw 238` on the branch hits only this readout (lines 4, 19, 162, 169) plus three unrelated coincidences (a CDK test count, a Northrop contract count, a URL fragment). `docs/research-questions.md` has no unmanned-systems or tech-census entry, so no research question is marked answerable on this work.

Worth stating plainly though: the non-regenerability finding was **already on main** at readout lines 157-165. The contribution is hoisting a buried caveat into a top banner — which is genuinely valuable, since the buried version was below the number it disclaims — but the PR body reads as if it's a new discovery.

**The version bump is not a synchronized-release violation.** `scripts/ci/check_versioning.py:12-21` compares a closed set: the four `pyproject.toml` files, `uv.lock`, `sbir_etl/__init__.py`, and `config/base.yaml`. It never parses class attributes, and `ComputeTechCensusTool.version` is a `BaseTool` class attribute surfaced only in `ToolMetadata.tool_version`. Corroborated three ways: `docs/steering/versioning.md` enumerates the same closed set; every other tool sits at `1.0.0` while the repo is at `0.3.0`, so main would already be failing if tool versions were in scope; and I ran it — `versioning check passed for 0.3.0`. MINOR is the right increment for an additive key, and nothing pins `"2.1.0"`.

**`_epistemic` really is on every reachable path.** `run_census` has exactly one `return` and no early-out for empty input, so empty falls through the same aggregation. `execute()` has six `return ToolResult` statements — five route through `_empty_result`, which now carries the metadata, and the sixth is the success path. The error path is a non-issue rather than a gap: `BaseTool.run()` catches, logs, and **re-raises** without ever constructing a `ToolResult`, so no unlabeled result object can be serialized.

**No stdout hazard and no consumer breakage.** The CLI has no JSON-on-stdout mode (its only `json.dumps` writes to a file), and grepping workflows, Makefile, docs, and notebooks for `build_tech_census` returns nothing that parses its output. Nothing in the repo reads `summary.json` for this tool, and the one structurally similar consumer uses defensive `.get()` access, so an additive key is safe.

### What I'd add

One inline comment: **`classified_awards.csv` carries no marker**, while the new helper's docstring says "every census output must preserve" it. The CSV is the artifact most likely to be opened, filtered, and quoted from — so the marker is present on the low-risk output and absent from the high-risk one.

**Second, and the bigger gap for the PR's stated purpose:** the readout is not in `docs/research/README.md`. That's the index carrying the "Evidence status" column and the "do not cite unless `study.yaml` says citable" rule, and `git grep massachusetts_unmanned` returns only the file itself — the document is entirely unlinked from anywhere in the repo. So this PR's label is visible only to someone who already found the file, and an auditor enumerating research outputs from the index never learns it exists. Adding a row there would do more for the PR's goal than the banner does.

### Minor

- The `## Result` table still reads `| Relevant awards | **238** |` under "Conservative reviewed result." The banner is ten lines up and does carry the caveat, and `exploratory`'s contract is satisfied by labeling — so this is presentation, not a contract breach. But a skimmer jumping to `## Result` meets a bolded number first.
- `tests/unit/scripts/test_build_tech_census.py:55` asserts only the `Status: EXPLORATORY / NON-CITABLE` header, not the notice line the CLI prints below it. The notice string could be emptied and the test would still pass.

**Verification:** 57 tests across the three touched test files pass; ruff check and format clean on all four Python files; `check_epistemic_tiers.py`, `check_tier_boundaries.py`, `check_architecture_boundaries.py`, and `check_versioning.py` all pass against the branch. Branch is one commit on current main, clean fast-forward, and all 16 checks are green.

---

PR #573, opened 2026-08-09
Make fiscal runtime claims truthful

hollomancer:
## Summary

- replace obsolete R/StateIO/USEEIOR instructions with the current Python BEA workflow
- label mock fiscal outputs as illustrative and non-citable
- remove the fabricated mock confidence score
- remove the empty `cloud` extra and nonexistent `r` extra
- synchronize `uv.lock` with the supported dependency contract

## Why

The repository currently advertises an R runtime, Docker packages, and dependency extras that do not exist. It also describes exploratory fiscal estimates as research-grade production output. These claims make the public setup surface unreliable and conflict with the project's evidence-tier policy.

## Impact

The supported fiscal path is now explicit: the deterministic mock demonstrates data flow, while the BEA-backed Python example requires `BEA_API_KEY` and remains exploratory. The heavier Docker image is intentionally retained until the default locked image covers its valid ML dependencies.

## Verification

- `uv lock --check`
- 41 focused fiscal/tool tests passed
- mock fiscal example executed end to end
- Ruff check and format check
- `make docs-check`
- `git diff --check`

### DIFF
--- a/docs/architecture/dagster-pipelines.md
+++ b/docs/architecture/dagster-pipelines.md
@@ -52,7 +52,7 @@ import, its job is replaced by an empty **placeholder** job (name suffixed
 | Job (`name`) | Selection | Notes |
 |---|---|---|
 | `cet_full_pipeline_job` | 8 CET keys (classifications, profiles, Neo4j CET nodes/edges) | Ships run config (parquet/json paths, `batch_size: 1000`, create constraints/indexes) |
-| `fiscal_returns_mvp_job` | 9 fiscal keys (NAICS→BEA, inflation, economic/tax) | Exploratory Python/BEA model; needs at least 8 GB RAM |
+| `fiscal_returns_mvp_job` | 9 fiscal keys (NAICS→BEA, inflation, economic/tax) | Exploratory Python/BEA model; memory requirement is not yet measured |
 | `fiscal_returns_full_job` | asset groups `fiscal_data_prep`, `economic_modeling`, `tax_calculation`, `sensitivity_analysis` | Full fiscal run incl. sensitivity sweep |
 | `modernbert_job` | ModernBERT award/patent embeddings + similarity | Requires ML extras |
 | `uspto_ai_extraction_job` | USPTO AI extract + dedup + human sample | — |
--- a/docs/architecture/dagster-pipelines.md
+++ b/docs/architecture/dagster-pipelines.md
@@ -52,7 +52,7 @@ import, its job is replaced by an empty **placeholder** job (name suffixed
 | Job (`name`) | Selection | Notes |
 |---|---|---|
 | `cet_full_pipeline_job` | 8 CET keys (classifications, profiles, Neo4j CET nodes/edges) | Ships run config (parquet/json paths, `batch_size: 1000`, create constraints/indexes) |
-| `fiscal_returns_mvp_job` | 9 fiscal keys (NAICS→BEA, inflation, economic/tax) | R/StateIO; needs at least 8 GB RAM |
+| `fiscal_returns_mvp_job` | 9 fiscal keys (NAICS→BEA, inflation, economic/tax) | Exploratory Python/BEA model; needs at least 8 GB RAM |
 | `fiscal_returns_full_job` | asset groups `fiscal_data_prep`, `economic_modeling`, `tax_calculation`, `sensitivity_analysis` | Full fiscal run incl. sensitivity sweep |
 | `modernbert_job` | ModernBERT award/patent embeddings + similarity | Requires ML extras |
 | `uspto_ai_extraction_job` | USPTO AI extract + dedup + human sample | — |
--- a/examples/README_FISCAL_EXAMPLE.md
+++ b/examples/README_FISCAL_EXAMPLE.md
@@ -43,13 +43,24 @@ input-output tables. Its tax and employment estimates remain exploratory and
 non-citable; see the epistemic-tier declaration in
 `sbir_etl/transformers/sbir_fiscal_pipeline.py`.
 
-Both examples expect awards with these columns:
+When `BEA_API_KEY` is absent, or a BEA request fails, this path falls back to
+hard-coded placeholder multipliers. Check `quality_flags`: only
+`bea_api_with_ratios` and `bea_api_default_ratios` identify BEA-backed rows;
+`placeholder_computation` is illustrative fallback output.
 
-- `award_id`
+The Python and mock calculator examples require these award columns:
+
 - `award_amount`
 - `state`
 - `naics_code`
 - `fiscal_year`
 
+The district example additionally requires `company_address`, `company_city`,
+`company_state`, and `company_zip` for geographic resolution. Its sample also
+includes `award_id` and `company_name` for readability, but neither calculator
+requires them.
+
 For the project-wide evidence boundaries, see the
 [research output status index](../docs/research/README.md).
+For the fiscal pipeline schema and quality flags, see the
+[fiscal pipeline guide](../docs/fiscal/sbir-fiscal-pipeline-guide.md).
--- a/examples/README_FISCAL_EXAMPLE.md
+++ b/examples/README_FISCAL_EXAMPLE.md
@@ -1,218 +1,55 @@
-# SBIR Fiscal Impact Example
+# SBIR Fiscal Impact Examples
 
-This directory contains examples demonstrating the SBIR fiscal impact analysis pipeline.
+These examples demonstrate the shape of the exploratory fiscal-impact pipeline.
+They are integration aids, not validated economic findings.
 
-## Files
+## Examples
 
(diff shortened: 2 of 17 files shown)


examples/README_FISCAL_EXAMPLE.md
```diff
@@ -1,218 +1,55 @@
...
+  illustrative and must not be cited.
+- `sbir_fiscal_impact_by_district_example.py` demonstrates how the mock output
+  can be grouped geographically. It carries the same evidence limitation.
 
-## Quick Start (Mock Version - No R Required)
+The repository no longer ships an R, StateIO, or USEEIOR runtime. Docker Compose
+does not install those tools, and there is no `r` dependency extra.
 
-The mock version provides a demonstration of the pipeline without requiring R installation:
+## Run the deterministic demonstration
```
hollomancer, line 19: "**Both** examples" sits under a list of **three** (lines 8-14), and the five columns below are wrong for the third one.

`sbir_fiscal_impact_by_district_example.py` also needs `company_address`, `company_city`, `company_state`, and `company_zip` — its resolver keys on `row.get("company_state")` / `row.get("company_city")` at lines 68-70.

The failure is quiet, which is what makes it worth fixing: a user who supplies exactly these five columns gets **no crash**. Every row's district resolves to `None`, the district groupbys drop them all, and the "435 districts" report comes out empty. They'll assume their data is wrong.

Also, `award_id` is listed as expected but isn't in either calculator's `required_columns` (`sbir_etl/transformers/sbir_fiscal_pipeline.py:137`; mock at `:120`).

This is new text, so the PR owns it — and it's the same defect class the PR exists to fix, one section down.

**One more thing about this section**, which I'd weigh higher than the column list: the new "Run the current BEA-backed implementation" block never says the real example *degrades to hard-coded placeholder multipliers* when `BEA_API_KEY` is absent (`bea_io_adapter.py:221-227`) or when the BEA fetch raises (`:245-253` → `_compute_placeholder_impacts` at `:455-474`). The deleted "Differences Between Mock and Real Versions" table at least drew that line.

It's not fully silent — loguru warns on stderr and `placeholder_computation` shows under DATA QUALITY — but `examples/sbir_fiscal_impact_example.py:92` prints `✓ BEA I-O economic models ready` **unconditionally, before any key check**. So the file this README now promotes as the real path opens with an untrue runtime claim. One sentence here closes it.

---

docs/architecture/dagster-pipelines.md
```diff
@@ -52,7 +52,7 @@ import, its job is replaced by an empty **placeholder** job (name suffixed
 | Job (`name`) | Selection | Notes |
 |---|---|---|
 | `cet_full_pipeline_job` | 8 CET keys (classifications, profiles, Neo4j CET nodes/edges) | Ships run config (parquet/json paths, `batch_size: 1000`, create constraints/indexes) |
-| `fiscal_returns_mvp_job` | 9 fiscal keys (NAICS→BEA, inflation, economic/tax) | R/StateIO; needs at least 8 GB RAM |
+| `fiscal_returns_mvp_job` | 9 fiscal keys (NAICS→BEA, inflation, economic/tax) | Exploratory Python/BEA model; needs at least 8 GB RAM |
```
hollomancer: This rewrite (`R/StateIO` → `Exploratory Python/BEA model`) deletes the only thing that ever justified the 8 GB number, then keeps the number and re-attributes it to the Python path.

The figure came from StateIO's state-level R models — the deleted `README_FISCAL_EXAMPLE.md` said so directly ("StateIO can use significant RAM… increase Docker memory limit to 4GB+"). The Python path does something much smaller: fetches summary-level **national** BEA Use tables and inverts a ~71×71 matrix (`bea_io_adapter.py:255-279`).

Nothing in the repo substantiates 8 GB for that. The only other references are `specs/archive/e2e-testing-enhancement/*` (about a MacBook Air) and `docs/archive/performance/legacy-performance-guide.md`. And the runbook contradicts it — `docs/deployment/self-hosted-server.md:568-570` says fiscal jobs are "capacity-gated… measure memory before scheduling," i.e. explicitly unmeasured.

An operator sizing the self-hosted host provisions 8 GB of headroom for `fiscal_returns_mvp_job` on a number that a truthfulness PR just re-endorsed under a new justification. Either drop the clause or replace it with something honest — "memory not yet measured; see workload placement" — which is what the runbook already says.

---

hollomancer (commented):
## A real truthfulness improvement — every claim it *adds* checks out. The gaps are all incompleteness in the same genre.

The dependency surgery is genuinely correct and genuinely a no-op: main's root `pyproject.toml` had `cloud = [ ]` (literally empty), main's `uv.lock` never listed `r` in `provides-extras` for `sbir-etl`, so `packages/sbir-analytics`'s `r = ["sbir-etl[r]"]` was dangling at nothing. The lock already resolved extras to `["monitoring","uspto"]` at `uv.lock:2589`, so the installed set is byte-identical before and after. `uv lock --check` exits 0.

I reproduced the verification: mock example runs end-to-end from a clean extraction (exit 0), the district example runs too (confirming the removed `confidence` column breaks no consumer), ruff check and format clean, `make docs-check` passes, and the `9 fiscal keys` count matches `fiscal_returns_job.py:26-36` exactly. **No true claim was deleted** — everything removed was R/StateIO-specific, a stale `src/` path, or duplicated verbatim in `docs/fiscal/sbir-fiscal-pipeline-guide.md`. The tier labeling is correct and doesn't promote anything: `EPISTEMIC_TIER = "exploratory"` was already on `sbir_fiscal_pipeline.py:31`.

Two inline comments on the sharper items. Beyond those:

### The confidence score was removed from the mock and left alive on the real path

This is the one I'd most want fixed, because the PR's own framing makes it worse. The mock lost `result_df["confidence"] = 0.85  # Mock confidence score`. The identical constant is still on the production path:

```
bea_io_adapter.py:374   "confidence": Decimal("0.85"),     # BEA success branch
bea_io_adapter.py:444   result_df["confidence"] = 0.85     # _ensure_impact_columns
bea_io_adapter.py:473   result_df["confidence"] = Decimal("0.75")   # placeholder path
```

and `examples/sbir_fiscal_impact_example.py:198-199` still prints it as `Average Confidence Score`. So the two sibling examples now disagree about whether the column exists, and the PR removed the low-traffic instance of the falsehood while leaving the high-traffic one — on the very script this README newly promotes as "the current BEA-backed implementation."

It's documented as a real quality signal too: `docs/fiscal/sbir-fiscal-pipeline-guide.md:306` and `:331-347` include a worked example showing `Average Confidence Score: 0.40`, a value the code can never emit. The only reachable values are {0.0, 0.75, 0.85}, which makes the guide's `reliable = impacts[impacts["confidence"] > 0.7]` filter a no-op that drops nothing but hard failures.

### An obsolete R claim survives in the file this commit edits

`examples/sbir_fiscal_impact_example_mock.py:6`:

> This version works WITHOUT R and provides realistic demo outputs.

Both halves are wrong now. There is no R alternative to work "without" — the README's own line 16 says so — and "realistic" contradicts the "illustrative and must not be cited" label this same commit adds at line 269. The module docstring is the likelier entry point than the README for anyone opening the file.

Same file, line 285: `print("  ✓ Using simplified economic multipliers (BEA_API_KEY not set)")` fires unconditionally, including when the key *is* set. Pre-existing, but squarely in scope for this PR.

### Smaller

- The rewrite drops "Understanding the Output", "Key Metrics", and "Additional Resources" — all of which were **accurate** — and drops the `../docs/fiscal/` link at the same time, so the surviving copy of that content is no longer reachable from this README. Near-zero information loss, but navigation regressed; a one-line pointer to `../docs/fiscal/sbir-fiscal-pipeline-guide.md` restores it. Credit for the same rewrite fixing a stale `src/transformers/sbir_fiscal_pipeline.py` path that `check_removed_src_references.py` doesn't catch, since `examples/` falls outside its `docs/`/`specs/` scope.
- Out of your declared scope, but relevant to the framing: `Dockerfile.full:6,13-19` still declares `Used by: AWS Batch` and pip-installs `boto3` + `cloudpathlib[s3]` — the exact packages `Dockerfile:13` calls "intentionally absent: the AWS data plane is retired," and the reason the `cloud` extra you just deleted was empty. Probably belongs in #580 rather than here.

### On the stack

**#580 is based on this branch**, so all of the above propagates into it. None are hard breakages, so #580 isn't endangered — but the `Dockerfile.full` item is the one that actually touches image building, so it's better resolved there than here.

Merges clean into current main (`git merge-tree` exits 0), all 16 checks green.

One verification note: "41 focused fiscal/tool tests passed" isn't reproducible as stated — I ran the three closest modules (`test_sbir_fiscal_pipeline`, `test_bea_io_adapter`, `test_fiscal_bea_mapper`) and got 47 passed, 0 failed. Not a discrepancy in outcome, just an unrecoverable selection; worth naming the paths in the body.

---

PR #574, opened 2026-08-09
Report canonical Neo4j load metrics

hollomancer:
## Summary

- calculate update counts from the canonical `FinancialTransaction`, `Organization`, and `Individual` labels
- checkpoint cumulative Organization metrics so company and research-institution counts remain distinct
- add a focused test for canonical-label and delta behavior

## Why

The SBIR loader migrated away from `Award`, `Company`, `Researcher`, and `ResearchInstitution` labels, but its run summary still queried those retired keys. Successful updates could therefore be reported as zero.

## Impact

Existing result-field names remain stable for callers. Their values now reflect the graph labels the loader actually writes.

## Verification

- `pytest tests/unit/assets/test_sbir_neo4j_loading.py -q` — 3 passed
- Ruff check and format check
- `git diff --check`

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
@@ -67,6 +67,12 @@ def _company_id_for_award(award: Award) -> str | None:
     return None
 
 
+def _updated_since(metrics: Any, label: str, previous: int = 0) -> tuple[int, int]:
+    """Return updates for one load step and the new cumulative checkpoint."""
+    current = int(metrics.nodes_updated.get(label, 0))
+    return current - previous, current
+
+
 def _get_neo4j_client() -> "Neo4jClient | None":
     """Open a Neo4j connection from config; return None when SKIP_NEO4J_LOADING is set."""
     skip_neo4j = os.getenv("SKIP_NEO4J_LOADING", "false").lower() in ("true", "1", "yes")
@@ -485,6 +491,7 @@ def neo4j_sbir_awards(
         award_institution_rels: list[tuple] = []
         researcher_award_rels: list[tuple] = []
         researcher_company_rels: list[tuple] = []
+        organization_updates = 0
 
         # Error counters (logged at end; first 10 of each kind go to debug log).
         skipped_zero_amount = 0
@@ -674,6 +681,9 @@ def neo4j_sbir_awards(
                 f"{metrics.nodes_created.get('Organization', 0)} created, "
                 f"{metrics.nodes_updated.get('Organization', 0)} updated"
             )
+        companies_updated, organization_updates = _updated_since(
+            metrics, "Organization", organization_updates
+        )
 
         if award_nodes:
             metrics = client.batch_upsert_nodes(
@@ -683,6 +693,7 @@ def neo4j_sbir_awards(
                 metrics=metrics,
             )
             context.log.info(f"Loaded {len(award_nodes)} FinancialTransaction nodes (AWARD type)")
+        awards_updated, _ = _updated_since(metrics, "FinancialTransaction")
 
         metrics = _load_nodes(
             client,
@@ -693,6 +704,7 @@ def neo4j_sbir_awards(
             description="researchers",
             context=context,
         )
+        researchers_updated, _ = _updated_since(metrics, "Individual")
         metrics = _load_nodes(
             client,
             label="Organization",
@@ -701,6 +713,9 @@ def neo4j_sbir_awards(
             metrics=metrics,
             description="research institutions",
             context=context,
+        )
+        institutions_updated, organization_updates = _updated_since(
+            metrics, "Organization", organization_updates
         )
 
         for rels, desc in (
@@ -871,13 +886,13 @@ def neo4j_sbir_awards(
         result = {
             "status": "success",
             "awards_loaded": len(award_nodes),
-            "awards_updated": metrics.nodes_updated.get("Award", 0),
+            "awards_updated": awards_updated,
             "companies_loaded": len(company_nodes_map),
-            "companies_updated": metrics.nodes_updated.get("Company", 0),
+            "companies_updated": companies_updated,
             "researchers_loaded": len(researcher_nodes_map),
-            "researchers_updated": metrics.nodes_updated.get("Researcher", 0),
+            "researchers_updated": researchers_updated,
             "institutions_loaded": len(institution_nodes_map),
-            "institutions_updated": metrics.nodes_updated.get("ResearchInstitution", 0),
+            "institutions_updated": institutions_updated,
             "relationships_created": sum(metrics.relationships_created.values()),
             "errors": metrics.errors,
             "duration_seconds": duration,
--- a/packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
@@ -67,10 +67,20 @@ def _company_id_for_award(award: Award) -> str | None:
     return None
 
 
+def _count_since(metrics: Any, field: str, label: str, previous: int = 0) -> tuple[int, int]:
+    """Return a non-negative counter delta and its new cumulative checkpoint."""
+    current = int(getattr(metrics, field).get(label, 0))
+    return max(0, current - previous), current
+
+
 def _updated_since(metrics: Any, label: str, previous: int = 0) -> tuple[int, int]:
     """Return updates for one load step and the new cumulative checkpoint."""
-    current = int(metrics.nodes_updated.get(label, 0))
-    return current - previous, current
+    return _count_since(metrics, "nodes_updated", label, previous)
+
+
+def _created_since(metrics: Any, label: str, previous: int = 0) -> tuple[int, int]:
+    """Return creations for one load step and the new cumulative checkpoint."""
+    return _count_since(metrics, "nodes_created", label, previous)
 
 
 def _get_neo4j_client() -> "Neo4jClient | None":
@@ -491,6 +501,7 @@ def neo4j_sbir_awards(
         award_institution_rels: list[tuple] = []
         researcher_award_rels: list[tuple] = []
         researcher_company_rels: list[tuple] = []
+        organization_creations = 0
         organization_updates = 0
 
         # Error counters (logged at end; first 10 of each kind go to debug log).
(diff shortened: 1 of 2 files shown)


packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
```diff
@@ -871,13 +886,13 @@ def neo4j_sbir_awards(
...
+            "awards_updated": awards_updated,
             "companies_loaded": len(company_nodes_map),
-            "companies_updated": metrics.nodes_updated.get("Company", 0),
+            "companies_updated": companies_updated,
             "researchers_loaded": len(researcher_nodes_map),
-            "researchers_updated": metrics.nodes_updated.get("Researcher", 0),
+            "researchers_updated": researchers_updated,
             "institutions_loaded": len(institution_nodes_map),
-            "institutions_updated": metrics.nodes_updated.get("ResearchInstitution", 0),
+            "institutions_updated": institutions_updated,
```
hollomancer, line 924: The four `*_updated` fields are now canonical. The four `*_loaded` fields interleaved with them still aren't, and they're the ones a reader trusts.

`awards_loaded = len(award_nodes)` counts rows **submitted**, not nodes written. `metrics.nodes_created` is computed and carried into `result["metrics"]` at line 909, but no headline field reads it.

That matters because a load check consumes it. Concretely: 10 batches × 1000 award rows, one batch raises inside `batch_upsert_nodes` (`client.py:348-350` → `metrics.errors += 1000`, zero nodes written). The result reports `awards_loaded: 10000` when 9000 were written, and `neo4j_sbir_awards_load_check` at line 984 computes 1000/10000 = 10%, stays under its 25% threshold, and emits **"✓ Neo4j load successful: 10000 awards"**. The check passes on a number that is wrong in the optimistic direction.

It's pre-existing, so I wouldn't block — but it's the same defect this PR is titled after, one field away, and it's a one-liner each:

```python
"awards_loaded": metrics.nodes_created.get("FinancialTransaction", 0),
```

If you'd rather keep "submitted" as a distinct concept, that's defensible — but then the pair wants renaming (`awards_submitted` / `awards_written`), because `loaded` next to `updated` reads as "written" to everyone.

Two smaller things in the same block, both latent rather than live:

`_updated_since` (line 70-73) returns `current - previous` with no clamp, so a cumulative counter that *decreases* yields a silent negative. That happens the moment `metrics` is rebound to a different object — which line 822 already does (`create_subsidiary_relationships` returns the loader's own fresh `LoadMetrics`). That path is currently unreachable, because `getattr(award, "sub_agency", None)` at line 756 is always `None` (`Award` has no such field and `sbir_etl/models/award.py:184` sets no `extra`, so pydantic drops it). But if someone fixes that and it lands above line 717, `institutions_updated` becomes `0 - 40 = -40` with no error. `max(0, current - previous)` closes it cheaply.

The same rebind would also collapse `relationships_created` at line 896 from every relationship type down to just `SUBSIDIARY_OF`, and reset the nested `nodes_updated`/`nodes_created` blocks to `{}`. Your four new fields are now the only ones immune to it. Worth a line in the PR body so the asymmetry is on the record.

---

tests/unit/assets/test_sbir_neo4j_loading.py
```diff
@@ -2,7 +2,24 @@
...
+        self.nodes_updated = nodes_updated
+
+
+def test_updated_since_reports_canonical_label_delta():
+    metrics = _Metrics({"FinancialTransaction": 7, "Organization": 11, "Individual": 3})
+
+    assert _updated_since(metrics, "FinancialTransaction") == (7, 7)
+    assert _updated_since(metrics, "Organization", previous=5) == (6, 11)
+    assert _updated_since(metrics, "Individual") == (3, 3)
+    assert _updated_since(metrics, "Award") == (0, 0)
```
hollomancer, line 35: **This test can't catch the bug the PR fixes.** It exercises `_updated_since`'s subtraction against a hand-built stub, not the asset's label choices or the checkpoint threading — which is where the actual defect lived.

Demonstrated two ways, both of which leave all 3 tests green:

- Revert line 707 to `_updated_since(metrics, "Researcher")` — back to a retired label, `researchers_updated` returns to a hard zero. **Tests pass.**
- Delete the `organization_updates` threading at 717-719 — `institutions_updated` becomes companies+institutions (40+11 = 51 instead of 11). **Tests pass.**

Both are exactly the regression class this PR exists to prevent.

Line 22's `_updated_since(metrics, "Award") == (0, 0)` looks like it guards the retired label, but it only asserts that a key absent from a dict yields zero — it says nothing about whether the asset still reads `"Award"`.

Worth noting `_Metrics` doesn't need to be a stub at all: the real `LoadMetrics` constructs fine as `LoadMetrics(nodes_updated={...})`. Using it would also let line 70 keep the quoted `"LoadMetrics"` forward ref that its neighbour `_load_nodes` (424-433) uses, instead of the weaker `Any` that exists only to accommodate the stub.

An asset-level test with a stub client is about 60 lines and does catch all of the above. I wrote one to confirm the fix is real before saying so — planted 40 `Organization` updates in the company step, 7 `FinancialTransaction`, 3 `Individual`, and 11 `Organization` in the institution step, **across 2 batches per step**, and got exactly `companies_updated: 40, awards_updated: 7, researchers_updated: 3, institutions_updated: 11`. No double-count at batch boundaries, no cross-step bleed. The fix is correct; it's just unpinned.

This matters more than usual because CI gives no cover here: the green "Neo4j Integration Tests" job doesn't touch this asset — `tests/integration/neo4j/*` exercises the client primitives directly (`test_multi_key_merge.py:53-157` asserts `nodes_created`/`nodes_updated` for `Organization`). So this helper test is the only thing standing over the attribution logic.

---

hollomancer (commented):
## The fix is right and the arithmetic checks out — I verified it end-to-end rather than by reading

The diagnosis is understated in the PR body. Those four keys (`Award`, `Company`, `Researcher`, `ResearchInstitution`) aren't written by **any** loader in the repo, so the four fields weren't occasionally undercounting — they were hard zeros on every run since the label migration. The branch reads the labels the asset actually writes (`FinancialTransaction`, `Individual`, `Organization` at lines 690/700/710).

**The `Organization` checkpointing is the non-obvious part and it's correct.** Four separate steps write `Organization`, so a naive read would attribute the company step's updates to the institution step as well. I ran the branch asset against a stub client mimicking the real cumulative-metrics contract, with **two batches per step**: planted 40 Organization updates in the company step, 7 FinancialTransaction, 3 Individual, and 11 Organization in the institution step. Got exactly `companies_updated: 40, awards_updated: 7, researchers_updated: 3, institutions_updated: 11`. Disjoint deltas, no double-count across batch boundaries, no cross-step bleed.

**It's a label fix, not a relocated bug.** I checked whether `nodes_updated` is secretly a mislabeled `nodes_created` — the classic MERGE-counting trap. It isn't. The asset never reads `result.consume().counters`; `batch_upsert_nodes` (`client.py:328-341`) uses `created_count`/`updated_count` from `build_batch_merge_query(include_hash_check=True)`, where `updated_count = count(CASE WHEN needs_update THEN 1 END)` and `needs_update` excludes content-identical matches. A MERGE that matched with identical content lands in neither bucket, which is the right meaning for a field named `*_updated`.

**No consumer flips.** Repo-wide, the only reader of these four fields is `scripts/data/run_neo4j_sbir_load.py:68-74`, a display-only markdown table with no threshold. `neo4j_sbir_awards_load_check` reads only `status`, `errors`, `awards_loaded`, `total_rows_processed`. The `companies_updated` in `company_categorization.py:578-595` that *does* threshold belongs to a different asset with its own dict — no cross-talk. Persisted metrics JSON keeps its schema; old files hold zeros, new ones hold real values.

### Two should-fix items, both inline

1. **The `*_loaded` fields still count submitted rows**, sitting interleaved with the four you just made canonical — and one of them feeds a load check that can report success on an inflated number.
2. **The new test can't catch this PR's own regression class.** Reverting the label fix or deleting the checkpoint threading both leave all 3 tests green. I confirmed both.

### Smaller

- Agency and sub-agency steps (lines 796-813) write `Organization` after the last checkpoint, so their updates are correctly excluded from the four fields but still included in the nested `result["metrics"]["nodes_updated"]["Organization"]`. In my run both read 51 only because agencies contributed 0; with agency updates the two blocks in the same JSON disagree with nothing explaining why. Correct by design, undocumented.
- `companies_updated` now folds UEI/DUNS dedup merges into "updated" — `client.py:629-631` adds Phase-2 `merged_count` (counted per incoming row, so two duplicates collapsing count as 2, and the loser is then `DETACH DELETE`d at 653-660) on top of Phase-3 `updated`. Bounded by `companies_updated ≤ companies_loaded` so it can't go absurd, but worth a comment.

Branch is one commit on the current main tip, fast-forwards clean, all 16 checks green.

---

packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
```diff
@@ -955,6 +999,7 @@ def neo4j_sbir_awards_load_check(neo4j_sbir_awards: dict[str, Any]) -> AssetChec
     status = neo4j_sbir_awards.get("status")
     errors = neo4j_sbir_awards.get("errors", 0)
     awards_loaded = neo4j_sbir_awards.get("awards_loaded", 0)
+    awards_submitted = neo4j_sbir_awards.get("awards_submitted", awards_loaded)
```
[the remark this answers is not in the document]
hollomancer: Fixed in 27c2896. The docstring now says the gate is on `awards_submitted` and states why `awards_loaded` can't be the trigger — an idempotent re-run of unchanged data legitimately writes zero nodes and must still pass. `test_load_check_accepts_idempotent_no_change_run` already pins that behavior.

---

packages/sbir-analytics/sbir_analytics/assets/sbir_neo4j_loading.py
hollomancer: Resolved on the current head (554759e), and it took the rename option rather than the one-liner.

`awards_loaded` is now `awards_created + awards_updated` — nodes written, not rows submitted — with `awards_submitted` carrying the submitted count as its own field, and the same split applied to companies, researchers, and institutions. The load check consumes `awards_submitted` for the zero gate and reports `{loaded}/{submitted} awards written`, so the inflated-success case you traced (10 batches, one failing, `✓ ... 10000 awards`) can no longer arise.

The negative-delta case is closed too: `_count_since` returns `max(0, current - previous)`, covered by `test_counter_deltas_never_go_negative_after_metrics_rebind`. The metrics-rebind asymmetry you flagged for `relationships_created` and the nested blocks is still latent and still undocumented — worth the line in the PR body you suggested.

---

tests/unit/assets/test_sbir_neo4j_loading.py
hollomancer: Resolved — `test_asset_reports_canonical_stage_deltas` is the asset-level test you described. It drives the real asset with a stub client and asserts the four headline fields plus the nested `nodes_updated` block, so both regressions you demonstrated now fail it: reverting `_updated_since(metrics, "Researcher")` breaks `researchers_updated == 1`, and deleting the `Organization` checkpoint threading makes `institutions_updated` read 3 instead of 1.

---

PR #575, opened 2026-08-09
Retire the all-assets schedule

hollomancer:
## Summary

- remove the repository-wide `sbir_analytics_job` and daily schedule
- remove its obsolete server environment toggle
- retain explicitly named product jobs and the opt-in weekly core refresh
- add a regression test that the broad job and schedule remain absent

## Why

An `AssetSelection.all()` schedule automatically absorbs every newly added asset. That allows exploratory or resource-intensive work to enter automation without an explicit operational decision.

## Impact

No schedule runs more broadly as a replacement. Operators retain the named jobs and the stopped-by-default `weekly_core_refresh` schedule.

## Verification

- 19 schedule, server-default, and Dagster-definition tests passed
- Dagster definitions validate through the focused suite
- server Compose configuration validates
- Ruff check and format check
- `make docs-check`
- `git diff --check`


hollomancer (commented):
## Verified surgical — I diffed the resolved Dagster inventory on both sides rather than reading the diff

Exactly one job (`sbir_analytics_job`) and one schedule (`daily_sbir_analytics`) disappear. All 109 asset keys, all 5 sensors, and the other 10 schedules — names, crons, jobs, default statuses — are byte-identical before and after. No replacement schedule is added and no cadence changes, so there's no daily→weekly gap. `Definitions.validate_loadable(defs)` (the literal CI command) passes against branch content with `DAGSTER_LOAD_HEAVY_ASSETS` both unset and `false`.

Measured: jobs 23→22, schedules 11→10, asset keys 109→109, **schedules RUNNING by default `['daily_sbir_analytics']` → `[]`**.

On the live host the retired schedule was already gated STOPPED (`.env.server.example:70`, `docker-compose.server.yml:47`), so live scheduled coverage goes from zero to zero. The real behavioral delta is that a default `dagster dev` now has no RUNNING schedule at all — which is precisely what CLAUDE.md's "keep schedules disabled until their jobs have completed successfully by hand" asks for.

Also worth crediting: the two deleted tests were **replaced**, not dropped — `tests/unit/test_server_schedule_gating.py:70-74` asserts both the job and the schedule are absent. That's the right shape for a removal PR.

### Two should-fix items

**1. The runbook claims a property the code doesn't have.** `docs/deployment/self-hosted-server.md:271-272` now says "New assets cannot enter automation without being added to an explicitly named job." But `definitions.py:99-107` still defines `core_refresh_job` as `AssetSelection.all() - AssetSelection.assets(*_heavy_keys)` — a repository-wide absorber for every non-heavy asset.

Concretely: a contributor adds a non-heavy exploratory asset; an operator has enabled `WEEKLY_CORE_REFRESH_ENABLED=true` exactly as lines 273-275 of the same runbook instruct; the asset materializes every Sunday 03:00 UTC with no explicit scheduling decision. That's the failure the PR's "Why" says it eliminates. `docs/architecture/dagster-pipelines.md:40` on this same branch correctly describes `core_refresh_job` as selecting "all currently-loaded non-heavy assets," so the two docs now contradict each other.

Suggested: "the *daily repository-wide* schedule has been retired; `core_refresh_job` remains a non-heavy all-assets selection and stays STOPPED." (`dagster-reorganization-plan.md:269` already anticipates retiring it once explicit jobs exist.)

**2. 14 heavy asset keys now belong to no named job.** `job_definitions` drops from `[etl_job, core_refresh_job]` to `[core_refresh_job]`, and nothing else selects: `enriched_uspto_ai_patent_join`, `raw_uspto_ai_human_sample`, `raw_uspto_ai_predictions`, `validated_uspto_ai_cache_stats`, `sbir_fiscal_impacts`, `sbir_fiscal_impacts_by_sector`, `sbir_fiscal_impacts_by_state`, `ml/cet_award_training_dataset`, `ml/enriched_cet_patent_classifications`, `ml/raw_cet_human_sampling`, `ml/train_cet_patent_classifier`, `ml/transformed_cet_analytics`, `ml/transformed_cet_analytics_aggregates`, `ml/validated_cet_iaa_report`.

Keys covered by ≥1 named job: **109 → 95**. Not a scheduled-coverage loss (none were on a RUNNING schedule), but an operator wanting to rebuild `sbir_fiscal_impacts_by_state` or `ml/validated_cet_iaa_report` now has no job to launch and must hand-select in the UI or use `--select`. "Operators retain the named jobs" glosses over this. Either say so explicitly or fold these into the existing heavy jobs.

**3. Test name no longer matches its assertion.** `tests/unit/test_server_env_defaults.py:42-46` — `test_heavy_assets_loaded_but_not_scheduled` now asserts only `DAGSTER_LOAD_HEAVY_ASSETS == "true"`; the comment at 43-44 ("the always-on host must never launch one on its own") is unverified. Someone could add `DAILY_CET_FULL_PIPELINE_ENABLED=true` to the template and this test would still pass under a name that says it guards against exactly that. Rename it, or assert no `SBIR_ETL__DAGSTER__SCHEDULES__*_ENABLED` key in the template is `true`.

### One operator step worth adding

`default_status` sets only the *initial* instigator state — live state lives in the `dagster_home` volume. If `daily_sbir_analytics` was ever started by hand in the UI, deploying this silently ends those runs and leaves an orphaned instigator row. Given CLAUDE.md's live-deployment rules, one line in the runbook telling the operator to confirm the schedule's current state in the UI before deploying would close it. I can't inspect the live instance, so this is a suggestion rather than a finding.

### Nits

- `docs/architecture/dagster-pipelines.md:91` still says "the two core jobs" — there's one now, and the file was edited three lines above.
- `config/prod.yaml:201` has a dangling `etl_job: "0 2 * * *"` under `dagster.schedules`. Pre-existing key, but this PR is what makes it dangle. Nothing reads it.
- `dagster-reorganization-plan.md:268` is the disposition row this PR implements ("Retire; it is an unsafe repository-wide scheduling boundary") and `:47` still describes the job in the present tense. Nothing records that the row is now done.
- `_schedule_status(..., default_running=True)` has no production caller left — exercised only by `test_schedule_status_helper`.
- `.test_durations:3565-3566` still lists the two deleted tests. pytest-split tolerates it; Fast Tests are green.

Fast-forwards onto current main, all 16 checks green. One gap in the green signal: `Docker Build and Smoke Test` — the lane that runs `dagster definitions validate` — was **skipped** despite `docker-compose.server.yml` changing. I ran the equivalent locally, so it's covered in substance, but the path filter arguably should have caught it.

---

PR #576, opened 2026-08-09
Remove self-fulfilling pipeline tests

hollomancer:
## Summary

- remove an E2E test that created three files and immediately asserted they existed
- remove schema checks that skipped whenever developer-local outputs were absent
- remove a hard-coded filename-uniqueness assertion with no production behavior
- remove the now-unused temporary-output fixture

## Why

These tests increased the visible test count without exercising a production writer or a clean-checkout contract. The transition analytics writer and its JSON/checks payload remain covered by the integration chain, while real asset materializations remain in the functional suite.

## Impact

No production behavior changes. The suite now reports fewer but more meaningful tests.

## Verification

- transition E2E, functional pipeline, and analytics integration suites — 8 passed, 1 environment-dependent skip
- Ruff check and format check
- `git diff --check`

### DIFF
--- a/tests/e2e/transition/test_detection_smoke.py
+++ b/tests/e2e/transition/test_detection_smoke.py
@@ -2,8 +2,6 @@
 
 from __future__ import annotations
 
-import json
-
 import pandas as pd
 import pytest
 
@@ -58,21 +56,3 @@ def test_transition_pipeline_summary(
     assert "award_transition_rate" in summary
     assert "company_transition_rate" in summary
     assert "cet_area_transition_rates" in summary
-
-
-def test_pipeline_output_files(tmp_path):
-    """Ensure downstream tasks write analytics + report artifacts."""
-    output_dir = tmp_path / "data" / "processed"
-    output_dir.mkdir(parents=True, exist_ok=True)
-
-    analytics_json = output_dir / "transition_analytics.json"
-    summary_md = output_dir / "transition_analytics_executive_summary.md"
-    checks_json = output_dir / "transition_analytics.checks.json"
-
-    analytics_json.write_text(json.dumps({"score_threshold": 0.6}))
-    summary_md.write_text("# Executive Summary")
-    checks_json.write_text(json.dumps({"ok": True}))
-
-    for path in (analytics_json, summary_md, checks_json):
-        assert path.exists()
-        assert path.stat().st_size > 0
--- a/tests/functional/test_pipelines.py
+++ b/tests/functional/test_pipelines.py
@@ -7,23 +7,11 @@ Tests each major pipeline function end-to-end:
 - ModernBert embeddings
 """
 
-import pytest
-import pandas as pd
-from pathlib import Path
-
-
-@pytest.fixture
-def output_dir(tmp_path):
-    """Create temporary output directory."""
-    output = tmp_path / "data" / "processed"
-    output.mkdir(parents=True)
-    return output
-
 
 class TestTransitionPipeline:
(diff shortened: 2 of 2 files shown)


hollomancer (commented):
## Every deletion checked against the code it claimed to cover — nothing real becomes unpinned

Deleting tests is the change I trust least on sight, so I verified each removal individually rather than accepting "self-fulfilling" as a category.

**Two are textbook self-fulfilling.** `test_pipeline_output_files` writes three files with `write_text(...)` and then asserts `path.exists()` on the same three paths — it imports nothing from production and runs in 0.04s. `test_pipeline_outputs_dont_conflict` asserts uniqueness over a list of six string literals written in the test body; it can only fail if someone edits the test. Both are correct removals.

**The third is more interesting than the PR body says, and it's the reason I'd approve rather than hesitate.** `test_pipeline_output_schema` isn't literally self-fulfilling — but it's provably dead, and worse than dead. Three of its four parquet paths are written by **no code anywhere in the repo**: `cet_classifications.parquet` (production writes `cet_award_classifications.parquet`), `fiscal_returns.parquet` (no fiscal asset writes parquet at all), and `modernbert_award_patent_similarity.parquet` (which exists only as an unbuilt target in `specs/modernbert_analysis_layer/design.md:33`). All four params skip permanently.

And the one path that *is* real asserts a schema production has never emitted. `scoring.py:69-78` writes `award_id, contract_id, score, method, signals, computed_at`; the test demanded `transition_score` and `confidence`. `git grep '"transition_score"'` across `packages/`, `sbir_etl/`, and `scripts/` returns nothing.

I confirmed that empirically: dropping a correctly-shaped `transitions.parquet` into place makes the surviving param **fail** with `AssertionError: Missing column: transition_score`. So the test wasn't a safety net that happened to be idle — it was a latent false alarm that stayed green only because the artifact is never present. Removing it is strictly better than leaving it.

**What still covers the real thing.** The transition analytics writer and its payload are genuinely pinned by `tests/integration/test_transition_mvp_chain.py:229-251`, which runs the asset for real and validates both `transition_analytics.json` and `.checks.json` including `score_threshold`, `award_transition_rate`, `company_transition_rate`, and `counts.total_awards`. That suite also catches mutations none of the removed tests could — deleting the `write_json(checks_path, checks)` call at `scoring.py:98` turns it red while all six removed tests stay green.

**The hard gate is untouched.** `tests/unit/scripts/test_phase_iii_precision_backtest.py` has an identical blob SHA on both refs, and no transition-scoring test is modified. Per CLAUDE.md that's the check that mattered.

**Scope is clean:** tests-only, no production file touched, collected tests 13 → 7 in the two files, no coverage `fail_under` gate to trip (and the removed tests executed zero project code, so they contributed no production coverage anyway). Fast-forwards onto current main.

### Three nits, none blocking

- **The PR body undersells its own case.** It describes these as "schema checks that skipped whenever developer-local outputs were absent," which reads as though a developer could have run them. Three of the four could never run for anyone, and the fourth would have failed if it had. Worth tightening so a future reader doesn't think a real safety net was traded away.
- **The branch name says `replace-weak-pipeline-tests`, but nothing is replaced** — the sole `+1` line is a changed signature. The title and body correctly say "remove"; only the branch name misleads.
- Removing the `fiscal_returns` and `modernbert_*` params drops the last test-side reference to two designed-but-unbuilt outputs. That's fine — `specs/modernbert_analysis_layer/design.md` is the correct record for unbuilt work, and a permanently-skipping test is not a tracker. Noting it only so the loss is deliberate.

---

PR #577, opened 2026-08-09
Repair and materialize NSF private-capital Phase 1 gate

hollomancer:
## Description

Repairs and materializes the NSF Phase 1 private-capital review gate.

The raw-data runner previously produced an empty NSF cohort because SBIR.gov uses canonical columns such as `Agency`, while cohort filtering expected normalized lowercase names. The graduation metric also counted any Phase II award for a firm, including awards before Phase I or outside a defined follow-up window.

This PR:

- accepts canonical SBIR.gov agency/ALN column aliases;
- resolves firms through exact connected components across every UEI, DUNS, and versioned normalized-name alias, including mixed-identifier and transitive cases;
- defines Phase I to Phase II graduation as a Phase II award no earlier than Phase I and within an inclusive, configurable five-year horizon;
- matches M&A events against the complete alias set for each resolved company component;
- emits a deterministic, portable `run_manifest.json` with input and output hashes, sizes, row counts, schemas, parameters, metric availability, horizon sensitivity, identity coverage, and explicit `exploratory` / `citable: false` status;
- labels unavailable outcomes as not measured rather than zero;
- materializes the pinned NSF 2015-2019 review result: 672/1,502 firms, 44.7% (95% Wilson interval 42.2%-47.3%);
- applies no external private-capital comparator and makes no relative program-performance claim;
- updates the research inventory and spec ledger to reflect that Phase 2 is a scaffold, not a validated matched comparison.

The Phase 1 checklist gate remains open pending review of the firm-level, pooled SBIR/STTR estimand, the exact-alias identity assumptions, and whether the unavailable transition, survival, M&A, and patent channels must be populated.

Scope is limited to Phase 1. A reproducible Form D control universe, symmetric FPDS/PATLINK/M&A outcomes, multi-agency matched reporting, the FY M&A trend, and filer-vs-non-filer or crowd-in/crowd-out analysis remain separate gated follow-ups.

## Type of Change

- [x] Documentation update

## Versioning Impact

- [ ] No release impact (internal or unreleased work)
- [x] PATCH: backward-compatible fix or documentation correction

See [the versioning policy](../docs/steering/versioning.md). The release PR applies the selected increment across all four packages.

## Testing

- [x] All tests pass locally

Results:

- focused Phase 1 unit and integration suite: 64 passed;
- full unit suite: 5,905 passed, 7 skipped;
- targeted Ruff: passed;
- targeted MyPy: passed for 13 source files;
- all six architecture, tier, file-size, configuration, repository-hygiene, and study-manifest checkers: passed when run directly with the existing project environment;
- real-data runner: reproduced 672/1,502 against the pinned 219,500-row SBIR.gov snapshot and regenerated provenance for the current four-entry baseline registry.

The `make lint-boundaries` and `make docs-check` wrappers were also attempted, but `uv` stopped before executing their checkers because GitHub refused the direct `en_core_web_sm` wheel HTTP/2 stream. The underlying checkers all pass directly.

## Checklist

- [x] I have commented my code where the implementation is not self-explanatory
- [x] I have added tests that prove the fixes are effective

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
@@ -7,7 +7,8 @@ all numeric fields are ``None`` so the downstream report can still render.
 
 Metrics:
     - phase_i_to_ii_graduation: any company with a Phase I award in stratum
-      that has a Phase II award (any year) in the cohort.
+      that has a Phase II award in the cohort no earlier than that Phase I
+      and within the configured graduation horizon (default: 5 years).
     - phase_ii_to_federal_contract_transition: Phase II awards with at least
       one upstream transition score >= the configured threshold (consumes
       ``transformed_transition_scores`` produced by the existing detector;
@@ -28,6 +29,8 @@ from dataclasses import dataclass, field
 from typing import Any
 
 import pandas as pd
+
+from .cohort import _award_year
 
 
 WILSON_Z_95 = 1.959963984540054
@@ -107,6 +110,9 @@ class OutcomeMetricsCalculator:
     """Compute per-stratum SBIR cohort outcome rates with Wilson CIs.
 
     Args:
+        graduation_horizon_years: maximum inclusive number of calendar years
+            from a Phase I award to a qualifying Phase II award. Default 5
+            matches the published-baseline cohort window.
         transition_score_threshold: minimum upstream transition-score for a
             Phase II award to count as transitioned. Default 0.65 matches
             the ``likely`` confidence threshold in
@@ -115,6 +121,7 @@ class OutcomeMetricsCalculator:
             recipient/vendor activity. Default 5.
     """
 
+    graduation_horizon_years: int = 5
     transition_score_threshold: float = 0.65
     survival_horizon_years: int = 5
     z: float = WILSON_Z_95
@@ -146,15 +153,32 @@ class OutcomeMetricsCalculator:
         cohort["_company_key"] = cohort.apply(_company_key, axis=1)
         cohort["_name_key"] = cohort.apply(_normalized_name_key, axis=1)
 
-        # Phase I->II graduation: per-vintage
+        # Phase I->II graduation: per-vintage. A Phase II that predates its
+        # company's Phase I is not a graduation, nor is one outside the
+        # configured cohort window.
         phase_i = cohort[cohort["phase_label"] == "I"]
-        phase_ii_companies = {
-            ck for ck in cohort.loc[cohort["phase_label"] == "II", "_company_key"] if ck
-        }
+        phase_ii = cohort[cohort["phase_label"] == "II"].copy()
+        phase_ii_years: dict[str, set[int]] = {}
+        for _, row in phase_ii.iterrows():
+            company_key = row["_company_key"]
+            year = _award_year(row)
+            if company_key and year is not None:
+                phase_ii_years.setdefault(company_key, set()).add(year)
+
         for vintage, group in phase_i.groupby("vintage_bucket", dropna=False):
             keys = [k for k in group["_company_key"].tolist() if k]
             unique_keys = set(keys)
-            graduated = unique_keys & phase_ii_companies
+            graduated: set[str] = set()
+            for _, row in group.iterrows():
+                company_key = row["_company_key"]
+                phase_i_year = _award_year(row)
+                if not company_key or phase_i_year is None:
+                    continue
+                if any(
+                    phase_i_year <= phase_ii_year <= phase_i_year + self.graduation_horizon_years
+                    for phase_ii_year in phase_ii_years.get(company_key, set())
+                ):
+                    graduated.add(company_key)
             records.append(
                 self._make_row(
                     vintage,
@@ -167,7 +191,6 @@ class OutcomeMetricsCalculator:
             )
 
         # Phase II -> federal-contract transition
-        phase_ii = cohort[cohort["phase_label"] == "II"].copy()
         transition_award_ids = self._transitioned_award_ids()
         for vintage, group in phase_ii.groupby("vintage_bucket", dropna=False):
             denom = len(group)
--- a/packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
@@ -6,9 +6,11 @@ is missing the corresponding metric is reported as ``available=False`` and
 all numeric fields are ``None`` so the downstream report can still render.
 
 Metrics:
-    - phase_i_to_ii_graduation: any company with a Phase I award in stratum
-      that has a Phase II award in the cohort no earlier than that Phase I
-      and within the configured graduation horizon (default: 5 years).
+    - phase_i_to_ii_graduation: any canonically resolved company with a Phase I
+      award in stratum that has a Phase II award in the cohort no earlier than
+      that Phase I and within the configured graduation horizon (default:
+      5 years). Every UEI, DUNS, and normalized-name alias present on a row is
+      connected before this firm-level calculation.
     - phase_ii_to_federal_contract_transition: Phase II awards with at least
       one upstream transition score >= the configured threshold (consumes
       ``transformed_transition_scores`` produced by the existing detector;
@@ -18,8 +20,7 @@ Metrics:
       vendor in any federal dataset >=5 years after Phase II award year.
       Denominator is unique companies (not award rows) per stratum.
     - ma_exit_rate: SBIR awardee appears in the M&A events JSONL (post-#286).
-      Join is UEI/DUNS-first; falls back to normalized company name when
-      UEI/DUNS are absent from the cohort row.
+      The join checks every alias attached to the resolved company component.
 """
 
 from __future__ import annotations
@@ -30,7 +31,7 @@ from typing import Any
 
 import pandas as pd
 
-from .cohort import _award_year
+from sbir_etl.identity import CompanyNameProfile, normalize_company_name
 
 
 WILSON_Z_95 = 1.959963984540054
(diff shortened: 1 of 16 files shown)


packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
```diff
@@ -146,15 +153,32 @@ def compute(self, cohort: pd.DataFrame) -> pd.DataFrame:
...
+        # company's Phase I is not a graduation, nor is one outside the
+        # configured cohort window.
         phase_i = cohort[cohort["phase_label"] == "I"]
-        phase_ii_companies = {
-            ck for ck in cohort.loc[cohort["phase_label"] == "II", "_company_key"] if ck
-        }
+        phase_ii = cohort[cohort["phase_label"] == "II"].copy()
+        phase_ii_years: dict[str, set[int]] = {}
+        for _, row in phase_ii.iterrows():
+            company_key = row["_company_key"]
```
hollomancer, line 325: `_company_key` chooses exactly one identifier per row (UEI first, then DUNS). If a Phase I row has only DUNS and a later Phase II row has the same DUNS plus a UEI, these become `duns:...` and `uei:...`, so this index never links them even though the records share DUNS. That can materially understate the numerator in the committed report. Please resolve/crosswalk all available identifiers before computing the horizon, and add a regression test for this mixed-identifier case.

hollomancer (commented):
The five-year horizon repair is directionally correct, but the published graduation metric still has an entity-resolution hole that can undercount real graduations. Please canonicalize a company across every identifier present on its rows (or build an alias/crosswalk) and regenerate the committed report before treating the result as the Phase 1 baseline.

hollomancer (commented):
## The gate was tightened, not weakened — verified by execution. The labeling is the strongest part; the estimand is the weak one.

I went in expecting the classic failure mode for a PR titled "repair and materialize a gate that was failing": criteria quietly relaxed until it passes. **The opposite happened.** The old predicate was `graduated = any company with any Phase II`. The new one (`outcomes.py:171-181`) additionally requires `phase_i_year <= phase_ii_year <= phase_i_year + 5` — a strict subset, so the new rate is necessarily ≤ the old. Confirmed by running the boundaries: prior-Phase-II (2014 → Phase I 2016) now scores 0/1; a 6-year gap scores 0/1 at horizon 5 and 1/1 at horizon 6; inclusive endpoints (2015 → 2020) score 1/1; same-year scores 1/1.

The repair is real too. On `main`, `_is_agency_row` read only `row.get("agency")` while the SBIR.gov bulk column is `Agency` — visible in the committed 42-column schema, which has no lowercase alias. **All three new tests fail on `origin/main`** (I ran them there): two on the old counting behavior, one with `TypeError: OutcomeMetricsCalculator.__init__() got an unexpected keyword argument 'graduation_horizon_years'`. Not tautological.

**On committing results:** yes, it commits `672 / 1,502 = 44.7%` (Wilson 42.2–47.3%, +11.7pp), and the labeling is genuinely good — `citable: false` / `exploratory` in four places, the question marked *Partial* rather than answerable, and the package already declared `EPISTEMIC_TIER = "exploratory"` on main so nothing was silently promoted. The arithmetic is internally exact; I recomputed with the repo's own `wilson_interval(672, 1502)` and got `0.447403 / 0.422423 / 0.472652`. I could not verify the counts themselves — the 394 MB snapshot isn't on this host.

### What I'd fix before this comes out of draft

**1. The Dagster asset silently adopts the new estimand with no knob and no record.** `asset.py:87-95` constructs `OutcomeMetricsCalculator(...)` without `graduation_horizon_years`, so it inherits the new default of 5. `AgencyPrivateCapitalConfig` gains no field, the metadata dict at `:112` records `headline_vintage` but not the horizon, and the docstring still describes the old semantics. Someone materializes the asset, gets a different graduation rate than the pre-merge asset produced, and the emitted comparison JSON contains nothing explaining why — with no way to reproduce the old number without editing source.

**2. Reproducibility stops one field short of where the manifest is aiming.** The input SHA-256, byte size, row count, and full schema are pinned, and I confirmed the committed manifest round-trips **byte-identically** through the writer's exact serialization — so it's a real emission, not hand-written. But there's **no fetch date and no source URL** for a rolling bulk export, so SHA `73d646fc…` can never be re-satisfied by a third party; and there's **no output hash** — `672/1,502` appears nowhere machine-readable, so a transcription error between the run and the prose would be undetectable. The manifest imitates the evidence-tier machinery (which requires input hashes *and* output hash) and omits exactly the field that binds the number to the inputs. Adding `run_date` and a small results block would close both.

**3. No horizon sensitivity is reported.** The horizon is the single largest lever on the headline, and the artifact exists to be reviewed. A four-row table (2 / 3 / 5 / unbounded) costs one flag and is exactly what a reviewer needs to accept or revise the estimand — which the report's own "Gate decision" item 2 asks them to do.

**4. The 5-year cutoff is anchored to a field the PR itself calls untrustworthy.** It appears nowhere in `requirements.md` or `design.md`. The docstring says it "matches the published-baseline cohort window," which traces to `published_baselines.yaml:26` — but that same entry's notes say "30-35% range **depending on cohort window**", and this PR declares that baseline provisional pending primary-source verification. Mitigating, and better than the repo norm: the value is in the manifest `parameters`, printed in the report banner, and CLI-configurable.

**5. The tier is demoted `pipelines` → `exploratory` in the same commit that declares the deliverable materialized.** `requirements.md:3`. Defensible on the merits — the package already declared exploratory on main, so the spec was aspirational — but it lowers the contract from "deterministic, reproducible from a declared data cut" to "labeled non-citable," which is precisely what findings 2 and 3 are about, inside the PR claiming to satisfy the gate. Worth being an explicit decision rather than a doc edit. Note `check_epistemic_tiers.py` only validates the string is one of four names, so "epistemic-tier check passed" is true but structurally blind to a demotion.

**6. A single unparseable `Award Year` now hard-crashes the run.** `cohort.py:177-180` — mixed-parseable years give `float64` with `NaN`, `NaN is not None` is `True`, and `vintage_bucket()` raises. Pre-existing code, but **newly reachable**: before this PR the raw-CSV cohort was empty and `build()` returned at the `if cohort.empty` guard. Reproduced with two rows, one `"n/a"` and one `"2016"` → `ValueError: year is required`. All-unparseable stays `object` dtype and passes, which is why the pinned snapshot survived. The next SBIR.gov refresh with one malformed NSF year kills the runner.

**7. Mixed UEI/DUNS keys systematically undercount graduation.** `_company_key` (`outcomes.py:76-96`) is UEI → DUNS → lowercased name, first-match-wins per row. A firm whose Phase I row carries only DUNS and whose Phase II row carries only UEI scores **0/1** — I ran it. SBIR.gov backfilled UEI unevenly across 2015–2019, so this isn't hypothetical. Line 46 discloses the cascade and 51 hedges vaguely, but neither names the same-firm/different-key-type case, which is the dominant mode. Directly connected: `requirements.md:130-131` (Requirement 6) **SHALL** report match rates and entity-resolution coverage as sensitivity metadata, and the artifact reports neither — a uei/duns/name breakdown of the 1,502 keys is exactly the diagnostic that would size this.

### Nits

- `specs/status.md:33` flips "Gated backlog" → "Active", but the registry's own vocabulary says Active means "implementation can proceed," while the entry and PR body both say Phase 2 is gated on human sign-off. "Gated backlog" reads more accurately.
- `tasks.md:67` edits an acceptance criterion `2015–2020` → `2015–2019` in the PR claiming to satisfy it. Defensible (5-year buckets can't produce that label, and the default predates this PR), but it's a criterion reshaped to match the output.
- SBIR and STTR are pooled into one "NSF Phase I→II graduation rate" — I confirmed both `Program` values survive the filter and a cross-program STTR→SBIR counts as graduation. The estimand section should say so.
- Graduation is firm-level, not project-level, so a Phase II deriving from an out-of-vintage Phase I can mark an in-vintage Phase I as graduated. Honestly declared in the Numerator bullet, but it means the figure isn't comparable to NSF's own project-level conversion statistic — worth a sentence now that it's in the canonical inventory.
- `asset.py:100-110` — the Dagster path still writes `agency_vs_published_baselines.md` with no banner and no `citable: false` in its Output metadata. `_annotate_report()` was added only to the archived script.
- "Labels unavailable outcomes as not measured rather than zero" is overstated for the parquet: `_make_row` still writes `numerator=0` alongside `available=False`, so a consumer ignoring `available` reads zeros. Pre-existing.
- `outcomes.py:34` imports the private `_award_year` across modules and re-derives per row a year `build()` already computed.
- **PR body checklist: "Integration tests added/updated [x]" is false.** No file under `tests/integration/` is touched, and `test_phase1_pipeline.py` uses lowercase normalized columns — so it doesn't exercise the repaired raw-casing path at all. The only end-to-end coverage of the repair is the new unit test.

Merges clean, all 16 checks green. One clarification on the title: "materialize" here means a local script run against a local CSV, not a Dagster materialization on the live host — no workflow, Docker, schedule, or job-wiring file is touched, so CLAUDE.md's live-deployment rules aren't engaged.

Also confirming the test-count claim: I get 5,826 passed / 4 skipped / **5 failed** on `tests/unit/`, but **all 5 reproduce on `origin/main`** (playwright binaries, plotly, HTML validator) — environment, not this branch. No regressions.

---

packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
```diff
@@ -227,9 +407,8 @@ def compute(self, cohort: pd.DataFrame) -> pd.DataFrame:
...
 
         # M&A exit rate (per stratum, all phases).
-        # Bug fix: join on UEI/DUNS key first; also check name key as fallback so
-        # that sbir_ma_events.jsonl (keyed by company_name) matches cohort rows
-        # that have a UEI/DUNS in addition to rows that only have a name.
+        # Match every alias on a resolved company component. The current M&A
+        # artifact is name-keyed, while future inputs may carry UEI or DUNS.
         for (vintage, phase), group in cohort.groupby(
             ["vintage_bucket", "phase_label"], dropna=False
         ):
```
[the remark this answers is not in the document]
hollomancer: Confirmed and fixed in 789c230. The comment directly above the loop already claimed a component-wide match; the loop underneath did a per-row one.

The aliases are now unioned onto each resolved component once, cohort-wide — identity belongs to the company, not to the row it appeared on — and the exit set is `aliases_by_company[key] & ma_event_companies`. Cohort-wide rather than per-stratum is deliberate: `_company_key` is resolved across the whole cohort, so the alias set has to be too, or a stratum containing only the UEI-bearing row would still miss the name.

Regression test `test_ma_exit_rate_matches_aliases_from_elsewhere_in_the_component` covers exactly your case — Phase II row with UEI only, name reaching the component through the Phase I row — and fails on the per-row join.

---

packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/outcomes.py
hollomancer: Resolved on the current head. `_company_key` no longer picks one identifier per row: `_resolve_company_identities` builds an exact alias graph over `uei:` / `duns:` / `name:` aliases and union-finds it into connected components, so a DUNS-only Phase I row and a DUNS+UEI Phase II row land in the same component and the graduation counts.

Two tests cover it — `test_phase_i_to_ii_graduation_crosswalks_mixed_uei_and_duns_rows` for your exact case, and `test_phase_i_to_ii_graduation_crosswalk_is_transitive_across_aliases` for the chained case. The component-basis breakdown you asked for in Requirement 6 is emitted too (`company_basis_counts`, `uei_duns_bridge_company_count`, `resolved_row_rate`).

The related per-row/per-component gap in the M&A join is fixed separately in 789c230.

---

PR #579, opened 2026-08-09
Make CET Neo4j loading failures visible

hollomancer:
## Summary

- fail CET Neo4j materializations on missing loaders, missing/malformed/empty artifacts, connection or loader failures, summary persistence failures, loader errors, and incomplete submitted-vs-processed counts
- call the production `CETLoader` APIs and persist a failure summary before raising
- close the Neo4j client on every path and share one repository-consistent `SKIP_NEO4J_LOADING` predicate between the gate and client factory
- carry a normalized `company_uei` through company profile generation, adapt it to loader field `uei`, and MATCH existing `Organization` nodes on `uei`
- keep legacy `company_id` only as the company aggregator compatibility alias; arbitrary company IDs and DUNS-only records are never relabeled as UEIs
- rebase onto main at `57514c9a` after #578 merged, preserving its fail-fast CET materialization and canonical artifact-path contracts

## Identity contract

The company profile artifact now names its graph identity explicitly as `company_uei`. The producer accepts only semantic UEI columns and validates/normalizes them; it does not infer UEI from `company_id` or `duns`. Both company enrichment and company-to-CET relationship assets map `company_uei` to `uei` and invoke the loader with `key_property="uei"`.

DUNS-only organizations remain outside this CET load until a typed DUNS-to-`organization_id` resolution path exists.

## Verification

- focused combined CET profile/loading/loader contracts: 144 passed
- full unit suite: 5,893 passed, 7 skipped
- Ruff lint and format checks across `sbir_etl`, all packages, and tests
- mypy: 305 source files
- architecture, epistemic-tier, file-size, config, repository-hygiene, study-manifest, and identity-boundary guards
- `git diff --check`

The focused Dagster tests still emit intermittent SQLAlchemy closed-database cleanup noise after successful execution; the suite exits successfully.

### DIFF
--- a/packages/sbir-analytics/sbir_analytics/assets/cet/loading.py
+++ b/packages/sbir-analytics/sbir_analytics/assets/cet/loading.py
@@ -11,6 +11,7 @@ This module contains:
 from __future__ import annotations
 
 import json
+import os
 from pathlib import Path
 from typing import Any
 
@@ -35,6 +36,82 @@ except Exception:
     CETLoaderConfig = None
 
 
+def _skip_requested(context, operation: str) -> bool:
+    if os.getenv("SKIP_NEO4J_LOADING", "false").lower() not in {"true", "1", "yes"}:
+        return False
+    context.log.warning(f"Skipping {operation}: SKIP_NEO4J_LOADING is enabled")
+    return True
+
+
+def _require_loader() -> None:
+    if CETLoader is None or CETLoaderConfig is None:
+        raise RuntimeError("CET Neo4j loader dependencies are unavailable")
+
+
+def _connected_client():
+    client = _get_neo4j_client()
+    if client is None:
+        raise RuntimeError("Neo4j loading was not skipped, but no client was created")
+    return client
+
+
+def _close_client(client, context) -> None:
+    try:
+        client.close()
+    except Exception as exc:  # pragma: no cover - defensive cleanup
+        context.log.warning(f"Failed to close Neo4j client cleanly: {exc}")
+
+
+def _ensure_successful_metrics(metrics, operation: str) -> None:
+    errors = int(getattr(metrics, "errors", 0) or 0)
+    if errors:
+        raise RuntimeError(f"{operation} completed with {errors} loader error(s)")
+
+
+def _write_summary(filename: str, result: dict[str, Any]) -> None:
+    out_path = DEFAULT_OUTPUT_DIR / filename
+    out_path.parent.mkdir(parents=True, exist_ok=True)
+    with out_path.open("w", encoding="utf-8") as fh:
+        json.dump(result, fh, indent=2)
+
+
+def _award_enrichments(classifications: list[dict]) -> list[dict]:
+    enrichments = []
+    for row in classifications:
+        supporting = row.get("supporting_cets") or []
+        enrichments.append(
+            {
+                "award_id": row["award_id"],
+                "cet_primary_id": row.get("primary_cet"),
+                "cet_primary_score": row.get("primary_score"),
+                "cet_supporting_ids": [
+                    item.get("cet_id") for item in supporting if isinstance(item, dict)
+                ],
+                "cet_taxonomy_version": row.get("taxonomy_version"),
+                "cet_classified_at": row.get("classified_at"),
+                "cet_model_version": row.get("model_version"),
+            }
+        )
+    return enrichments
+
+
+def _company_enrichments(profiles: list[dict]) -> list[dict]:
+    enrichments = []
+    for row in profiles:
+        cet_scores = row.get("cet_scores") or {}
+        enrichments.append(
+            {
+                "company_id": row["company_id"],
+                "cet_dominant_id": row.get("dominant_cet"),
+                "cet_dominant_score": row.get("dominant_score"),
+                "cet_specialization_score": row.get("specialization_score"),
+                "cet_areas": list(cet_scores) if isinstance(cet_scores, dict) else [],
+                "cet_taxonomy_version": row.get("taxonomy_version"),
+            }
+        )
+    return enrichments
+
+
 @asset(
     name="loaded_cet_areas",
     description="Load CETArea nodes into Neo4j from CET taxonomy artifact.",
@@ -50,13 +127,9 @@ except Exception:
 )
 def loaded_cet_areas(context, cet_taxonomy) -> dict[str, Any]:
     """Upsert CETArea nodes based on taxonomy output."""
-    if CETLoader is None or CETLoaderConfig is None:
-        context.log.warning("CETLoader unavailable; skipping CETArea loading")
-        return {"status": "skipped", "reason": "loader_unavailable"}
-
-    client = _get_neo4j_client()
-    if client is None:
-        return {"status": "skipped", "reason": "neo4j_skipped"}
+    if _skip_requested(context, "CETArea loading"):
+        return {"status": "skipped", "reason": "explicit_skip"}
+    _require_loader()
 
     # Config
     taxonomy_parquet = Path(
@@ -68,10 +141,11 @@ def loaded_cet_areas(context, cet_taxonomy) -> dict[str, Any]:
     batch_size = int(context.op_config.get("batch_size", 1000))
 
     # Read taxonomy (expect: cet_id, name, definition, keywords, taxonomy_version)
-    expected_cols = ("cet_id", "name", "definition", "keywords", "taxonomy_version")
+    expected_cols = ("cet_id", "name", "taxonomy_version")
     areas = _read_parquet_or_ndjson(taxonomy_parquet, taxonomy_json, expected_columns=expected_cols)
     context.log.info(f"Loaded CET taxonomy records for Neo4j: {len(areas)}")
 
+    client = _connected_client()
     try:
         loader = CETLoader(client, CETLoaderConfig(batch_size=batch_size))
         if create_constraints:
@@ -80,33 +154,20 @@ def loaded_cet_areas(context, cet_taxonomy) -> dict[str, Any]:
             loader.create_indexes()
 
         metrics = loader.load_cet_areas(areas)
+        _ensure_successful_metrics(metrics, "CETArea loading")
         result = {
             "status": "success",
             "areas": len(areas),
             "metrics": _serialize_metrics(metrics),
         }
 
-        # Persist a small run summary
-        out_path = DEFAULT_OUTPUT_DIR / "neo4j_cetarea_nodes.checks.json"
-        out_path.parent.mkdir(parents=True, exist_ok=True)
-        try:
-            with out_path.open("w", encoding="utf-8") as fh:
-                json.dump(result, fh, indent=2)
-        except Exception:
-            pass
-
-        try:
-            client.close()
-        except Exception:
-            pass
+        _write_summary("neo4j_cetarea_nodes.checks.json", result)
         return result
     except Exception as exc:
         context.log.exception(f"CETArea loading failed: {exc}")
-        try:
-            client.close()
-        except Exception:
-            pass
-        return {"status": "error", "error": str(exc)}
+        raise
+    finally:
+        _close_client(client, context)
 
 
 @asset(
@@ -129,13 +190,9 @@ def loaded_award_cet_enrichment(
     context, enriched_cet_award_classifications, loaded_cet_areas
 ) -> dict[str, Any]:
     """Upsert CET enrichment properties onto Award nodes."""
-    if CETLoader is None or CETLoaderConfig is None:
-        context.log.warning("CETLoader unavailable; skipping Award CET enrichment")
-        return {"status": "skipped", "reason": "loader_unavailable"}
-
-    client = _get_neo4j_client()
-    if client is None:
-        return {"status": "skipped", "reason": "neo4j_skipped"}
+    if _skip_requested(context, "Award CET enrichment"):
+        return {"status": "skipped", "reason": "explicit_skip"}
+    _require_loader()
 
     # Config
     award_class_parquet = Path(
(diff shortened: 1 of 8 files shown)


packages/sbir-analytics/sbir_analytics/assets/cet/utils.py
```diff
@@ -159,24 +159,37 @@ def text(s: Any) -> Any:
...
+                if not isinstance(record, dict):
+                    raise ValueError(f"Expected an object in {json_path} at line {line_number}")
+                missing = set(expected_columns) - set(record)
+                if missing:
+                    names = ", ".join(sorted(missing))
+                    raise ValueError(
+                        f"{json_path} line {line_number} is missing required fields: {names}"
+                    )
+                records.append(record)
+        return records
```
hollomancer, line 214: A blank NDJSON file returns `[]`; the callers then invoke the loader with zero records and write `{status: "success"}`. Zero-row Parquet behaves the same way. That preserves the most important silent-success case this PR is meant to remove. Please raise for an empty decoded artifact unless the caller explicitly opts into an allow-empty policy, and test both formats.

hollomancer (commented):
Malformed and missing inputs now fail honestly, but empty artifacts still produce green zero-record Neo4j loads. Please reject unexpected zero-row artifacts (or require an explicit, documented allow-empty policy) and cover both empty Parquet and blank NDJSON.

packages/sbir-analytics/sbir_analytics/assets/cet/loading.py
```diff
@@ -35,6 +36,82 @@
...
+    try:
+        client.close()
+    except Exception as exc:  # pragma: no cover - defensive cleanup
+        context.log.warning(f"Failed to close Neo4j client cleanly: {exc}")
+
+
+def _ensure_successful_metrics(metrics, operation: str) -> None:
+    errors = int(getattr(metrics, "errors", 0) or 0)
+    if errors:
+        raise RuntimeError(f"{operation} completed with {errors} loader error(s)")
```
hollomancer: **A load that writes zero rows still reports success**, because `metrics.errors` doesn't count the failure mode that actually happens.

`batch_set_existing_node_properties` (`client.py:361-451`) treats a row whose key matched no node as a `logger.warning`, not an error. `batch_create_relationships` (`client.py:773-873`) just returns `created_count == 0` when the `MATCH` finds nothing. Neither increments `errors`.

Reproduced with a fake session: 6 company rows, zero matched → `nodes_updated={'Organization': 0}`, `errors=0`, gate doesn't fire, asset returns `{"status": "success", "companies": 6}` having written nothing. That's the same invisible failure this PR exists to eliminate, just relocated from "swallowed exception" to "silent zero."

This matters more because of the key change at line 274-276. Moving Organization from the loader's default `uei` to `key_property="company_id"` is a substantive decision the PR body describes only as "including the upstream `company_id` key." `Organization.company_id` does exist (`sbir_neo4j_loading.py:344`) but holds the *canonical* id from `_resolve_canonical_company` (`:378-410`) — a bare UEI **or** `DUNS:<duns>` **or** `NAME:<normalized>` — while migration `007_unify_company_into_organization.py:145` explicitly `REMOVE o.company_id` on re-homed nodes, and the loader's own docstring (`cet.py:228`) names `organization_id` as authoritative with `uei` the house convention for enrichment. The artifact side is whatever `ColumnFinder.find_id_column` picked, most likely a raw `company_uei`.

So UEI-bearing firms match, DUNS- and name-keyed firms silently never match, and the gate above reports success. The gate needs a matched-vs-submitted ratio, not just an error count.

Two smaller things on the same gate:

**The message throws away the only record of how far the load got.** `f"{operation} completed with {errors} loader error(s)"` is a bare count, and the `raise` discards `metrics`. I reproduced a partial write — 6 rows, batch 2 of 3 throws — leaving rows 0,1,4,5 committed and 2,3 dropped. The operator gets an error count, no row identifiers, no matched/submitted numbers. Worse, because the raise happens before `_write_summary` at line 164, the on-disk `neo4j_*.checks.json` still holds the **previous** run's `"status": "success"`. Log the metrics before raising.

**Three swallows survive inside the path this PR hardens:** `base.py:301-302` turns any `create_constraints`/`create_indexes` failure into `logger.warning` (and `loading.py:151-154` calls both with defaults `True`, so a privilege error on the live DB is invisible); `cet.py:445` and `cet.py:383` do `except Exception: continue` on malformed supporting-CET and evidence entries with **no** `errors` increment, so they're invisible even to this gate.

---

hollomancer (commented):
## The corrected loader calls can never run — the connection factory these assets use is unconditionally broken

The diagnosis behind this PR is right, and worse than stated: those four `load_*_cet_*` methods genuinely don't exist on `CETLoader` (verified by introspection, and there's no `__getattr__`), so on `main` four of five CET load assets could only ever throw `AttributeError` into a broad `except`, report `{"status": "error"}`, and still let Dagster record a successful materialization. Fixing the API names, the artifact→loader schema mapping, the `finally` close, and the strict artifact reader is all correct work. Ruff is clean and the 41 tests pass — I reproduced both.

**But none of it executes.** `_get_neo4j_client()` in `cet/company.py:387-392` — the sole connection factory for all five assets — constructs `Neo4jConfig(uri=..., user=..., password=..., database=...)`. The field is `username`:

```
$ python -c "Neo4jConfig(uri='bolt://localhost:7687', user='neo4j', password='secret', database='neo4j')"
ValidationError: 1 validation error for Neo4jConfig
  username: Field required [type=missing, ...]
```

It raises before any network I/O, 100% of the time. Every other call site in the repo passes `username=` — `sbir_neo4j_loading.py:78`, `company_categorization.py:480`, `transition/utils.py:228`, `uspto/utils.py:528`, `sec_edgar_enrichment.py:287`. This one is the outlier.

So with `SKIP_NEO4J_LOADING` unset, all five assets die at `_connected_client()` with `RuntimeError: Neo4j connection failed but Neo4j loading not skipped: 1 validation error for Neo4jConfig...`, and every correction in this diff is dead code on a live host. It's a one-word fix and it belongs here, because this PR is the one claiming these assets now fail visibly rather than silently — they do, just for the wrong reason and before reaching any of the fixed code.

**The reason CI doesn't catch it is structural, not incidental.** Every asset-level test in the new file mocks `_connected_client` away, so no test ever calls `_get_neo4j_client`. Worth adding one that does.

Related, on test shape: the exception *types* are pinned well (`pytest.raises(RuntimeError, match="loader dependencies")`, `match="database rejected"`, `OSError match="disk full"`), but `test_nonzero_loader_errors_fail_materialization` calls the private `loading._ensure_successful_metrics(SimpleNamespace(errors=2), …)` directly against a hand-built object. There's no asset-level test proving a loader returning `errors > 0` actually fails the materialization, and nothing at all covers the `key_property="company_id"` path end-to-end.

### Partial-write state

Not transactional across the load, and the operator isn't told how far it got. Each batch is its own transaction, every batch method catches, increments `errors`, and **continues**; the gate fires only after the loop. Recovery is genuinely safe — everything is MERGE or MATCH-and-SET, so a rerun converges — but nothing in the failure output says so, and the stale `checks.json` actively misleads. Given CLAUDE.md treats materialization as a live-data mutation, this is the weakest part of the PR. Detail inline.

### Credentials — checked, no practical leak, but the protection is accidental

`company.py:403` interpolates the raw exception into the error string, and because of the bug above that exception is always a pydantic `ValidationError` whose repr embeds the full input dict **including `password`**. In practice pydantic's length-based truncation elides it — I tested `pw`, `hunter2`, and `S3cret!` with a realistic URI and database, and the password was cut in every case; only a contrived minimal config prints it. So: not leaking today, but the safeguard is field-blind repr truncation, not design.

This PR does raise the exposure at the margin, since it now re-raises from the asset (`loading.py:166-168`), pushing that string into Dagster's persisted event log and the Dagit UI rather than a single log line. Fixing the `user=`/`username=` bug removes the trigger outright; `password: SecretStr` on `Neo4jConfig` would make it structural. Note the CI security job runs bandit + detect-secrets — neither catches a runtime error-message leak.

### Interaction with #578 — complementary, merge both

No file overlap and no conflict (`git merge-tree` exits 0): #578 touches `classifications.py`/`company.py`/`training.py`, this touches `loading.py`/`utils.py`. One positive interaction worth knowing: on `main`, `transformed_cet_company_profiles` `touch()`es a 0-byte `.parquet` placeholder, and **this PR alone** turns that into a raw pyarrow error via the stricter reader at `utils.py:163`. #578 removes the `touch()` and returns the `.json` path instead, which resolves it cleanly. Merging both is better than either alone.

### Smaller

- `loading.py:32-36` still swallows the import into `CETLoader = None`, and `_require_loader` raises with no `from` cause — an operator learns the loader is unavailable but not why. Capture the exception in a module global and chain it.
- House exception types aren't used: `sbir_etl/exceptions.py` has `DependencyError` (exact fit for `_require_loader`), `FileSystemError`, and `ValidationError`, and sibling asset modules use them (`phase_iii_census/assets.py:24`, `uspto/utils.py:22`, `transition/contracts.py:19`).
- `company.py:407-430` still carries a dead near-duplicate `_read_parquet_or_ndjson` with the old `except Exception: pass; return []` body, next to the hardened `utils.py` copy. Uncalled, but a divergence hazard.
- No `RetryPolicy` exists anywhere in `packages/` — a transient blip mid-load now hard-fails the asset where it previously degraded to a false success. Since every write is MERGE/MATCH-and-SET, a Dagster `RetryPolicy` on these five is safe and cheap.
- `utils.py:183-188` validates required fields per-record on the NDJSON path but only per-column on the parquet path, so one malformed line in a 100k-line artifact hard-fails the whole load while a malformed parquet cell doesn't. Defensible; just note the asymmetry.

Merges clean into current main.

---

hollomancer:
Fixed the blocker in **#581** (`19430e10`) rather than asking you to carry it — `_get_neo4j_client` lives in `cet/company.py`, which isn't in this PR's diff, and the bug predates this branch.

`user=` → `username=`, plus the contract test that would have caught it. Nothing previously called the real factory (every test reference replaces it with a fake), which is why a 100%-reproducible failure stayed green in CI. Both new tests fail when the keyword is reverted; verified.

Once #581 lands, rebase and the corrections in this PR become reachable for the first time. **Everything else in my review still stands** — in particular the `metrics.errors` gate, which reports success on a load that matched zero rows, and the `key_property="company_id"` change, which is where that gap bites hardest.

I did not touch `loading.py` or `utils.py`, so there's no overlap with your diff.

---

packages/sbir-analytics/sbir_analytics/assets/cet/loading.py
```diff
@@ -35,6 +36,131 @@
     CETLoaderConfig = None
 
 
+def _skip_requested(context, operation: str) -> bool:
+    if os.getenv("SKIP_NEO4J_LOADING", "false").lower() not in {"true", "1", "yes"}:
+        return False
+    context.log.warning(f"Skipping {operation}: SKIP_NEO4J_LOADING is enabled")
+    return True
```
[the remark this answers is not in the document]
hollomancer: Addressed in 0e49c5c, but not by tightening the gate — that would have introduced a worse bug.

`_skip_requested` has to agree with the client factory it guards. `company._get_neo4j_client` accepts `true`, `1`, and `yes`, so restricting the gate to `true` alone would make `SKIP_NEO4J_LOADING=1` mean "do not skip" at the gate and "skip" at the factory: `_connected_client` would then raise `Neo4j loading was not skipped, but no client was created` instead of skipping. The truthy set is also the repo-wide convention — `company_categorization.py:461`, `transition/utils.py:214`, `uspto/utils.py:516`, `sec_edgar_enrichment.py:269`, and `sbir_neo4j_loading.py:72` all use it.

So the real defect was the duplication, not the values. Both now call `utils.neo4j_skip_requested()`, with the accepted set documented at `SKIP_NEO4J_VALUES`, and parametrized tests assert the gate and the factory accept and reject exactly the same values. The PR contract wording is what was wrong; the code was right.

---

packages/sbir-analytics/sbir_analytics/assets/cet/utils.py
hollomancer: Resolved on the current head. `_read_parquet_or_ndjson` raises `"<path> contains no records"` for an empty decoded artifact on both routes — zero-row Parquet and blank/whitespace-only NDJSON — and callers opt in explicitly via `allow_empty=True` rather than getting it by default.

Both formats are covered: `test_reader_rejects_empty_ndjson`, `test_reader_rejects_empty_parquet`, and `test_reader_allows_empty_only_when_explicitly_requested`.

---

packages/sbir-analytics/sbir_analytics/assets/cet/loading.py
hollomancer: Resolved on the current head — the gate is now a matched-vs-submitted ratio, not an error count.

`_complete_load` computes `match_rate = processed / submitted` and fails on `processed != submitted` with `"<op> processed {processed}/{submitted} submitted item(s) ({rate:.1%})"`, independently of `errors`. Your reproduction — 6 company rows, zero matched, `errors=0` — now fails the materialization instead of returning `{"status": "success", "companies": 6}`.

The two smaller points are closed as well. The summary is written *before* the raise, so `neo4j_*.checks.json` records the failing run (`status`, `submitted`, `processed`, `match_rate`, `errors`, serialized metrics) rather than leaving the previous run's `"success"` on disk — `test_zero_progress_fails_after_persisting_failure_summary` and `test_nonzero_loader_errors_fail_asset_and_persist_summary` pin that. `test_company_enrichment_fails_when_no_organizations_match` covers the `key_property="company_id"` mismatch you traced.

Still open from this thread: the three swallows below the asset layer — `base.py:301-302` on constraint/index creation, and `cet.py:383`/`cet.py:445` on malformed supporting-CET and evidence entries — none of which increment `errors`, so they stay invisible to this gate. That's loader-side work rather than asset-side.

---

PR #580, opened 2026-08-09
Build deployment image from uv.lock

hollomancer:
## Stack note

This PR is stacked on #573 (`agent/retire-obsolete-runtime-claims`). Review only the changes above that branch. Retarget to `main` after #573 merges.

## Summary

- build the production image from the committed `uv.lock` with `uv sync --locked`, rejecting manifest/lock drift without updating the lock during the build
- pin the Python base image by multi-architecture digest and pin the uv installer version
- replace the obsolete split base/full Dockerfiles with one deployment image covering Dagster, graph, ML/NLP, fiscal, streaming, and USPTO browser workloads
- make Docker CI react to lockfile and workspace dependency changes
- keep server preflight checks ahead of image build and update deployment documentation

## Verification

- `uv lock --check --offline`
- focused Docker/server/runtime/import suite: 51 passed
- `uv run ruff check` and formatting checks
- `make docs-check`
- development, CI, and server Compose configurations parse successfully
- native arm64 runtime and test targets build successfully
- the runtime image reports Dagster 1.13.17, loads the locked `en_core_web_sm` 3.8.0 model, and validates `sbir_analytics.definitions`
- `git diff --check`

## Dependency note

Adding the explicit server runtime contract moves the locked Dagster family from 1.13.1 to 1.13.17 and adds the webserver/model dependencies captured in `uv.lock`. The application environment is lock-driven; the separately pinned uv bootstrap and isolated wheel-build tools remain outside that application lock.


Dockerfile
```diff
@@ -1,53 +1,60 @@
...
+    && python -m pip install "uv==${UV_VERSION}"
+
+# Copy the complete Python workspace before syncing so first-party packages are
+# installed as ordinary wheels rather than relying on broad hand-maintained
+# `pip install` ranges or PYTHONPATH-only imports.
+COPY pyproject.toml uv.lock README.md ./
+COPY sbir_etl/ ./sbir_etl/
+COPY packages/ ./packages/
+
+RUN uv sync --frozen --no-dev --extra server --no-editable
```
hollomancer: **`--frozen` doesn't do what the header comment says it does.** Line 5-6 states "`uv sync --frozen` refuses an out-of-date lock." It doesn't — that's `--locked`.

- `--frozen` = "sync without updating `uv.lock`." Installs the lock as written, never compares it against the manifests.
- `--locked` = "assert the lock is up to date," non-zero exit otherwise.

So the reproducibility half of the goal *is* met (no re-resolution at build time), but **drift detection is missing**, and there's no compensating guard anywhere — I checked `.github/workflows/`, the `Makefile`, and `scripts/setup_dev.sh` for `uv lock --check` or `--locked` and found none.

Concrete path: someone adds a dependency to `packages/sbir-analytics/pyproject.toml` and forgets `uv lock`. Local `uv sync` re-resolves and works fine. CI's Docker job — now correctly triggered by your new `pyproject.toml` path filter — builds green against the stale lock. The missing package surfaces as an `ImportError` inside a Dagster asset on the live host.

One word, plus fixing the comment.

**Two other things about this stage, both operational:**

**Layer ordering means every code-only rebuild reinstalls everything.** `uv.lock` is correctly copied before the sync (line 38 — the caching concern is handled), but `COPY sbir_etl/` (39) and `COPY packages/` (40) come *before* this `RUN`. Any source edit invalidates the sync layer and the `playwright install --with-deps chromium` layer at line 46, forcing a full reinstall of spacy/scikit-learn/pyarrow/duckdb/dagster plus a fresh ~150 MB Chromium download. `# syntax=docker/dockerfile:1.7` on line 1 already enables cache mounts and none are used. This is operationally load-bearing: `Makefile:613` warns that `server-rebuild` kills in-flight Dagster runs, so a longer rebuild is a longer outage. The standard split is `uv sync --frozen --no-install-project --no-install-workspace` with `--mount=type=cache,target=/root/.cache/uv`, then copy source, then sync again.

**spacy is installed but `en_core_web_sm` isn't.** The retired `Dockerfile.full:29` ran `python -m spacy download en_core_web_sm`; this image doesn't, yet the PR claims ML/NLP coverage and `docker-compose.server.yml:43` defaults `DAGSTER_LOAD_HEAVY_ASSETS=true`. `evidence_extractor.py:106-119` wraps `spacy.load(...)` in `try/except Exception` and sets `self.nlp = None`, so the result is a logged error and **silently empty evidence extraction** rather than a crash — the worse outcome. Add the download or state that NLP evidence extraction is out of scope for this image.

---

hollomancer (commented):
## Sound design, one word short of its central claim — and the stacking needs care

The core is genuinely better than what it replaces: the image really is built from the committed lockfile, the Python base is pinned by a digest I verified is a true multi-arch index (linux/amd64, arm64/v8, arm/v5, arm/v7, 386, ppc64le, and the current digest of `python:3.11.9-slim-bookworm`), `uv.lock` is copied before the sync, the `server` extra covers the runtime, and nothing pip-installs at container start in the deployment path. `uv lock --check --offline` passes at the head, `check_versioning.py` passes at 0.3.0, and the green "Docker Build and Smoke Test" job empirically confirms the image builds and `dagster definitions validate` loads.

**The stale `workspace.server.yaml` assertion is already fixed** — `04323dc8` aligned `test_server_runtime_contract.py:63` with the relative `COPY` form. That was the one hard CI failure.

**The remaining substantive item is `--frozen` vs `--locked`** — the Dockerfile's own comment claims a guarantee the flag doesn't provide, and nothing else in CI or the Makefile catches lock drift. Detail inline, along with layer ordering and the missing spacy model.

### Stacking — the thing to be careful about

**#573 must merge first.** This PR's base is `agent/retire-obsolete-runtime-claims`, not `main`.

`git merge-tree --write-tree origin/main origin/agent/reproducible-deployment` exits 0 — but **that clean merge carries #573's commits too**. Merging #580 into main right now lands the entire stack, so #573's content would arrive under #580's review. That's the failure mode to guard against, not a conflict.

If #573 changes, the exposed files are `docs/deployment/self-hosted-server.md`, `docs/development/docker.md`, and `tests/unit/test_server_ops.py`. `uv.lock` is the dangerous one — if #573 ever touches it, resolve by re-running `uv lock`, never by hand-merging hunks. If #573 is rejected, nothing in #580's functional payload depends on it semantically; a rebase onto main should replay cleanly except those three files.

### Undisclosed dependency bump

`uv.lock` moves `dagster`, `dagster-pipes`, and `dagster-shared` from **1.13.1 → 1.13.17**, and adds 17 packages (`dagster-graphql`, `dagster-webserver`, `starlette`, `uvicorn`, `uvloop`, `watchfiles`, `websockets`, `graphene`, `gql`, `yarl`, …). It's forced by adding `dagster-webserver>=1.7.0,<2.0.0` at `packages/sbir-analytics/pyproject.toml:17`, it's within the declared range, and both `uv lock --check` and the versioning script pass — so no policy violation. But the PR body lists no dependency version changes, and per CLAUDE.md's live-deployment section this ships a new orchestrator patch series to the live host. Worth stating explicitly and classifying at release time.

Related: making `dagster-webserver` a hard dependency of the code-location package pulls the webserver and those 17 transitive packages into the gRPC code server and into every dev/CI environment via `stack-dev`. A `webserver` extra consumed only by the image would be tighter. Simplest-thing-that-works is defensible; flagging for awareness.

### Two compose-level breakages, neither on the server path

- **`docker-compose.yml:215` — `- ./:/app:ro` on `etl-runner` shadows the new `/app/.venv`.** All dependencies now live there (`UV_PROJECT_ENVIRONMENT=/app/.venv`), and `.venv/` is in `.dockerignore:15`, so bind-mounting the host repo over `/app` leaves nothing to restore it. `docker compose --profile dev up` resolves `python` to the system interpreter and dies on `import pydantic`. On a macOS host that has run `make install`, worse: `/app/.venv/bin/python` exists but is a Mach-O binary that can't exec in Linux. This worked before because the old image pip-installed into system site-packages. **Not on the server profile** — `docker-compose.server.yml` mounts only `config`, `data`, `reports`, `logs`, `artifacts`, `dagster_home`, none of which collide.
- **`docker-compose.yml:274-289` — the `ci`-profile `app` service still installs deps at container start**, with `uv pip install --system`, which targets `/usr/local/lib/python3.11/site-packages` while every project dependency now lives in `/app/.venv`. So `pytest -m fast` collects against an interpreter with no `pydantic`/`loguru`/`dagster`. Reached via `Makefile:379` (`docker-test`), `:396`, `:450`. Dev-machine only — GitHub CI builds the image and runs `docker run` smoke tests without compose — but worth fixing or deleting given the PR's premise.

### Smaller

- `Dockerfile:20` + `:39-40` contradict the comment at `:35-37`: `PYTHONPATH=/app` is still set and `/app/sbir_etl/` is still in the final image, so `import sbir_etl` resolves to the copied source ahead of the wheel `uv sync` just built. `sbir_analytics`/`sbir_ml`/`sbir_graph` do come from wheels. Identical contents today, but the source tree is dead weight and the two copies can diverge. Either drop `PYTHONPATH` and the redundant copies, or drop the claim.
- `Dockerfile:5` "dependencies come exclusively from uv.lock" isn't literally true — `--no-editable` builds wheels in isolated environments that fetch `hatchling` from PyPI at build time, and `:33` pins `uv==${UV_VERSION}` by version but not by hash. Neither is a defect; just soften the claim, since its strength is the point of the PR.
- Build toolchain dropped (`build-essential`, `gcc`, `g++`, `libpq-dev`, `git`, `postgresql-client` → `ca-certificates`, `curl`, `netcat-openbsd`). CI proves amd64 resolves entirely from wheels, and I confirmed `curl`/`nc` are the only binaries the runtime scripts probe for and nothing shells out to `git`. But `docker-compose.server.yml:3-4` says the server builds natively — on arm64, any dependency lacking an aarch64 wheel (`pyreadstat` is closest) fails with a confusing missing-compiler error.
- Image still runs as root (`entrypoint.sh:79-84` logs "continuing as root"). Pre-existing, but a from-scratch Dockerfile rewrite was the natural moment for a `USER`.

**Volume safety checked and clean:** `Dockerfile:58` creates an empty `/app/dagster_home` before the named-volume mount, so Docker's first-use copy is a no-op and existing run/schedule state is preserved. `.env.server` is untouched and still excluded by `.dockerignore:50-52`. No ingress, port, or Tailscale change.

One note on the body's "focused Docker/server unit tests: 25 passed" — that's `test_server_ops.py` (23) + `test_import_isolation.py` (2), i.e. exactly the two files the PR updated. The selection excluded `test_server_runtime_contract.py`, which is why the stale assertion got through. Worth widening the command in the body.

---

PR #582, opened 2026-08-09
Build audited Form D control-identity universe

hollomancer:
## Summary

- add a maintained producer over the official SEC DERA quarterly Form D bulk archives, frozen to 2009Q1–2024Q4, with cache checkpoints, schema/join gates, per-quarter URLs and SHA-256 hashes, deterministic issuer aggregation, and atomic outputs
- emit a broad issuer universe, conservative candidate SBIR-CIK exclusion evidence from every historical SBIR company name using the frozen `organization-key-v1` policy, and a disjoint provisional identity-only control frame
- publish the pinned real-data audit and truth-sync the Phase 2 spec: exclusion recall remains unknown, Form D supplies SIC rather than NAICS, and the current controls are explicitly not match-ready

## Real materialization

- 64/64 quarterly archives validated
- 673,656 retained filings
- 311,809 unique issuer CIKs
- 4,465 candidate-excluded CIKs from 4,423 exact normalized-name keys
- 307,344 provisional retained identities
- zero control/exclusion overlap
- a second cache-only build reproduced all three output hashes byte-for-byte

The 1.4 GB of derived JSONL and 174 MB SEC cache remain gitignored. The tracked report and 64-quarter manifest publish their exact counts, source/output hashes, and readiness flags.

## Interpretation boundary

This PR closes the broad-source and deterministic-materialization prerequisites only. It deliberately sets:

- `complete_sbir_exclusion=false`
- `exclusion_recall="unknown"`
- `identity_only=true`
- `covariates_ready=false`
- `ready_for_matching=false`

Task 2.2 remains open pending higher-recall authoritative CIK/alias evidence and a validated SIC-to-NAICS-2 strategy. Symmetric federal-contract, patent, and M&A outcomes remain follow-on work.

## Validation

- `5840 passed, 7 skipped` across `tests/unit`
- `65 passed` across the producer plus agency-private-capital focused suite
- Ruff format/check passed
- targeted MyPy passed
- `make docs-check` passed
- `make lint-boundaries` passed
- manifest JSON/readiness/count assertions passed
- real SEC endpoint archives 2009Q1 and 2024Q4 parsed independently

## Stack

Originally a stacked draft targeting `agent/private-capital-analysis` (PR #577). It now targets `main`. The stacked children (#584–#596) were closed rather than restacked; later slices will be revived as new PRs after this producer lands.


hollomancer (commented):
## This stack forked from a superseded revision of #577, and merging it in order would revert that PR's identity fix

Reviewing this as the base of the 13-PR chain (#582 → #584 → #585 → #586 → #587 → #591, plus #589 → #592, #590, and #594 → #595 → #596), since the finding applies to all of them.

**The fork point is stale in a way that matters.** This PR's base is `agent/private-capital-analysis` at `5c7aacb`. That branch has since moved:

```
5c7aacb  ← where this stack forked
7a1907a  Fix Phase 1 company identity crosswalk
789c230  fix: match M&A events against the whole company component
```

Both of those commits exist because the identity resolution at `5c7aacb` was found defective in review. `7a1907a` replaced the first-match-wins `_company_key` cascade with an exact alias graph resolved into connected components — the fix for the mixed UEI/DUNS undercount reported on #577. `789c230` made the M&A exit join component-wide instead of per-row.

This stack still carries the old implementation. `outcomes.py` on this branch has the `_company_key` / `_name_key` two-key fallback with `already_counted_ueis` bookkeeping (`:273-283`) that `7a1907a` deleted, and the same pre-fix code is present unchanged on `agent/symmetric-outcome-contracts` and `agent/sbir-formd-identity-validation` — so it propagates the length of the chain.

Merging the stack in order re-lands the defective resolver on top of the fix. Nothing will conflict — the fix and the stale copy touch overlapping regions of a file the stack also edits, so git will take the stack's version and the regression will be silent. Worse, the committed artifacts in every downstream PR were computed against the old resolver, so the numbers in them are not reproducible from the merged code.

**Recommended:** rebase this branch onto the current `agent/private-capital-analysis` head, propagate up the chain, and regenerate the committed artifacts before any of these leave draft. Doing it now costs one rebase; doing it after #577 merges means auditing ~24,000 lines of committed evidence for which resolver produced them.

### What I did and didn't review

Reviewed: the chain topology, the fork point, and how the identity primitive flows through it. Also checked tier labeling across the stack, which is **good** — every new artifact in #582, #584, #585, #589, #590, #591, #594, #595, and #596 carries an explicit `Exploratory` label, so the epistemic-tier contract is being honored rather than quietly skipped.

Not reviewed: the substance of the individual analyses — the control-universe construction, the contamination audit, the proxy ledgers, the fail-closed source contracts, the covariate feasibility work. That is roughly 24,000 lines across 13 PRs and needs its own pass. This review is deliberately about the one defect that affects all of them, because it is cheapest to fix before that pass rather than after.

One structural note for whenever that pass happens: each PR here targets its predecessor rather than `main`, so GitHub's diff for each shows only its own increment, but merging any one of them into `main` brings the whole chain beneath it. Worth stating the intended merge order in the PR bodies.

---

hollomancer:
Rebased onto current `main` (post-#668 / #670). The three producer commits replayed cleanly; the only conflict was `docs/research-questions.md`.

Inventory resolution: kept the post-#657 refusal (&#34;It is not a crowd-in estimate&#34;) and added a non-reserved note that the 2009Q1–2024Q4 identity audit stages 311,809 issuer CIKs / 307,344 provisional retained identities, with unknown exclusion recall, no NAICS, and controls not match-ready. No `Computable` / `Validated` / `Citable` claim. The Form D control-identity audit is linked as a second report, not as a study.

`specs/status.md` already recorded the producer without closing task 2.2; that wording survived the rebase.

Local checks on the rebased HEAD: inventory Status guard, 12 producer tests, 18 Status-guard tests, Ruff on the producer, and epistemic-tier declarations — all passed.

The stacked children (#584–#596) are being closed rather than restacked. Revive later slices as new PRs after this producer lands.

specs/agency-private-capital-comparison/requirements.md
```diff
@@ -157,34 +155,50 @@ cohort estimand and identity approach still require review.
...
+   While exclusion recall is unknown, the manifest SHALL state
+   `complete_sbir_exclusion=false`; retained means only not exact-name-matched to
+   observed SBIR history, not "never SBIR." Requirement 8 remains open until a
+   higher-recall authoritative CIK/alias union is validated.
+9. **SHALL** match the agency and eligible control cohorts on vintage year,
+   validated NAICS-2, and state using coarsened-exact matching; no propensity
+   scoring in v1. DERA supplies SIC and Form D industry group, not NAICS. The
+   staging producer SHALL NOT infer NAICS and SHALL state
+   `covariates_ready=false` until a validated SIC-to-NAICS-2 strategy exists.
+   The existing matched asset SHALL refuse staging output while either gate is
```
hollomancer, line 183: **[suggestion]** Requirement 9 now says the existing matched asset SHALL refuse staging output while `complete_sbir_exclusion` or `covariates_ready` is false, and tasks 2.3/2.6 correctly stay open. This PR does not wire the new universe into `agency_private_capital_form_d_matched_comparison` (default remains `data/form_d_control_universe.jsonl`). But the new issuer records still use `cik` and `issuer_name`, which is exactly the shape `load_form_d_control_universe` already accepts in `form_d_inputs.py`. That loader then re-keys names with `CompanyNameProfile.FORM_D_JOIN_V1` (uppercase/whitespace only; it does not strip legal suffixes), ignores nested `filings`, and treats missing `filing_date` as “no year filter.” Pointing the asset at `form_d_control_identity_universe.provisional.jsonl` would therefore ingest the staging set, skip the producer’s fail-closed flags, and join on a different identity profile than the exclusion used. The split itself is explicit in the producer (`NORMALIZER = CompanyNameProfile.ORGANIZATION_KEY_V1`, manifest `normalizer_version: organization-key-v1`) and is distinct from both Phase 2 `FORM_D_JOIN_V1` and the form-d-fundraising uppercase-strip study rule; nothing in this PR currently performs that wrong join on the default path.

**Suggestion:** While 2.3/2.6 remain open, add a cheap refuse in `load_form_d_control_universe` / the matched asset: reject `*.identity-staging.jsonl`, `*.provisional.jsonl`, `firm_key`/`schema_version` issuer records, or any sibling manifest with `ready_for_matching=false` / `complete_sbir_exclusion=false` / `covariates_ready=false`. Do not wait until matching is implemented to close this accidental-consumption hole.

docs/research/agency-private-capital-form-d-control-universe.manifest.json
```diff
@@ -0,0 +1,7085 @@
+{
+  "artifact_role": "tracked_research_manifest",
+  "caveats": [
+    "Exact normalized-name exclusion has unknown recall.",
+    "Aliases, renames, spelling variation, acquisitions, and unmatched CIKs can remain.",
+    "A retained issuer means only not exact-name-matched to the observed SBIR history.",
+    "SEC Form D supplies SIC and industry group, not NAICS."
+  ],
```
hollomancer, line 9: **[suggestion]** The tracked manifest pins `code_commit` to `053ce78a3a19f0004ee3eba7857104b9ab4580f8`. That object is not an ancestor of PR HEAD `262f77ac` and is not on any branch; it is a pre-rebase commit with the same message as `ceb7e5b3`. The producer blob is identical across `053ce78a`, `ceb7e5b3`, and HEAD, and the recorded invariants/output hashes match what `build()` emits (64 quarters, 673,656 emitted filings + 23 invalid names = 673,679 selected, 311,809 − 4,465 = 307,344, archive bytes 174,385,952, required DERA headers present in every quarter). So the JSONL hashes can still be reproduced from this branch given the same zip cache and `award_data.csv`. Two paperwork gaps remain: (1) a fresh clone of this PR cannot `git show` the pinned SHA; (2) `build()` does not emit `artifact_role`, and `cache_archive()` only checks zip structure, not the tracked SHA-256s, so a network rerun can change inputs without failing closed against this ledger. The phase-2 writeup’s claim that checksums mean “a rerun cannot silently change its inputs” overstates what the CLI enforces.

**Suggestion:** Set `code_commit` to a SHA that will exist after merge (`ceb7e5b3` or HEAD) and document that the docs copy adds `artifact_role`. If the tracked file is meant to be a lockfile, add an optional `--expected-manifest` check that compares catalog/archive/output hashes before succeeding.

scripts/data/build_form_d_control_universe.py
```diff
@@ -0,0 +1,921 @@
...
+            "SEC network access requires --sec-user-agent with a project name and contact email"
+        )
+    request = urllib.request.Request(
+        url,
+        headers={"User-Agent": user_agent, "Accept": "text/html,application/zip,*/*"},
+    )
+    try:
+        with urllib.request.urlopen(request, timeout=90) as response:  # noqa: S310
+            return response.read()
+    except Exception as exc:  # pragma: no cover - exercised through an injected fetcher
```
hollomancer: **[nit]** The `# pragma: no cover - exercised through an injected fetcher` comment is false. `_fetch()` has no fetcher injection; tests only hit the cache path. The comment narrates a test seam that does not exist.

**Suggestion:** Drop the injected-fetcher claim, or thread an optional fetch callable through `load_catalog` / `cache_archive` and test the `BuildError` wrapping.

hollomancer:
Re-verified all three open findings against head `262f77ac`. **All three hold.** Posting as a single comment because the review they belong to is still in `PENDING` state, which blocks threaded replies.

### 1. Accidental-consumption hole (`requirements.md:183`) — holds, and should block

I ran a producer-shaped issuer record through `load_form_d_control_universe` rather than reasoning about it. It ingests silently:

```
rows ingested: 1
company_key          ACME ROBOTICS, INC.   <- FORM_D_JOIN_V1
first_form_d_date    (empty)
first_form_d_year    None                  <- year_min/year_max become no-ops
offering_count       1                     <- actual filing_count is 2
total_form_d_raised  0.0                   <- amounts live in `filings`, never read
```

against the producer's exclusion key `ORGANIZATION_KEY_V1` → `ACME ROBOTICS`. The mechanism is `_iter_company_form_d_rows` (`form_d_inputs.py:100-107`): `issuer_name` satisfies the name fallback, `cik` the CIK fallback, and with `offerings` absent it synthesizes one flat offering, so nested `filings` is never read. `control_cohort.py:172` passes `vintage_year=None` through, so this surfaces as **zero matched pairs rather than an error** — the worst shape for it.

What makes it blocking is that it's a requirements-vs-code gap: `requirements.md:183` states it as a SHALL, and `tasks.md` (2.3, 2.6), `specs/status.md`, and the research doc all assert the property as though it already held. Grepping `ready_for_matching|covariates_ready|complete_sbir_exclusion|provisional|staging` across `assets/agency_private_capital/` returns zero hits; `form_d_matched_asset.py:71-77` checks only `path.exists()`.

Mitigating, and worth recording: the default path is `data/form_d_control_universe.jsonl` while the producer writes under `data/processed/agency_private_capital/control_universe/`, so consumption needs a config override. A footgun, not an automatic misfire — but shipped alongside four written promises that it cannot fire.

**Fix:** one choke point at the top of `load_form_d_control_universe`, covering the asset, the integration test, and any script. Reject `.provisional.jsonl` / `.identity-staging.jsonl` filenames; a first record carrying `firm_key` or `schema_version`; or a sibling manifest with any of the three gates false. `ValueError` naming the failed gate, ~20 lines. Existing tests at `test_form_d_inputs.py:83` and `:114` already pin that the legitimate shape still loads.

### 2. Manifest `code_commit` — holds, with a correction to the original wording

I said the object "is not on any branch"; anything stronger would have been wrong:

```
git fetch origin 053ce78...                       -> succeeds (GitHub serves it by SHA)
git merge-base --is-ancestor 053ce78... 262f77ac  -> exit 1
git branch -a --contains 053ce78...               -> (empty)
```

It's the pre-rebase twin of `ceb7e5b`, orphaned by a force-push — reachable only via the PR ref, GC-eligible. But the pin is *content*-accurate: `git diff 053ce78 <branch> -- scripts/data/build_form_d_control_universe.py` is empty. This is provenance paperwork, not a wrong number.

**Constraint that matters:** the tracked manifest can't be regenerated without a full 174 MB SEC re-download, so don't change the producer's manifest schema to fix it. Two hand edits: set `code_commit` to `262f77ac`, and add one sentence to "Products and provenance" noting the tracked copy adds `artifact_role: tracked_research_manifest`. Caveat — **a squash merge orphans `262f77ac` too.** A content-addressed `code_blob_sha1: d0356810f8d528f670665e7096dd231127044a9d` survives rebase and squash alike and is strictly better here than any commit SHA.

Withdrawing the `--expected-manifest` half of the original suggestion as a blocker — it's a real scope increase over a documentation gap, and task 2.2 is its right home.

### 3. False pragma (`build_form_d_control_universe.py:209`) — holds, and is weaker than first described

`_fetch` takes no fetcher parameter and no test has a fetch seam. The part missed first time: `--cov` in `Makefile:178` and `ci.yml:239-244` scopes to `sbir_etl` plus the three `packages/` trees — **`scripts/` isn't measured at all**, so the pragma is inert as well as false. Delete it; nothing else changes.

### Mergeability

Close. The producer is deterministic (byte-identical re-run asserted at `test_producer.py:463-466`), fails closed on six named invariants, and the epistemic labeling is disciplined — `identity_only`, `complete_sbir_exclusion=false`, `covariates_ready=false`, `exclusion_recall="unknown"` propagate consistently into the manifest, `specs/status.md`, `docs/research-questions.md`, and the non-citable banner. Nothing is promoted to `evidence` tier.

Total remediation is ~30-40 lines across four files plus two tests, **none of it requiring an SEC re-download** provided the manifest schema is left alone. Items 2 and 3 are ten minutes. Item 1 is the only one carrying weight, and it's the one thing here that could quietly produce a wrong number later.

Remaining gaps are correctly tracked as open tasks 2.1-2.9, not defects in this PR.

---

hollomancer (commented):
Submitting the previously pending review so the three inline threads are visible. Re-verified against `262f77ac`: all three findings still hold. Item 1 (accidental consumption of staging JSONL) is the merge blocker; items 2 and 3 are paperwork. See also the 2026-08-31 conversation comment for the measured reproduction.

hollomancer:
Independent re-review of the three now-visible inline threads (the pending review is submitted). **Resolve none of them.** All three still hold at `262f77ac`.

Additional facts from this pass:

- **Behind `main` by one commit** (`a66e435` → `241fe9d`, the v0.11.0 versioning bump). `git merge-tree` is clean. Nice-to-have rebase, not a content blocker.
- **No current file-level collision with #691 on `form_d_inputs.py`.** #582 does not touch that file. The collision is future ownership of Form D amounts (DERA producer vs a later `form_d_inputs.py` rewrite). Land this producer first; put any chain-resolution task on this spec, not on a new stub.
- Thread 2 paperwork: prefer `code_blob_sha1: d0356810f8d528f670665e7096dd231127044a9d` over pinning HEAD. A squash merge would orphan `262f77ac` the same way rebase orphaned `053ce78a`. The tracked copy already has `artifact_role`.
- Thread 3: delete the pragma; do not add a fetcher seam (`scripts/` is exploratory).

Merge bar is still thread 1: a refuse in `load_form_d_control_universe` for staging/provisional JSONL, `firm_key`/`schema_version` records, or a sibling manifest with any of the three gates false. Path separation is not a substitute for a SHALL.

hollomancer:
The staging-input guard from finding 1 is implemented on branch [`claude/review-open-prs-3y08y9`](https://github.com/hollomancer/sbir-analytics/compare/main...claude/review-open-prs-3y08y9) (commit `b6f52cd6`), since I couldn't push here.

`load_form_d_control_universe` now fails closed on all three vectors before parsing: a `.provisional.jsonl` / `.identity-staging.jsonl` filename, a first record carrying `firm_key` or `schema_version`, or a sibling `form_d_control_universe.manifest.json` with `ready_for_matching` / `complete_sbir_exclusion` / `covariates_ready` false. `ValueError` naming the failed gate, ~45 lines including the comment explaining why the shapes collide. Three `pytest.raises` regressions added; the two existing tests pinning the legitimate shape still pass.

I put it in the loader rather than the asset so it covers the asset, the integration test, and any script — one choke point, per the original suggestion.

**Two things still open on this PR**, both hand edits I did not make since they belong to your branch: `code_commit` → a SHA that survives merge (with the `code_blob_sha1` caveat about squash merges in my comment above), and deleting the false pragma at `build_form_d_control_universe.py:209`.

**One process note:** this PR has an unsubmitted **PENDING review** (`#pullrequestreview-5034877711`) holding the three findings as draft comments. GitHub allows one pending review per user, so it blocks all threaded replies — which is why my verification is posted as plain comments. I tried to submit it for you and couldn't: the API tool resolves "the latest review" to the most recently *submitted* one (`#pullrequestreview-4920359158`) and refuses, so it can't reach the pending draft. Submitting it takes one click in the GitHub UI, and doing so will turn those three drafts into real threads.

---

hollomancer:
**Heads-up for whoever picks this branch up: finding 1's staging guard is already on `main`.**

It shipped in **v0.12.0** via #692, in `load_form_d_control_universe` — rejecting `.provisional.jsonl` / `.identity-staging.jsonl` filenames, records carrying `firm_key` or `schema_version`, and a sibling manifest with `ready_for_matching` / `complete_sbir_exclusion` / `covariates_ready` false, with three regression tests. This branch predates it, so after a rebase you'll get the guard for free. **Don't write a second one.**

It's now tracked as task **F.1** in `specs/agency-private-capital-comparison/tasks.md` (see #693), which is where Form D input fidelity lives — that spec owns `form_d_inputs.py` and is the only consumer of it.

Two items on this PR are still open and are both hand edits on your branch: `code_commit` pointing at a SHA that survives merge (with the `code_blob_sha1` caveat about squash merges), and deleting the false pragma at `build_form_d_control_universe.py:209`.

One thing worth knowing that isn't about this PR directly: v0.12.0 also changed `total_form_d_raised` and `offering_count` in that same loader, to stop Form D amendments double-counting. Since Phase 1 is materialized-but-unsigned, any figure carried forward from before that release should be re-checked before sign-off.

---

specs/agency-private-capital-comparison/requirements.md
hollomancer: Implemented in `73937dba`. Agreed with the framing: the hole is worth closing now precisely *because* nothing performs the wrong join yet.

The trap is that the shapes are compatible. The staging producer emits `cik` and `issuer_name` (`build_form_d_control_universe.py:546`, `:689`), and `_iter_company_form_d_rows` accepts exactly `rec.get("form_d_cik") or rec.get("cik")` and `issuer_name` — so a mis-pointed asset would load **cleanly**, with no error to notice, while skipping `ready_for_matching` / `complete_sbir_exclusion` / `covariates_ready` and joining on `FORM_D_JOIN_V1` instead of the `ORGANIZATION_KEY_V1` the exclusion used.

`_refuse_staging_input` refuses three ways, matching the list in your suggestion:

1. **Filename suffix** — `.identity-staging.jsonl` and `.provisional.jsonl`.
2. **Sibling manifest** — any of the three ready flags reporting `false`.
3. **Record shape** — `firm_key` / `schema_version`, which catches a *renamed* staging file. This is the one that matters most, since the first check is defeated by `cp`.

The error names the key-profile mismatch rather than only the filename, so someone hitting it learns why it is refused rather than just that it is.

Four regressions: provisional filename, identity-staging filename, staging-shaped records under an innocent filename, and a not-ready manifest. 63 `agency_private_capital` tests pass; `make lint` (mypy 329 files) and `make lint-boundaries` pass. Tasks 2.3/2.6 stay open — this closes the accidental-consumption path, not the matching work.

---

scripts/data/build_form_d_control_universe.py
hollomancer: Confirmed and fixed in `73937dba`. There is no fetcher injection anywhere in `build_form_d_control_universe.py` — `_fetch` calls `urllib.request.urlopen` directly, and the tests only reach the cache path. The comment narrated a test seam that does not exist.

Took the first of your two options: the pragma now reads `# pragma: no cover - live network path, not exercised in tests`, which is what is actually true.

Did not thread an optional fetch callable through `load_catalog` / `cache_archive`. That adds a parameter to two public functions to test one `raise ... from exc`, and tasks 2.3/2.6 are still open on this draft — a bigger change than the comment defect warrants. Say the word if you want the seam anyway and I will add it with the `BuildError`-wrapping test.

---

docs/research/agency-private-capital-form-d-control-universe.manifest.json
hollomancer: Confirmed and fixed in `fd922521`. Verified the unreachability directly: `git cat-file -t 053ce78a3a19f0004ee3eba7857104b9ab4580f8` fails in this clone. It is not an ancestor of any branch, so a fresh checkout of this PR cannot resolve it — which defeats the point of pinning it.

Repinned to `ceb7e5b35853e119774e5a95ee4bfee906b0f65b`, confirmed an ancestor of the branch head. Added `code_commit_note` recording the old pin and why it was replaced, so the change is auditable rather than looking like a silent correction.

Also added `artifact_role_note`. Your point that `build()` does not emit `artifact_role` is right, and it is more isolated than it looked: `artifact_role` appears **nowhere else in the repository** — no producer, no consumer, no documentation. A reader comparing this tracked file against producer output would find an unexplained extra key with nothing to look it up in. The note says it belongs to the docs copy and is not part of the producer's output schema.

**Not done: the `--expected-manifest` check.** That is the one that would actually make this a lockfile, and your framing of why is correct — `cache_archive()` validates zip structure, not the tracked SHA-256s, so a network rerun can still change inputs without failing closed against this ledger. It is a new CLI feature rather than a paperwork correction, so I left it out of a change that is otherwise fixing a bad pin. It also means the phase-2 writeup's claim that checksums mean "a rerun cannot silently change its inputs" overstates what the CLI enforces today — that sentence should be softened or the check should be built. Your call which.

`make lint-boundaries` passes; the 12 `build_form_d_control_universe` tests pass.

---

packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/form_d_inputs.py
```diff
@@ -68,6 +68,55 @@ def load_form_d_matches(
...
+            raise ValueError(
+                f"{manifest} reports {', '.join(not_ready)} = false; "
+                f"{path} is not ready for matching."
+            )
+
+    if records and all(any(key in rec for key in _STAGING_RECORD_KEYS) for rec in records[:10]):
+        raise ValueError(
+            f"{path} carries staging issuer records ("
+            f"{'/'.join(_STAGING_RECORD_KEYS)}) and is not a control universe."
+        )
```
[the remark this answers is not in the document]
hollomancer: Confirmed, and worth being precise about the mechanism: `any(key in rec for key in _STAGING_RECORD_KEYS)` per record, checked across the first 10 records, means a dataset where *every one of those 10 records* carries `schema_version` alone (no `firm_key`) would still trip the guard. `schema_version` is a generic enough name that a future, unrelated control-universe format could reasonably carry it.

Fixed by requiring both keys per record: `all(key in rec for key in _STAGING_RECORD_KEYS)`. I checked the actual producer to make sure this doesn't weaken the real detection — `firm_key` and `schema_version` are always emitted together in the same record (`build_form_d_control_universe.py:546/554` and `:689/691`), so requiring both still catches every real staging file.

One more thing worth flagging: **this exact bug also existed in the version already on `main`** (shipped in v0.12.0, before this PR's own copy of the guard existed) — same `any`-over-one-key pattern, checked on only the first record rather than 10. Merging this branch with `main` produced two live, duplicate implementations of the same guard function; I consolidated them into one, carrying this fix, rather than keeping either copy. See the commit for the full account.

Added a regression that's proven to exercise the fix rather than pass trivially — I ran it against the old `any`-based logic first and confirmed it fails there, then confirmed it passes against the fix.

---

scripts/data/build_form_d_control_universe.py
```diff
@@ -0,0 +1,921 @@
...
+    except BuildError:
+        counters["invalid_optional_date_values"] += 1
+        return None
+
+
+def _aliases(row: Mapping[str, str]) -> list[str]:
+    return sorted(
+        {str(row.get(column) or "").strip() for column in ISSUER_ALIAS_COLUMNS if row.get(column)},
+        key=lambda value: (normalize_company_name(value, profile=NORMALIZER), value),
+    )
```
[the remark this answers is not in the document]
hollomancer: Confirmed. `if row.get(column)` checks the raw value's truthiness before `.strip()` runs, so a whitespace-only column (`"   "`) is truthy, passes the filter, and only then gets stripped to `""` inside the set comprehension — an empty string ends up in the alias set.

Fixed by stripping first, then filtering on the stripped result:
```python
def _aliases(row: Mapping[str, str]) -> list[str]:
    values = (str(row.get(column) or "").strip() for column in ISSUER_ALIAS_COLUMNS)
    return sorted(
        {value for value in values if value},
        key=lambda value: (normalize_company_name(value, profile=NORMALIZER), value),
    )
```

---

scripts/data/build_form_d_control_universe.py
```diff
@@ -0,0 +1,921 @@
...
+    parser.add_argument("--code-version", help="Pinned producer commit for the manifest")
+    return parser.parse_args(argv)
+
+
+def main(argv: list[str] | None = None) -> int:
+    try:
+        manifest = build(parse_args(argv))
+    except BuildError as exc:
+        print(f"error: {exc}")
+        return 1
```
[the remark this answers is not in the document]
hollomancer: Confirmed and fixed — `print(f"error: {exc}")` now writes to `file=sys.stderr` (already imported). The success path's `json.dumps(...)` output stays on stdout, so a caller piping stdout into a JSON parser no longer risks getting the error string instead.

---

packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/form_d_inputs.py
```diff
@@ -79,10 +128,16 @@ def load_form_d_control_universe(
 
     The control universe is expected to be broader than SBIR matches. Any CIK
     already present in the SBIR matched set is removed before matching.
+
+    Staging and provisional identity artifacts are refused outright; see
+    ``_refuse_staging_input``.
     """
 
+    records = list(read_jsonl(path))
```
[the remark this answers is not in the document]
hollomancer: Correct, and resolved as a side effect rather than a direct edit: this PR's copy of the guard duplicated one already shipped on `main` in v0.12.0, and I consolidated the two into one function while merging `main` in (see the commit). The surviving version reads `records = read_jsonl(path)` directly, matching `read_jsonl`'s declared `list[dict[str, Any]]` return type with no redundant wrap.

---

hollomancer:
**Merged `main` in and fixed Copilot's four findings** (`ad0c2bfd`).

This PR had gone into conflict — `main` shipped v0.12.0 with its own copy of the same staging-input guard this branch adds (my finding 1 from a few days ago, fixed independently on both sides). Merging produced two live, duplicate implementations of `load_form_d_control_universe`'s guard plus a small textual conflict. I consolidated into one function rather than keeping either copy or picking a side — see the four inline replies for the detail on which parts of each survived and why.

The headline: **Copilot's `any`/`all` finding was real, and it was in both copies**, including the one already merged to `main`. Fixed by requiring both `firm_key` and `schema_version` together, which is what the real producer always emits (`build_form_d_control_universe.py:546/554, 689/691`), so nothing real slips through — only the false-positive on a hypothetical future format that reuses the generic name `schema_version` alone. Added a regression proven to fail against the old logic and pass against the fix, not just pass trivially.

Also fixed: `_aliases()` filtering on truthiness before `.strip()`, letting whitespace-only columns leak an empty string into the alias set; and the CLI error path printing to stdout, where it would mix with the success path's JSON payload. The fourth finding (redundant `list()` wrap) resolved itself as a side effect of the consolidation.

Verified: 78 tests pass across `agency_private_capital` and the producer's own suite; `make lint-boundaries` passes.

Nothing else outstanding from my side — the three findings from my original review are confirmed genuinely fixed, and now Copilot's four are too.

---

hollomancer:
Fresh review of head `ad0c2bfd`, then pushed `d9c5dcf`.

**One real bug.** The invariant gate at the old line 816 tested `value is True or value == 0` over a dict mixing five booleans and one count. `False == 0` in Python, so a false boolean invariant (`broad_ciks_unique`, `control_ciks_unique`, and the rest) passed the gate and the build exited 0 with `complete: true`. Only the overlap count could ever fail it. Booleans and counts are now checked separately in `failed_invariants()`, with a test that pins the `False` case; the existing tests asserted happy-path values only.

**TSV hardening.** DERA tables are unquoted TSV, but the reader used csv's default `QUOTE_MINIMAL`. Probed: an entity name starting with `"` swallowed the rest of its row and the next row, with `CIK: None`. The live run's equal row counts do not rule that out because accession is column 1. Reader now uses `QUOTE_NONE`; a UTF-8 decode failure raises `BuildError` instead of a traceback. Test covers both.

**Provenance.** `docs/research/agency-private-capital-phase2-form-d.md:17-18` said the manifests pin checksums "so a rerun cannot silently change its inputs." `cache_archive()` only checks zip structure and `build()` never compares against a prior manifest, so the sentence now says what the code does and points at task 2.2. The tracked manifest gains `code_blob_sha1: d0356810…` (`git hash-object` of the producer at `code_commit`); `ceb7e5b3` is an ancestor of this head but not of `main`, so a squash merge orphans it exactly as the earlier rebase orphaned `053ce78a`. The blob hash survives both.

**Verified clean:** exactly one `_reject_staging_universe` after the merge; `main`'s amendment double-count fix is intact (`form_d_inputs.py:196-243`); nothing from #577 in the diff; quarter parsing, CIK normalization, deterministic ordering, atomic writes, join gates, and the `ORGANIZATION_KEY_V1` exclusion all trace correctly. 80 tests across the producer and `agency_private_capital` pass; `ruff format --check .` and docs-check pass.

The PR body's "Stack" section still says this targets #577. It targets `main` now; worth a one-line edit.

---

PR #583, opened 2026-08-09
Repository presentation: honest README framing, repo-wide lint, consistent naming

hollomancer:
## Description

Addresses how this repository reads to an outside reviewer who clones it cold. The engineering baseline is already strong — ~5,800 unit tests pass in under a minute, MyPy is clean, six custom architecture guards pass — but several things a reviewer hits first worked against that.

**1. `ruff check .` reported 72 errors from a clean checkout.** CI linted only `sbir_etl`, `packages/`, and `tests`, so the obvious command disagreed with the gate and the ~41k lines under `scripts/` went unchecked. Fixed the errors rather than excluding the directory — unused imports and locals, f-strings without placeholders, ambiguous `l` names, lambda assignments, and implicit `zip()` truncation (made explicit with `strict=False` to preserve current behavior). CI now runs `ruff check .` over everything.

Formatting stays scoped to the primitives/pipelines trees via `[tool.ruff.format] exclude`, matching the epistemic-tier contract that exploratory work is not held to a house format. `ruff format --check .` now passes too.

**2. The README graded its own engineering.** It opened with "this code definitely isn't professional-grade engineering" directly above a description of the gates listed above. Replaced the self-assessment with what is verifiable: the research design is the contribution, the implementation is agent-written against specs and gates, and the epistemic-tier system is what makes the output checkable. The honesty about AI authorship stays; the disclaimer about engineering quality goes. Added CI/Python/license badges, promoted epistemic tiers into the reading path, and added a "Verifying a checkout" block listing the five gates that run with no credentials, no network, and no Neo4j.

**3. Naming inconsistencies.** Three spec directories used snake_case against the kebab-case convention of the other 34; `docs/nano_phase3_methodology.md` broke the same convention; `examples/` carried two SCREAMING_SNAKE README variants and a `_mock` script sitting next to the real one. All renames are content-identical with references updated, including the spec status registry.

**4. `examples/` had no index.** Added one covering what each script shows and what it needs — including the fact that the four `army_science_technology_*` files are byte-equality fixtures for `test_procurement_transition.py`, not stray output.

[...]

PR #604, opened 2026-08-12
Make E2E tests reflect production workflows

hollomancer:
## Summary
- replace the API-dependent enrichment placeholder with a hermetic execution of the production Dagster enrichment job
- keep only real cross-component production workflows in tests/e2e and reclassify mocked, component, and utility tests
- align the E2E runner, integrity contracts, and testing documentation with the executable suite

## Validation
- 6200 passed, 20 skipped; 71.02% branch coverage
- standard E2E runner: 2 passed
- make docs-check
- ruff check and format verification

## Stack
- builds on #603
- #603 builds on #602

Draft until the stack is reviewed in order.

### DIFF
--- a/docs/architecture/neo4j-epistemic-assertions-plan.md
+++ b/docs/architecture/neo4j-epistemic-assertions-plan.md
@@ -84,7 +84,7 @@ Important test gaps are:
 
 - `tests/unit/loaders/neo4j/test_transitions.py` and `test_transitions_nodes.py` use hand-built
   frames, not the active Dagster adapter.
-- `tests/e2e/transition/test_graph_queries.py` mocks result shapes rather than exercising Neo4j
+- `tests/unit/transition/test_graph_queries.py` mocks result shapes rather than exercising Neo4j
   projection semantics.
 - `tests/unit/transition/test_transition_scores_stability.py` removes dynamic timestamps before
   comparison; it does not prove stable record identity.
@@ -1252,7 +1252,7 @@ Likely files:
 - `tests/unit/loaders/neo4j/test_assertions.py`
 - `tests/unit/loaders/neo4j/test_contracts.py`
 - `tests/integration/neo4j/test_assertion_projection.py`
-- `tests/e2e/transition/test_graph_queries.py`
+- `tests/unit/transition/test_graph_queries.py`
 - `tests/unit/assets/test_dagster_definitions.py`
 - `tests/unit/api/test_repository.py`
 - `tests/unit/api/test_analytics_api.py`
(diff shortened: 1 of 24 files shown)


hollomancer (commented):
## The replacement E2E test is the real thing. The guard meant to keep it that way checks one import spelling.

Verified by running the branch, not by reading it: `pytest tests/e2e/ -m "not requires_api and not real_data"` → 2 passed; the moved unit and integrity tests → 533 passed.

**The rewritten `test_enrichment_job.py` is a genuine upgrade.** The old version was marked `requires_api`, which CI excludes everywhere, so it had never run. The new one builds real fixtures, materializes the actual Dagster job, and asserts specific behavior — `stale` resolves to exactly `["AWARD-STALE"]`, the ledger holds both awards, and one asset check named `stale_awards_threshold_check` evaluated and passed. The part I like most is `reject_api_client` raising `AssertionError` if the job ever constructs a client: hermeticity is asserted rather than assumed, so the test fails loudly if someone reintroduces a network call instead of silently going slow or flaky.

**The reclassification checks out.** Every module moved to `tests/integration/` still carries `pytest.mark.integration`, so the Neo4j job's `-m "integration and not slow"` filter still selects them, and `Full Tests` runs `tests/` wholesale on push. The three modules moved to `tests/unit/` are net-positive: they were previously reachable only from the push-only lane and now run on every PR in Fast Tests.

### Should fix — the guard is much narrower than its docstring

`test_e2e_modules_do_not_mock_pipeline_boundaries` matches `ast.ImportFrom` with `node.module == "unittest.mock"`. I probed it by inserting each spelling into a real E2E module and running the guard:

| inserted into `tests/e2e/` | guard result |
|---|---|
| `import unittest.mock` | **passes** — `ast.Import`, not `ImportFrom` |
| `from unittest import mock` | **passes** — module is `"unittest"` |
| `from pytest_mock import MockerFixture` | **passes** — different module |

So three trivially equivalent spellings walk straight through, and `monkeypatch` — pytest's built-in, needing no import at all — is invisible to it. That last one matters here because **this PR's own flagship E2E test monkeypatches two pipeline boundaries**, `FreshnessStore` at `:53` and `USAspendingAPIClient` at `:58`.

To be clear, I don't think that test is doing anything wrong: swapping in a temp store and a client that refuses to be constructed is exactly what makes it hermetic. The problem is the mismatch. A future reader sees a guard promising "no mocked pipeline boundaries in E2E," sees the flagship E2E test mocking two, and learns the guard is decorative.

Two options, either fine:
- Keep the intent and close the gap — also walk `ast.Import`, catch `from unittest import mock`, and add `pytest_mock`. `monkeypatch` still can't be caught structurally, so the docstring has to narrow regardless.
- Keep the check and narrow the claim — rename to something like `test_e2e_modules_do_not_import_unittest_mock` and say in the docstring that hermetic substitution via `monkeypatch` is the sanctioned technique, with `unittest.mock` reserved for unit and integration suites.

The second is more honest about what is mechanically enforceable, and it is the one that matches how this PR actually writes E2E tests.

### Smaller

- `run_e2e_tests.py` swapping the hardcoded `/app/artifacts/htmlcov` for `E2E_ARTIFACT_DIR` is a good catch — that path only existed inside the container.
- The `minimal` scenario now selects `-m "smoke and ..."`, and exactly one module carries `pytest.mark.smoke`. Worth a sentence in `docs/testing/e2e-testing.md` that adding an E2E test without the `smoke` marker silently keeps it out of the minimal scenario.
- The stack forked at `662675d`, one commit before #583 merged, which moved the pytest-split timings to `tests/.test_durations`. It merges clean, but this PR relocates a number of test modules, so the recorded durations for those IDs go stale and shard balance degrades until someone regenerates the file. Not a breakage — unknown tests get the average — just worth folding into the same pass.

---

PR #605, opened 2026-08-13
Cover operated analytics and enrichment paths

hollomancer:
## Summary\n- add hermetic tests for CET analytics, human sampling, inter-annotator agreement, and drift detection assets\n- cover weekly report enrichment orchestration across reuse, fallback, timeout, error, and local transformation paths\n- cover congressional-district impact allocation, NAICS mapping, summaries, and rankings\n\n## Coverage impact\n- targeted modules: 4% to 82% aggregate branch coverage\n- CET analytics: 3% to 74%\n- CET validation: 2% to 74%\n- weekly enrichment: 7% to 92%\n- district allocator: 0% to 96%\n- repository full-suite branch coverage: 71.02% to 72.73%\n\n## Validation\n- 6237 passed, 20 skipped; 72.73% branch coverage\n- 37 focused tests passed serially\n- ruff check and format verification\n- make lint-boundaries\n\n## Stack\n- builds on #604\n- #604 builds on #603\n- #603 builds on #602\n\nDraft until the stack is reviewed in order.

### DIFF
--- a/tests/unit/assets/cet/test_analytics.py
+++ b/tests/unit/assets/cet/test_analytics.py
@@ -1,0 +1,143 @@
+"""Hermetic execution tests for the operated CET analytics assets."""
+
+import json
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+from sbir_analytics.assets.cet.analytics import (
+    transformed_cet_analytics,
+    transformed_cet_analytics_aggregates,
+)
+
+
+pytestmark = pytest.mark.fast
+
+
+def _write_ndjson(path: Path, rows: list[dict]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text("".join(f"{json.dumps(row)}\n" for row in rows), encoding="utf-8")
+
+
+def _read_json(path: Path) -> dict | list:
+    return json.loads(path.read_text(encoding="utf-8"))
+
+
+def test_analytics_uses_ndjson_fallback_and_emits_successful_metrics(monkeypatch, tmp_path):
+    monkeypatch.chdir(tmp_path)
+    processed_dir = tmp_path / "data/processed"
+    _write_ndjson(
+        processed_dir / "cet_award_classifications.ndjson",
+        [
+            {"award_id": "A-1", "primary_cet": "quantum", "primary_score": 0.9},
+            {"award_id": "A-2", "primary_cet": "biotech", "primary_score": 0.8},
+        ],
+    )
+    _write_ndjson(
+        processed_dir / "cet_company_profiles.ndjson",
+        [
+            {"company_id": "C-1", "specialization_score": 0.25},
+            {"company_id": "C-2", "specialization_score": 0.75},
+            {"company_id": "C-3", "specialization_score": None},
+        ],
+    )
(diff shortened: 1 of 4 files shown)


hollomancer (commented):
## These are real tests, which is not the default outcome for a coverage PR

I went in expecting the failure mode this repo has already had to clean up once (#576, "Remove self-fulfilling pipeline tests"): 1,067 lines of new tests, a coverage number in the PR title, and assertions that only confirm a mock returned what it was configured to return. That is not what this is.

I parsed all 37 tests and classified every assertion as behavioral versus mock-call bookkeeping. **Every test has at least one behavioral assertion.** Two initially flagged as having none — `test_fetch_solicitation_topics_closes_extractor_after_error` and `test_allocate_state_impacts_fails_fast_when_district_column_is_missing` — turned out to use `pytest.raises`, which my scanner didn't count; both are fine, and the first pairs `pytest.raises(RuntimeError, match=...)` with `close.assert_called_once_with()` to pin cleanup-on-error, which is exactly the assertion that test should make. 37 passed in 4.75s locally.

The technique in `test_lookup_pi_external_data_reuses_company_results_and_deduplicates` is worth calling out as the pattern to copy:

```python
company_lookup = Mock(side_effect=AssertionError("company lookup should be reused"))
```

The wrong path is wired to explode rather than merely being asserted absent. Combined with the input fixture — `" Jane Doe "` and `"Jane Doe"` collapsing to one `"JANE DOE"` key, and a whitespace-only PI being skipped — that test pins normalization, deduplication, and cache reuse in one go. Only the external I/O boundaries are stubbed; the orchestration logic under test is real.

`test_district_allocator.py` is the strongest file: 12 behavioral assertions on the proportional-allocation path, zero mock-call assertions, and explicit coverage of the empty-input and unresolved-district cases. Allocation arithmetic is exactly the kind of thing that should be assertable without mocks, and it is.

### One thing to check before this leaves draft

**The `--cov-fail-under=70` gate this stack adds lands on a push-only lane.** `test-full` carries `if: github.event_name != 'pull_request'`, so the coverage gate never evaluates on a pull request — it evaluates after merge, on `main`. A PR that drops branch coverage below 70% goes green and turns `main` red, and whoever pushed next inherits it.

That matters more for this PR than most, because coverage is its headline. The margin is 72.73% against a 70% floor: 2.73 points. This PR is what buys that margin; the lane that enforces it can't see the PR that spends it.

Same structural trap the CET review hit on #578 — `tests/functional/` failing only in the post-merge lane. Either mirror a coverage check into the PR lane, or state in `docs/testing/` that the floor is a post-merge guard so nobody reads a green PR as coverage-safe.

### Smaller

- The coverage deltas in the body are large enough to be worth a word on what drove them — district allocator 0% → 96% is a module that had no tests at all, whereas CET analytics 3% → 74% is partial. "4% to 82% aggregate" reads as one uniform improvement across four modules that are in quite different states.
- The PR body's markdown is double-escaped: literal `\n` sequences instead of newlines, so it renders as one unbroken line.
- Same stale-fork note as #604 — forked at `662675d`, one commit before #583. Merges clean; no action needed beyond regenerating `tests/.test_durations` once the stack settles.

Nothing here blocks. The coverage-gate lane is the only item I would want resolved before merge, and it belongs to #602/#603 rather than to this PR.

---

PR #611, opened 2026-08-13
Fix v0.5.1 deployment regressions

### LINKED ISSUE #607: Fix `server-health` to validate the production server contract
## Summary

`make server-health` fails on a healthy v0.5.1 deployment because the server-profile probe still validates E2E/development assumptions instead of the variables and dependencies supplied by `docker-compose.server.yml`.

## Observed on v0.5.1

The four production containers were healthy, Dagster returned HTTP 200 over loopback and Tailscale HTTPS, and a direct read-only Neo4j query using `NEO4J_URI`, `NEO4J_USER`, and `NEO4J_PASSWORD` returned `1`.

Despite that, `make server-health` reported:

```text
Missing environment variables: NEO4J_USERNAME, SBIR_ETL__NEO4J__BOLT_URL
Missing Python packages: pytest
Neo4j connection failed: Failed to DNS resolve address neo4j-e2e:7687
```

## Root cause

`scripts/e2e_health_check.py --profile server`:

- requires `pytest`, which is intentionally absent from the locked production image;
- requires `NEO4J_USERNAME` and `SBIR_ETL__NEO4J__BOLT_URL`;
- defaults to the E2E hostname `neo4j-e2e`.

The server Compose contract instead supplies `NEO4J_USER` and `NEO4J_URI=bolt://neo4j:7687`.

## Expected behavior

`make server-health` should pass when the production stack satisfies its actual runtime contract and should fail only for a real environment, dependency, or Neo4j connectivity problem.

## Acceptance criteria

- Server-profile checks use the variables supplied by `docker-compose.server.yml`.
- The server profile does not require test-only dependencies.
- Neo4j is queried through the production service hostname/configuration.
- A regression test covers the server profile independently from E2E assumptions.

### LINKED ISSUE #608: Make host deployment helpers compatible with or enforce their Python requirement
## Summary

On macOS, the v0.5.1 `make server-check` preflight fails before deployment because the host helper is executed with Apple's default Python 3.9, while the helper uses Python 3.10-only type syntax.

## Reproduction

Host runtimes:

```text
/usr/bin/python3 --version
Python 3.9.6

/opt/homebrew/bin/python3.11 --version
Python 3.11.15
```

Running `make server-check` with the default PATH fails:

```text
File "scripts/server/tailscale-route-state.py", line 24
def classify(..., tls_host: str | None = None) -> str:
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
Tailscale Serve returned invalid JSON.
```

Putting `/opt/homebrew/bin` first in PATH allows preflight to pass.

## Expected behavior

The documented host deployment command should either:

1. work with the macOS system Python used by `python3`, or
2. explicitly detect and require Python 3.10+ with an actionable error before running the helper.

## Acceptance criteria

- `make server-check` does not surface a misleading Tailscale JSON error for an unsupported Python runtime.
- The supported host-Python version is enforced or the helper remains compatible with the documented macOS environment.
- Tests cover the minimum supported host runtime or version guard.

### LINKED ISSUE #609: Remove obsolete Compose services during `server-rebuild`
## Summary

Upgrading a live deployment from historical `v0.11` to `v0.5.1` left the retired analytics API running because `make server-rebuild` does not remove orphaned Compose services.

## Observed behavior

The build and current-service recreation succeeded, but Compose warned:

```text
Found orphan containers ([sbir-server-api]) for this project.
If you removed or renamed this service in your compose file,
you can run this command with the --remove-orphans flag to clean it up.
```

The final `server-rebuild` status still showed five services, including the retired API on loopback port 8010, although the v0.5.1 runbook defines exactly four.

The orphan was removed safely by running the documented `make server-down` target (which already uses `--remove-orphans`) followed by `make server-up`. Persistent bind mounts and the `dagster_home` volume were preserved.

## Expected behavior

A release rebuild should converge the live stack to the services in the current Compose definition, including removal of services retired by the release.

## Acceptance criteria

- `make server-rebuild` removes obsolete services without deleting persistent volumes or bind-mounted data.
- An upgrade test or scripted assertion verifies the final service set after a service is removed from Compose.
- The target continues to warn that active Dagster runs will be interrupted.

### LINKED ISSUE #610: Create and run production containers as the non-root `sbir` user
## Summary

After deploying v0.5.1, the Dagster webserver and daemon log that the intended runtime account does not exist and then continue as root.

## Observed logs

```text
Warning: sbir user is unavailable; continuing as root
Starting service: dagster-webserver

Warning: sbir user is unavailable; continuing as root
Starting service: dagster-daemon
```

All services become healthy, so this is not an availability failure, but it defeats the entrypoint's apparent least-privilege intent and makes permissions on persistent bind mounts less predictable.

## Expected behavior

The production image and entrypoint should agree on the runtime user. Dagster services should run as an intentionally created non-root account, with only the permissions needed for mounted data, reports, logs, artifacts, and Dagster metadata.

## Acceptance criteria

- The production image contains the expected runtime user/group or the entrypoint is updated to the intended account.
- Webserver, daemon, and code-server processes run non-root where feasible.
- Mounted-path ownership/permissions remain compatible with the self-hosted server workflow.
- Startup does not emit the fallback-to-root warning.

hollomancer:
## Summary

Fixes the four defects found while deploying v0.5.1 to the operated local server:

- make the server health profile validate the production Neo4j variables and dependency set;
- defer Tailscale-helper annotations so host preflight works with macOS Python 3.9;
- remove retired Compose services during `server-rebuild`;
- restore the intended non-root `sbir` runtime for all three Dagster services, including an identity-aware ownership migration for existing persistent mount trees.

The runtime account is pinned to UID/GID `1001`, migration markers record the applied identity, and CI exercises stale-marker repair against a real Docker volume. The runbook documents the host bind-mount ownership change, timeout behavior, and rollback procedure.

Closes #607.
Closes #608.
Closes #609.
Closes #610.

## Root causes

- The server health profile reused E2E-only variable names, hostname, and the test-only `pytest` dependency.
- The host route helper evaluated Python 3.10 union annotations under Apple's Python 3.9.
- The rebuild target omitted Compose's `--remove-orphans`.
- The consolidated production image stopped creating the `sbir` account and privilege-drop tool, while the entrypoint still expected them; the code server also bypassed that entrypoint.

## Validation

- `make lint`
- `make docs-check`
- 48 focused server/deployment regression tests passed after the review changes
- Tailscale route helper executed successfully under `/usr/bin/python3` 3.9
- production Compose configuration validated with the operated server environment
- complete unit suite before the review update: 5,998 passed, 7 skipped, one unrelated parallel failure; the failing Neo4j loader test passed immediately in isolation
- GitHub Docker build and entrypoint smoke tests passed before the review update; the updated workflow additionally verifies identity-aware runtime migration in a real Docker volume

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -398,7 +398,7 @@ jobs:
             sh -c 'touch /app/reports/legacy && printf "999:999\n" > /app/reports/.sbir-runtime-owner-v1'
           docker run --rm -v "$runtime_volume:/app/reports" \
             sbir-analytics:ci-${{ github.sha }} \
-            sh /app/scripts/docker/entrypoint.sh -- true
+            sh /app/scripts/docker/entrypoint.sh true
           docker run --rm -v "$runtime_volume:/app/reports" \
             sbir-analytics:ci-${{ github.sha }} \
             sh -c 'test "$(cat /app/reports/.sbir-runtime-owner-v1)" = "1001:1001" && test "$(stat -c %u:%g /app/reports/legacy)" = "1001:1001"'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -390,6 +390,19 @@ jobs:
             sbir-analytics:ci-${{ github.sha }} \
             dagster definitions validate -m sbir_analytics.definitions
 
+          runtime_volume="sbir-runtime-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          docker volume create "$runtime_volume"
+          trap 'docker volume rm "$runtime_volume"' EXIT
+          docker run --rm -v "$runtime_volume:/app/reports" \
+            sbir-analytics:ci-${{ github.sha }} \
+            sh -c 'touch /app/reports/legacy && printf "999:999\n" > /app/reports/.sbir-runtime-owner-v1'
+          docker run --rm -v "$runtime_volume:/app/reports" \
+            sbir-analytics:ci-${{ github.sha }} \
+            sh /app/scripts/docker/entrypoint.sh -- true
+          docker run --rm -v "$runtime_volume:/app/reports" \
+            sbir-analytics:ci-${{ github.sha }} \
+            sh -c 'test "$(cat /app/reports/.sbir-runtime-owner-v1)" = "1001:1001" && test "$(stat -c %u:%g /app/reports/legacy)" = "1001:1001"'
+
   verify-setup-script:
     name: Verify Developer Setup Script
     needs: [detect-changes]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ version.
 
 ## [Unreleased]
 
+### Fixed
+
+- Corrected the live-server health check to use production Neo4j variables and
+  dependencies instead of E2E-only assumptions.
+- Made the Tailscale route helper runnable with the macOS system Python used by
+  host preflight checks.
+- Made server rebuilds remove services retired from the Compose definition.
+- Restored the non-root `sbir` runtime contract for all three Dagster services,
+  including one-time ownership migration for existing persistent directories.
+
 ## [0.5.1] — 2026-08-12
 
 ### Added
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,12 @@ WORKDIR /app
 RUN apt-get update && apt-get install -y --no-install-recommends \
         ca-certificates \
         curl \
+        gosu \
         netcat-openbsd \
     && rm -rf /var/lib/apt/lists/* \
-    && python -m pip install "uv==${UV_VERSION}"
+    && python -m pip install "uv==${UV_VERSION}" \
+    && groupadd --system sbir \
+    && useradd --system --gid sbir --home-dir /app --no-create-home --shell /bin/sh sbir
 
 # Install locked dependencies before copying source. Workspace metadata is
 # enough for uv to select the `server` extra, while `--no-install-workspace`
@@ -72,7 +75,8 @@ COPY studies/phase-iii-census/ ./studies/phase-iii-census/
 COPY workspace.server.yaml ./workspace.server.yaml
 COPY data/reference/ ./data/reference/
 
-RUN mkdir -p data logs reports artifacts dagster_home
+RUN mkdir -p data logs reports artifacts dagster_home \
+    && chown -R sbir:sbir /app
 
 # Compose's CI profile adds test dependencies at image-build time. It never
 # mutates the environment when the test container starts.
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,8 +35,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         netcat-openbsd \
     && rm -rf /var/lib/apt/lists/* \
     && python -m pip install "uv==${UV_VERSION}" \
-    && groupadd --system sbir \
-    && useradd --system --gid sbir --home-dir /app --no-create-home --shell /bin/sh sbir
+    && groupadd --system --gid 1001 sbir \
+    && useradd --system --uid 1001 --gid sbir --home-dir /app --no-create-home --shell /bin/sh sbir
 
 # Install locked dependencies before copying source. Workspace metadata is
 # enough for uv to select the `server` extra, while `--no-install-workspace`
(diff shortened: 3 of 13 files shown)


hollomancer (commented):
## Four real fixes. The ownership migration is the one to look at again before this runs on the server — the `sbir` UID is unpinned, and the mounts it rewrites are bind mounts.

45 server/deployment tests pass on the branch. The diagnoses all hold up: the code server genuinely did bypass the entrypoint (it carried a bare `command:` list), `_make_exec_prefix` genuinely did have no `sbir` account to drop to, and `server-rebuild` genuinely omitted `--remove-orphans`.

**`--remove-orphans` is safe here, and I checked rather than assuming.** My first concern was the known Compose behavior where `--remove-orphans` under `--profile` sweeps away containers belonging to services in non-activated profiles. `docker-compose.server.yml` declares exactly one profile — `server` — across all four services, so there is no other profile whose containers could be caught. The only things removed are containers carrying the project label that no longer exist in the file, which is precisely the retired-service cleanup this is for.

**The `set -eu` interaction with the migration is the right way round.** `chown -R` failing aborts the entrypoint before `touch "$marker"`, so a partial migration is never recorded as complete and gets retried on the next start. Worth stating explicitly because the opposite ordering is the classic version of this bug.

### Should fix — the marker records *that* the tree was migrated, not *to which UID*

```dockerfile
&& groupadd --system sbir \
&& useradd --system --gid sbir --home-dir /app --no-create-home --shell /bin/sh sbir
```

No `--uid` / `--gid`, so `sbir` lands on whatever system UID is free at build time on that base image. That is stable *today*, but it is a property of the base image's package set, not something this repo controls — add a package that creates a system user earlier in the sequence, or bump the pinned Python image, and the number can shift.

When it shifts, the migration does not re-run. `prepare_runtime_directories` skips any tree that already has `.sbir-runtime-owner-v1`, so the container comes up as a UID that no longer owns `/app/data`. The failure surfaces as scattered permission errors deep in a materialization rather than as a startup failure — the worst possible shape for something operating on live data.

Two changes close it, either alone helps and both together are cheap:
- Pin the ids: `groupadd --system --gid 1001 sbir && useradd --system --uid 1001 --gid sbir …`.
- Make the marker carry the identity it asserts — write the uid into it (or name it `.sbir-runtime-owner-v1-<uid>`) and re-run when it does not match `id -u sbir`.

### Should fix — the migration rewrites ownership on the *host*, and the runbook does not say so

Four of the five trees are bind mounts, not named volumes:

```yaml
- ${SERVER_DATA_DIR:-./data}:/app/data:rw
- ${SERVER_REPORTS_DIR:-./reports}:/app/reports:rw
- ${SERVER_LOGS_DIR:-./logs}:/app/logs:rw
- ${SERVER_ARTIFACTS_DIR:-./artifacts}:/app/artifacts:rw
```

Only `dagster_home` is a named volume. So `chown -R sbir:sbir /app/data` is not a container-local repair — it rewrites ownership on the server's filesystem, and the operator account that currently owns those directories loses write access to its own data, reports, and logs. That is a one-way change on exactly the trees CLAUDE.md says to preserve, and it happens automatically on the next container start rather than as a deliberate operator step.

The runbook addition is good on cost ("plan the first restart for a quiet window when those trees are large") but silent on this. It should say plainly: which uid the host directories end up owned by, that the operator will need `sudo` for direct access afterward, and how to reverse it if the migration was a mistake. As written, an operator reads "transfers the five application mount roots to the container's `sbir` account" as an in-container detail.

Related and worth a number in the runbook: `server-rebuild` runs `up -d --wait --wait-timeout 300`. A recursive chown over the bulk corpora on `/app/data` can outlast 300s, in which case the rebuild reports failure while the chown is still legitimately working.

### Smaller

- `RUN mkdir -p … && chown -R sbir:sbir /app` sits after every `COPY`, so it rewrites the whole tree in a fresh layer and duplicates all of `/app` in the image. `COPY --chown=sbir:sbir` on the source copies avoids the second copy entirely.
- `entrypoint.sh:81` still reads `# The lightweight runtime image does not currently create this account.` This PR is what makes that false. The surrounding fallback is still worth keeping for other images, but the comment now describes the bug rather than the behavior.
- `test_runtime_image_provides_non_root_account_and_privilege_drop_tool` asserts the Dockerfile *contains* `useradd --system --gid sbir`. If you pin the uid as suggested, that substring assertion breaks — worth loosening to a regex now so the contract test does not argue with the fix.
- The blocked production image build (`xvfb` at `deb12u13` vs `xserver-common` at `deb12u12`) is upstream mirror skew, not this PR. But it does mean the non-root image has not been built end to end, so the `gosu` layer and the runtime `chown` are verified by string assertions and a partial build rather than by a container that actually started. Worth saying in the PR body that the migration path itself is untested against a real image, since that is the part with irreversible effects.

---

hollomancer:
Addressed all review feedback in ab81a4e7: pinned the sbir UID/GID to 1001, made migration markers identity-aware, documented host bind-mount ownership/timeout/rollback behavior, replaced the full-tree image chown with COPY --chown and owned directory creation, corrected the stale entrypoint comment and contract tests, and added a real Docker-volume migration smoke test. Local validation: make lint, make docs-check, production Compose config, shell syntax, and 48 focused tests all pass.

PR #615, opened 2026-08-14
Add STTR spinout–subcontract linkage Phase 0 spec (exploratory)

hollomancer:
## What

Phase 0, **design-only** specification for `specs/sttr-spinout-linkage/`. Exploratory tier, non-citable, criteria **not frozen**. Implementation of the RQ1 classifier is blocked until the owner resolves `open-questions.md`.

Two research questions, with dedicated inventory entries:

- **B2 — STTR spinout vs. subcontract relationship (RQ1).** Classify every STTR award's small-business↔research-institution relationship as `SPINOUT_T1` / `SPINOUT_T2` / `SUBCONTRACT` / `INDETERMINATE` from public data only. **Not run in this PR.**
- **B1 — STTR research-institution partner types.** List-based RI classifier with the headline readout: has a non-university, non-FFRDC nonprofit ever served as an STTR partner?

Plus a **design-only** matched outcome comparison (RQ2) — specified, not run.

This PR also adds a **workbench probe** (not Phase 1, not RQ2): `notebooks/explorations/b1_sttr_partner_type_commercialization.ipynb` stratifies STTR Phase II firms by a coarse RI-name heuristic (`UNIVERSITY` / `FFRDC` / `COMMUNITY_COLLEGE` / `UNTYPED`) across existing Phase III / Form D / M&A artifacts. Channel rates are reported separately; no combined commercialization winner. The notebook degrades to an input-status table when local artifacts are absent.

[...]

PR #620, opened 2026-08-15
Resolve all 12 open questions and freeze sttr-spinout-linkage as Revision 1

hollomancer:
## What

Completes `specs/sttr-spinout-linkage/` Phase 0 end-to-end:

1. **Resolves 10 of 12 open questions** (O-0 through O-5, O-8 through O-11) against their proposed defaults, three with deviations (O-1 widens D2 to PI+founders, O-2 widens the window to ±5 years, O-3's circular dependency on the task-1.4 adjudication sample is broken by freezing the method now and deferring the numeric cutoff to a post-1.4 amendment).
2. **Resolves O-12** (Bayh-Dole government-interest data source for D3) after a second research pass beyond #617's first pass — checked AUTM STATT/TransACT (both ruled out: paid, and either aggregate-only or explicitly de-identified), UCC-1 filings (ruled out: wrong transaction type, and IP-collateral parsing was never built), and SEC EDGAR full-text search (genuinely new candidate, reuses the existing `sec_edgar` client, real evidence when it fires but low/biased population coverage). Sharpened the local USPTO proxy from generic "license" wording to the actual Bayh-Dole term "confirmatory license" (37 CFR 401.14) — 12,946 verified hits in a 3M-row sample of the local bulk data. Verdict unchanged: no source, public or paid, directly supplies a named RI→SBC license record.
3. **Freezes the design as Revision 1** — all resolutions recorded in `amendments.md`, `design.md`'s raw-byte SHA-256 pinned, status updated from "Phase 0 draft, not frozen" to "FROZEN." `specs/status.md` updated from "Gated backlog" to "Active." Phase 1 (`tasks.md`) is now unblocked to begin — seed-list capture, the kernel, the cascade implementation, and the negative-control/adjudication gates remain as separate, uncompleted Phase 1 tasks. Citable status stays blocked until those gates pass.

Everything here is documentation/spec-text only — no classification code, no data ingestion, no materialization. Every amendments.md entry states explicitly that no classification result had been computed or seen.

## Type of Change

- [x] Documentation update

## Versioning Impact

## Testing

- [x] `uv run python scripts/ci/check_removed_src_references.py` (`make docs-check`) — passed
- [x] `uv run python scripts/ci/check_epistemic_tiers.py` — passed

### DIFF
--- a/specs/sttr-spinout-linkage/amendments.md
+++ b/specs/sttr-spinout-linkage/amendments.md
@@ -77,3 +77,34 @@ of [`open-questions.md`](open-questions.md). Do not treat Revision 0 as a freeze
   finding. Revision 1 cannot be recorded until O-12 resolves.
 - **Visibility at authoring:** Same as Revision 0. **No classification result had been computed or
   seen.**
+
+## Revision 0.3 — O-12 resolved after a second research pass; task 0.5 complete, still NOT frozen
+
+- **Authority:** 2026-08-14 — owner declined to accept the first research pass's verdict outright,
+  requested a second pass, then accepted the combined findings ("Satisfied").
+- **Reason:** Record the second Bayh-Dole-sourcing research pass and its acceptance, completing
+  task 0.5 (all of O-0 through O-12 now resolved).
+- **Second-pass findings:** AUTM STATT (paid, aggregate-only — ruled out on two independent
+  grounds) and AUTM TransACT (paid, explicitly de-identified — wrong grain at any price) both
+  ruled out; UCC-1 filings ruled out (IP-collateral parsing was explicitly out of scope for the
+  repo's existing UCC-1 pilot, and a UCC-1 records the SBC's own IP as loan collateral, not an
+  inbound RI license, regardless). One genuinely new candidate surfaced: **SEC EDGAR full-text
+  search** — the repo's existing `sec_edgar` client can search filing exhibits (EX-10 material
+  contracts) since 2001 via a verified multi-phrase query; real positive evidence when it fires,
+  but coverage is limited to STTR firms that later became SEC filers and each hit needs manual
+  confirmation. The local USPTO `convey_text` proxy was sharpened from a generic "license" search
+  (first pass) to the actual Bayh-Dole regulatory term **"confirmatory license"** (37 CFR 401.14):
+  12,946 hits directly verified in a 3,000,001-row sample of the local `assignment.csv.zip`
+  (~45,000 extrapolated repo-wide) — a materially more specific proxy, though it still answers
+  federal-funding nexus (contractor-to-government), not an RI-to-SBC license.
+- **Criteria impact:** `design.md`'s D3 discipline note updated to fold in both second-pass proxies
+  (sharper `convey_text` term, optional corroborating EDGAR search) as the v1 sourcing plan for
+  `D3.recorded_license_RI_to_SBC`, should it ship at all. No change to the Order-1/Order-3 cascade
+  predicates themselves — this question only ever concerned D3's *source*, not its scoring logic.
+- **Freeze status:** **Still NOT frozen — task 0.5 is now complete; task 0.6 (Revision 1 freeze) is
+  unblocked but has not been authorized.** Freezing is a separate, higher-stakes action (locks the
+  cascade, the lexicon, and the seed-list versions at a commit SHA and unlocks Phase 1
+  implementation) and requires its own explicit go-ahead, not an automatic consequence of O-12
+  resolving.
+- **Visibility at authoring:** Same as Revision 0. **No classification result had been computed or
+  seen.**
--- a/specs/sttr-spinout-linkage/amendments.md
+++ b/specs/sttr-spinout-linkage/amendments.md
@@ -42,3 +42,38 @@ of [`open-questions.md`](open-questions.md). Do not treat Revision 0 as a freeze
   the remaining open questions.
 - **Visibility at authoring:** Same as Revision 0. **No classification result had been
   computed or seen.**
+
+## Revision 0.2 — 10 of 12 open questions resolved; still NOT frozen
+
+- **Authority:** 2026-08-14 — owner resolved O-0 through O-5 and O-8 through O-11 in conversation;
+  recorded here as the numbered revision required by `open-questions.md`'s own header.
+- **Reason:** Record the owner's resolutions so `design.md` and `open-questions.md` stop describing
+  these as pending. Resolutions, several diverging from the proposed default:
+  - **O-0** (kernel): build here at exploratory tier — proposed default accepted.
+  - **O-1** (D2 scope): PI **and** founders in v1 — wider than the PI-only default. Founders are
+    scoped to D4 Form-D-derived officer/director names only; no new founder-discovery pipeline.
+  - **O-2** (D2 window): **±5 years** — wider than the proposed ±3. Sensitivity reporting
+    (±1/±2/±3/±5) remains required.
+  - **O-3** (tier thresholds): method frozen now (`company_name_similarity` under
+    `CompanyNameMetric.JARO_WINKLER` + `generic_token_guard`); **numeric cutoff explicitly deferred**
+    to a post-task-1.4 amendment to break the circular dependency on the adjudication sample.
+    `SPINOUT_T2` scoring cannot run until that follow-on amendment lands.
+  - **O-4** (D5 lexicon): ship a small hand-curated v1 lexicon — proposed default accepted; exact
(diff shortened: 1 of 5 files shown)


specs/sttr-spinout-linkage/design.md
```diff
@@ -96,21 +99,24 @@ see [O-0](open-questions.md)). **Missing or null data never stands in for one of
 | Dim | Name | Sources | Positive signal | Typed absence encodes |
 |-----|------|---------|-----------------|-----------------------|
 | **D1** | Award spine | SBIR.gov award data, `program = STTR` | SBC, RI, PI, agency, FY, abstract present (the join spine) | Missing RI/PI blocks scoring downstream; row is `INDETERMINATE` if D1 incomplete |
-| **D2** | Person trail | OpenAlex / PubMed authorship, ORCID | PI (and founders, if in scope — [O-1](open-questions.md)) matched to RI-affiliated authorship within ±N years of award (N default ±3, [O-2](open-questions.md)) | No RI-affiliated authorship found *after search* vs. person unresolvable (`generic_token_guard` fail) vs. source not queried |
+| **D2** | Person trail | OpenAlex / PubMed authorship, ORCID | PI and founders (founders = D4 Form-D-derived officer/director names only, no separate discovery pipeline — [O-1](open-questions.md), resolved) matched to RI-affiliated authorship within ±5 years of award ([O-2](open-questions.md), resolved) | No RI-affiliated authorship found *after search* vs. person unresolvable (`generic_token_guard` fail) vs. source not queried |
```
hollomancer, line 102: **The self-corroboration hole closed for D5 is still open via D4 — and this PR freezes it.**

#615's review flagged that Order 2 let a single `D5.spinout_phrase` corroborate itself. That was fixed well: line 141 now requires a "second, **distinct** dimension" and says outright that a single D5 phrase cannot corroborate itself.

But O-1's resolution, recorded on this line, scopes D2's founders to **"D4 Form-D-derived officer/director names only"** — and Order 2 still accepts `D4.form_d_officer_ri_affiliated` as a corroborator with no distinctness condition. So a single Form D officer name can be both the D2 fuzzy person match *and* the D4 corroborator, yielding `SPINOUT_T2` from one signal. Same defect class as the D5 case, reached by a different route.

This matters more than the D5 version did, because O-1 makes the two dimensions share a name source *by construction* rather than by coincidence.

PR #623 implements exactly this (`kernel.py:452`: the D4 clause is the one corroboration branch with no `primary != ...` guard), so the hole propagates from frozen spec into code.

Worth fixing before the freeze — the same "excluding whichever dimension supplied the primary positive" language already applied to D5 would do it. Afterwards it costs an amendment.

---

specs/sttr-spinout-linkage/open-questions.md
```diff
@@ -116,6 +153,11 @@ reserved for an unverified Jones & Fearon deposit (see the literature-map audit
 [`docs/research-questions.md`](../../docs/research-questions.md)). The next unreserved slot
 is `[L50]`. Do not cite Bayh-Dole informally until the anchor is added.
 
+**RESOLVED (2026-08-14):** Add `[L50]` for the statutory Bayh-Dole anchor (35 U.S.C. §§ 200–212) to
```
hollomancer, line 156: **O-8's deliverable was never carried out, and no remaining task covers it.**

This resolution commits to adding `[L50]` for 35 U.S.C. §§ 200–212 to `docs/research-questions.md`. That file is untouched by this PR and contains zero occurrences of `L50` (verified at head). Task 0.5 is checked off and no Phase 1 task in `tasks.md` picks it up.

The consequence is the rule two lines above this one — *"Do not cite Bayh-Dole informally until the anchor is added"* — becomes permanently unsatisfiable, while O-8 reads RESOLVED. Either add the `[L50]` entry in this PR, or leave O-8 open with an explicit task.

---

specs/sttr-spinout-linkage/amendments.md
```diff
@@ -42,3 +43,103 @@ of [`open-questions.md`](open-questions.md). Do not treat Revision 0 as a freeze
...
+## Revision 1 — FREEZE
+
+- **Approved:** 2026-08-14 — repository owner authorized the freeze explicitly ("Yes, freeze it"),
+  after accepting all 12 open-question resolutions (Revisions 0.2 and 0.3).
+- **Frozen file:** `design.md`, raw-byte **SHA-256:**
+  `52d8b531d56f3b91e1d3b0946e1ac91dd6f5dfeab371e3d48f87dc5e6095ac49`. A materializing asset (task
+  1.3+) must recompute this hash against the working copy before running and fail closed on any
+  mismatch — including a mismatch caused by a well-intentioned edit to `design.md` that was never
+  recorded as a further amendment here.
+- **Git-history anchor:** the commit that introduces this Revision 1 entry is the approval-record
```
hollomancer, line 122: **The "git-history anchor" convention cites the wrong precedent for an initial freeze.**

Task 0.6's own title is "freeze at a commit SHA," and this entry records none. The cited precedent doesn't support the substitution: `specs/phase-iii-census/amendments.md` pins an explicit `Commit:` on its **initial** freeze (Revision 0) and uses the identifier-free anchor only for **later** amendments. This is an initial freeze, so it maps to the Revision 0 case, not Revision 10.

The self-reference problem is real but solvable the way phase-iii-census solved it — record the SHA in a follow-up commit, or pin the parent commit that contains the frozen `design.md`.

Related, and the reason this isn't just bookkeeping: the header at line 6 requires a materializing asset to verify the SHA-256 of `design.md` **"and this file"**, but no hash or commit for `amendments.md` is recorded anywhere, and `design.md`'s freeze mechanics cover only `design.md` plus the seed lists. As written, the append-only guarantee on `amendments.md` has nothing enforcing it.

---

specs/sttr-spinout-linkage/design.md
hollomancer, line 1: **Stale "pending resolution" text survives inside the hash-pinned file** (file-level: these lines are outside the diff).

Line 3 of this file now states all open questions are resolved and the design is frozen, but three passages still read as unresolved:

- **line ~181** — the partner-type precedence order, listed in `amendments.md` as frozen under Revision 1, still says *"Proposed default … pending final ordering confirmation (O-5)"*
- **line ~273** — same pattern for O-9
- **line ~84** — same pattern for O-8

Because these bytes are what the Revision 1 SHA pins, a future reader verifying the hash gets a file that contradicts its own freeze header, and correcting the wording later requires a new hash plus an amendment. Cheaper to fix now, in the same commit that establishes the pin.

---

specs/sttr-spinout-linkage/tasks.md
hollomancer, line 1: **The O-3 circularity is deferred, not broken** (file-level: the affected line is outside the diff).

O-3's resolution freezes the similarity *method* and defers the numeric cutoff to a post-1.4 amendment, so `SPINOUT_T2` scoring cannot run until that amendment lands. But task 1.4's verify gate (line ~73) and `design.md:225` still require **"precision/recall by tier"** — which includes T2. Task 1.4 is therefore unsatisfiable as written: it cannot report T2 precision without the cutoff that 1.4 itself is supposed to inform.

Two gaps follow: no task exists for the follow-on cutoff amendment, and none exists for re-scoring T2 once the cutoff lands. Suggest scoping 1.4's gate to T1/SUBCONTRACT/INDETERMINATE explicitly, and adding a task 1.5 for the amendment plus T2 re-scoring.

---

hollomancer (commented):
Reviewed the freeze contract itself rather than prose. **The freeze mechanics check out**: I recomputed the pinned raw-byte SHA-256 of `design.md` and it matches `52d8b531…` exactly at head, and the file is byte-stable under the repo's `trailing-whitespace` / `end-of-file-fixer` hooks (no trailing whitespace, single terminating newline), so the pin will hold. `check_epistemic_tiers.py` passes, the `specs/status.md` Gated backlog → Active edit is consistent with that file's structure, and the primitives the frozen design names (`CompanyNameMetric.JARO_WINKLER`, `search_filing_mentions()`) exist as described — while `generic_token_guard` correctly does not yet and is tracked as task 1.2.

Five findings, all in the freeze contract. The first is the one I'd resolve before merging: the `SPINOUT_T2` self-corroboration hole that #615's review closed for D5 is **still open via D4**, because O-1 scopes D2's founders to Form-D-derived names — so the two dimensions now share a name source by construction. #623 already implements it that way. Fixing it here costs one clause; fixing it after the freeze costs an amendment.

The rest: stale "pending O-5/O-8/O-9" wording preserved inside the hash-pinned bytes, O-8's `[L50]` deliverable never carried out with no task covering it, task 1.4's gate unsatisfiable as written under O-3's deferral, and an initial freeze recording no commit SHA against a precedent that does record one.

---

PR #623, opened 2026-08-15
STTR spinout-linkage: exploratory kernel (task 1.2)

hollomancer:
## Summary

Implements task 1.2 of `specs/sttr-spinout-linkage/tasks.md`: build the exploratory kernel that O-0 (`open-questions.md`, RESOLVED) determined does not already exist in the repo. Four functions in `scripts/sttr_spinout_linkage/kernel.py`, grounded in `sbir_etl.identity` primitives per O-0's "(a)" resolution -- new exploratory-tier code local to this spec, not a primitives-tier promotion.

- **`resolve_identity`** -- resolves an org or person name to a normalized, guard-checked identity. Organization resolution is a thin wrapper over `sbir_etl.identity.normalize_company_name` (verifiable by grep: `normalize_company_name`/`company_name_similarity`/`SUFFIX_TOKENS` are imported from `sbir_etl.identity` and used directly, no forked normalizer). Person-name normalization is net-new (no repo primitive exists for it): case/whitespace normalization, "Family, Given" reordering, given/family split, generic-token stripping.
- **`generic_token_guard`** -- rejects a match when a name reduces to only generic tokens. Reuses `sbir_etl.identity.SUFFIX_TOKENS` as the base for organization names, extended with institutional nouns design.md calls out by name ("institute", "center", etc., which aren't legal-entity suffixes); a net-new `GENERIC_PERSON_TOKENS` set covers person-name titles/suffixes ("Dr.", "Jr.", ...).
- **`SignalAbsentReason`** -- a typed `StrEnum` modeled on `sbir_etl.identity.RecoveryStatus`'s pattern, with one member per "Typed absence encodes" case in design.md's D1-D5 evidence-dimension table: `SPINE_INCOMPLETE` (D1), `NAME_GENERIC_TOKEN_GUARD_FAILED` (D2/org), `SOURCE_NOT_QUERIED` (D2), `LICENSE_RECORDS_SPARSE` (D3, the exact code design.md names for O-12's finding), `NON_GRANT_INSTRUMENT` (D4), `SOURCE_FIELD_UNAVAILABLE` (D4/D5). A local `DimensionStatus` StrEnum mirrors ADR-005's status names (`MEASURED`/`NOT_MEASURABLE`/`NOT_APPLICABLE`/`NOT_EVALUATED`/`EVALUATION_FAILED`) since the ADR's own `sbir_etl/assertions/enums.py` doesn't exist yet.
- **`classify_linkage`** -- implements only the frozen cascade **rule structure** (Order 0-4) over pre-scored dimension evidence, not dimension scoring itself. It is pure and deterministic: no I/O, no data fetching. It does not implement D2-D5 dimension scoring (source adapters for OpenAlex/PubMed, USPTO, USASpending, Form D, the D5 phrase lexicon) -- that's task 1.3 and future PRs. The O-3-frozen similarity method (`company_name_similarity` under `CompanyNameMetric.JARO_WINKLER`, gated by `generic_token_guard`) is applied inside `classify_linkage`, but `similarity_cutoff` is a **required parameter with no default** -- O-3 explicitly deferred the numeric cutoff to a post-task-1.4 amendment, so hardcoding a value here would silently un-defer it.

No new causal graph edges, no `SPUN_OUT_OF` edge type, no `ACCEPTED`/`REJECTED` claim status -- this kernel only supports the `CANDIDATE` assertion contract from ADR-005 (`docs/architecture/neo4j-epistemic-assertions-plan.md`).

## Test plan

- [x] `uv run pytest tests/unit/sttr_spinout_linkage/test_kernel.py -v` -- 37 passed. Covers `generic_token_guard`'s rejection cases (org names dominated by suffix/institutional tokens, person names dominated by titles/suffixes), `SignalAbsentReason`'s mapping onto the D1-D5 table, and `classify_linkage`'s Order 0-4 behavior for each case, including the "absence never advances a label" rule (a `NOT_MEASURABLE` D3 blocks `SUBCONTRACT` even with a strong D4 subaward signal) and the "license absence cannot create a `SUBCONTRACT`" rule explicitly called out in design.md.
- [x] `uv run ruff check` / `uv run ruff format --check` over the new module and test paths -- pass.
- [x] `uv run python scripts/ci/check_epistemic_tiers.py` -- pass (spec's `requirements.md` already declares `exploratory`).
- [x] `uv run python scripts/ci/check_removed_src_references.py` (docs-check) -- pass.
- [x] `uv run python scripts/ci/check_architecture_boundaries.py` / `check_tier_boundaries.py` -- pass.

### DIFF
--- a/scripts/sttr_spinout_linkage/kernel.py
+++ b/scripts/sttr_spinout_linkage/kernel.py
@@ -1,0 +1,486 @@
+"""Exploratory kernel for STTR spinout/subcontract classification.
+
+Four building blocks the RQ1 cascade and the partner-type classifier will
+consume (`specs/sttr-spinout-linkage/design.md`): identity resolution, a
+generic-token guard, a typed absence-reason enum, and the pure cascade rule
+over already-scored dimension evidence.
+
+**What this module does NOT do.** It does not score any dimension (D1-D5) --
+that requires the source adapters (OpenAlex/PubMed authorship, USPTO
+assignment data, USASpending subawards, Form D, the D5 phrase lexicon) that
+task 1.3 builds. `classify_linkage` here is the frozen *cascade rule
+structure* (Order 0-4) applied to caller-supplied `DimensionAssessment`-style
+evidence -- pure, deterministic, no I/O. The `SPINOUT_T2` similarity cutoff
+(O-3) is left as a required parameter with no default: Revision 1 froze the
+comparison *method* only (`company_name_similarity` under
+`CompanyNameMetric.JARO_WINKLER`, gated by `generic_token_guard`); the numeric
+cutoff is deferred to a post-task-1.4 amendment
+(`specs/sttr-spinout-linkage/open-questions.md#o-3`).
+
+Org-name matching goes through `sbir_etl.identity.normalize_company_name` /
+`company_name_similarity` -- no forked normalizer. Person-name normalization
+has no existing repository primitive (confirmed during Phase 0, O-0) and is
+net-new here, scoped to what the D2 person trail needs: case/whitespace
+normalization, given/family splitting, and generic-token stripping built on
+the same `generic_token_guard` used for organization names.
+
+Every emitted assertion this kernel supports is a `CANDIDATE` assertion only
+(`docs/architecture/neo4j-epistemic-assertions-plan.md`, ADR-005): no
+`SPUN_OUT_OF` edge type, no `ACCEPTED`/`REJECTED` claim status. This module
+does not touch Neo4j or any graph writer; it returns plain Python values.
+
+Epistemic tier: exploratory. This is new, spec-local code per the O-0
+resolution ("(a) build the four kernel functions as new exploratory-tier code
+in this spec"), grounded in `sbir_etl.identity` primitives but not itself a
+primitives-tier promotion. Promote to a shared primitive only if a second
+consumer appears outside this spec.
+"""
+
+from __future__ import annotations
+
+import re
+import unicodedata
+from collections.abc import Iterable
+from dataclasses import dataclass
+from enum import StrEnum
+
+from sbir_etl.identity import (
+    SUFFIX_TOKENS,
+    CompanyNameMetric,
+    CompanyNameProfile,
+    company_name_similarity,
+    normalize_company_name,
+)
+from sbir_etl.utils.coercion import _blank
+
+
+EPISTEMIC_TIER = "exploratory"
+
+
+# ---------------------------------------------------------------------------
+# generic_token_guard
+# ---------------------------------------------------------------------------
+
+# Organization-name guard vocabulary: sbir_etl.identity.SUFFIX_TOKENS covers
+# legal-entity suffixes only (Inc, LLC, ...); it does not cover generic
+# institutional-noun tokens ("Institute", "Center" alone) that design.md's
+# discipline notes and this task's brief both call out by name. Built from
+# SUFFIX_TOKENS, extended -- not a forked normalizer, a guard-only addition.
+ORG_GENERIC_TOKENS: frozenset[str] = SUFFIX_TOKENS | frozenset(
+    {
+        "institute",
+        "institutes",
+        "center",
+        "centre",
+        "foundation",
+        "university",
+        "college",
+        "laboratory",
+        "laboratories",
+        "group",
+        "association",
+        "society",
+    }
+)
+
+# Person-name analog of sbir_etl.identity.SUFFIX_TOKENS: titles, generational
+# suffixes, and post-nominals that, alone or in combination, do not identify a
+# specific person. Net-new -- no existing repository list covers person names.
+GENERIC_PERSON_TOKENS: frozenset[str] = frozenset(
+    {
+        "dr",
+        "mr",
+        "mrs",
+        "ms",
+        "mx",
+        "prof",
+        "professor",
+        "phd",
+        "md",
+        "jr",
+        "sr",
+        "ii",
+        "iii",
+        "iv",
+    }
+)
+
+
+def generic_token_guard(tokens: Iterable[str], *, generic_tokens: frozenset[str]) -> bool:
+    """Return ``True`` (guard passes) iff at least one non-generic token remains.
+
+    A name dominated by generic tokens -- "The", "Inc", "Institute", "Center"
+    alone for an organization, or a bare title/suffix like "Dr." or "Jr." for
+    a person -- must never be allowed to produce an accepted match. Mandatory
+    on D2 person-name matching and on all organization-name matching
+    (`design.md`'s "Evidence dimensions" discipline notes).
+
+    Callers select the generic-token vocabulary for their name kind:
+    organization names reuse `sbir_etl.identity.SUFFIX_TOKENS`; person names
+    use `GENERIC_PERSON_TOKENS`.
+    """
+
+    return any(token and token.lower() not in generic_tokens for token in tokens)
+
+
+# ---------------------------------------------------------------------------
+# resolve_identity
+# ---------------------------------------------------------------------------
+
+
+class IdentityKind(StrEnum):
+    """Which normalization path `resolve_identity` applies."""
+
+    ORGANIZATION = "organization"
+    PERSON = "person"
+
+
+@dataclass(frozen=True)
+class ResolvedIdentity:
+    """A name resolved to a normalized, guard-checked identity.
+
+    `given_name` / `family_name` are populated only for `IdentityKind.PERSON`
+    (best-effort split, last token treated as family name, or the explicit
+    "Family, Given" order when a comma is present in the raw input).
+    """
+
+    kind: IdentityKind
+    raw: str
+    normalized: str
+    tokens: tuple[str, ...]
+    guard_passed: bool
+    given_name: str | None = None
+    family_name: str | None = None
+
+
+def _normalize_person_name(value: str) -> tuple[str, tuple[str, ...]]:
+    """Net-new person-name normalization: no existing primitive covers this.
+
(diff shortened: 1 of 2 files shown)


scripts/sttr_spinout_linkage/kernel.py
```diff
@@ -0,0 +1,486 @@
...
+    fuzzy_person = (
+        d2.person_similarity is not None
+        and d2.person_similarity >= similarity_cutoff
+        and d2.person_guard_passed
+    )
+    any_person_link = d2.exact_person_ri_affiliation or fuzzy_person
+    any_ip_link = d3.patent_assigned_to_ri_with_sbc_inventor or d3.recorded_license_ri_to_sbc
+
+    # Order 1: one exact person or IP link with affiliation evidence.
+    if d2.status is DimensionStatus.MEASURED and d2.exact_person_ri_affiliation:
```
hollomancer, line 433: **`generic_token_guard` is bypassed on the exact path — this is the leakage negative control #2 exists to catch.**

`fuzzy_person` (line 423) correctly requires `person_guard_passed`, but this Order-1 branch tests `d2.exact_person_ri_affiliation` alone. Confirmed by execution: `exact_person_ri_affiliation=True, person_guard_passed=False` returns `SPINOUT_T1`. `design.md` calls the guard mandatory on D2 person names without qualifying it to the fuzzy path.

The Order-3 consequence is worse. `any_person_link` (line 428) is `exact or fuzzy_person`, so a **0.99 similarity that fails only the guard** collapses to `any_person_link=False`, and Order 3 reads that as `D2 measured and not any_person_link` — a **measured negative** — returning `SUBCONTRACT` on a positive subaward. Confirmed by execution.

That inverts the spec's own distinction: a guard failure means *person unresolvable* (`NAME_GENERIC_TOKEN_GUARD_FAILED`, which this module defines), not *no link found after search*. Typed absence is supposed to block a label, never advance one. Suggest gating the exact branch on the guard too, and treating a guard failure as `NOT_MEASURABLE` for D2 rather than folding it into the negative.

---

scripts/sttr_spinout_linkage/kernel.py
```diff
@@ -0,0 +1,486 @@
...
+            rationale="exact D3 IP link (patent assignment or recorded license)",
+        )
+
+    # Order 2: a fuzzy positive corroborated by a second, distinct dimension.
+    d2_fuzzy_positive = d2.status is DimensionStatus.MEASURED and fuzzy_person
+    d5_positive = d5.status is DimensionStatus.MEASURED and d5.spinout_phrase
+    if d2_fuzzy_positive or d5_positive:
+        primary = "D2" if d2_fuzzy_positive else "D5"
+        corroborated = (
+            (primary != "D3" and d3.status is DimensionStatus.MEASURED and any_ip_link)
```
hollomancer, line 452: **Two problems in the corroboration predicate: one dead branch, and the D4 self-corroboration hole inherited from the frozen spec.**

**Dead branch.** `primary` is assigned on line 449 as `"D2" if d2_fuzzy_positive else "D5"` — never `"D3"`. So `primary != "D3"` is always true, and the remainder (`d3.status is MEASURED and any_ip_link`) is precisely the Order-1 predicate that already returned `SPINOUT_T1` two branches up. The D3 corroboration branch can never fire. `test_order_2_d5_phrase_corroborated_by_d3_ip_link` looks like it covers this, but it builds a `NOT_MEASURABLE` D3 and asserts `INDETERMINATE`, so the dead branch is untested as well as unreachable.

**The D4 hole.** The D4 clause is the only corroborator with no distinctness guard. Since O-1 scopes D2's founders to "D4 Form-D-derived officer/director names only," one Form D officer name can serve as both the D2 fuzzy positive and its D4 corroborator — `SPINOUT_T2` from a single signal, which is the rule Order 2 exists to prevent. This originates in the frozen `design.md` (commented on #620), so fixing it properly means fixing the spec; worth at least not hardening the bug in code first.

---

scripts/sttr_spinout_linkage/kernel.py
```diff
@@ -0,0 +1,486 @@
...
+    supplied here, so the two sub-signals stay the single source of truth."""
+
+    status: DimensionStatus
+    patent_assigned_to_ri_with_sbc_inventor: bool = False
+    recorded_license_ri_to_sbc: bool = False
+    reason: SignalAbsentReason | None = None
+
+
+@dataclass(frozen=True)
+class D4MoneyTrail:
```
hollomancer, line 355: **`D4MoneyTrail` collapses two independently-scored directions into one `status`.**

`design.md`'s D4 row defines two distinct signals with opposite meanings — RI subaward share (a **subcontract** marker) and Form D officer RI-affiliation (a **spinout** marker) — and its typed-absence column lists their absence reasons separately (`NOT_APPLICABLE` / non-grant instrument for the subaward side, Form D absent for the other). One shared `status` field cannot express that.

Confirmed by execution: a contract-instrument STTR (`status=NOT_APPLICABLE`, `NON_GRANT_INSTRUMENT` — correct for the subaward side) that has a genuine RI-affiliated Form D officer plus D2 fuzzy at 0.97 returns `INDETERMINATE` instead of `SPINOUT_T2`, because the Order-2 D4 clause requires `d4.status is MEASURED`. The non-grant instrument suppresses a signal that has nothing to do with subawards.

Suggest splitting into `subaward_status` and `form_d_status`, each with its own `SignalAbsentReason`.

---

scripts/sttr_spinout_linkage/kernel.py
```diff
@@ -0,0 +1,486 @@
...
+
+    Case-folds, strips diacritics and punctuation other than hyphens/
+    apostrophes within a name, and reorders an explicit "Family, Given" input
+    to given-then-family order so a trailing-token family-name assumption
+    holds consistently downstream.
+    """
+
+    text = unicodedata.normalize("NFKD", str(value).strip().lower())
+    text = "".join(ch for ch in text if not unicodedata.combining(ch))
+    if "," in text:
```
hollomancer, line 167: **Any comma is treated as "Family, Given", so credentialed names silently fail D2 matching.**

PI and Form D officer names routinely carry credentials. `"Jane Smith, PhD"` is parsed as family=`"Jane Smith"`, given=`"PhD"` and normalizes to `"phd jane smith"`, which scores **0.838** against `"Jane Smith"` — below the 0.9 cutoff the tests use. The match is lost with no signal that a parse went wrong; it just looks like a person who doesn't match.

Related, same function: `GENERIC_PERSON_TOKENS` is defined and read by the guard but never actually stripped, despite the docstring's "generic-token stripping" claim. `"John Smith Jr."` → `family_name="jr"`; `"Dr. Jane Smith"` → `given_name="dr jane"`.

Both push real pairs below the cutoff, which biases D2 recall in the direction the adjudication sample will later be calibrated against. Suggest stripping credential/title tokens before the comma heuristic, and only reordering when the trailing segment isn't a known generic token.

---

scripts/sttr_spinout_linkage/kernel.py
```diff
@@ -0,0 +1,486 @@
...
+    # Order 0: the join spine itself is incomplete.
+    if not (d1.ri_present and d1.pi_present):
+        return LinkageDecision(
+            LinkageLabel.INDETERMINATE,
+            cascade_order=0,
+            rationale="D1 spine incomplete: RI or PI absent",
+        )
+
+    fuzzy_person = (
+        d2.person_similarity is not None
```
hollomancer, line 425: **`similarity_cutoff` is unvalidated, so the likeliest caller mistake fails open.**

The parameter is deliberately defaultless because O-3 defers the number to a post-1.4 amendment — which makes "caller passes a placeholder to get the code running" the expected failure mode. With `0.0`, this `>=` comparison turns a **measured 0.0 similarity** into a qualifying fuzzy positive, and any corroborator then produces `SPINOUT_T2`.

A `ValueError` for anything outside `(0, 1]` would keep the deferral honest at effectively no cost. Two smaller items in the same area:

- **line 69** — `ORG_GENERIC_TOKENS` inherits `"l.l.c"` and `"l l c"` from `SUFFIX_TOKENS`; neither can equal a whitespace-split token, so `"L.L.C."` → tokens `('l','l','c')` → `guard_passed=True`. The guard misses a name that is purely a legal suffix.
- **line 230** — `identity_similarity` never compares `left.kind` to `right.kind`; an organization identity and a person identity both named `"Smith Robotics"` return 1.0.

---

hollomancer (commented):
Reviewed against the frozen `design.md` (Revision 1) and the `sbir_etl.identity` primitives, executing the module to confirm each finding rather than reading alone. The reuse claim holds up: `normalize_company_name` / `company_name_similarity` / `SUFFIX_TOKENS` are imported and used directly with no forked normalizer, and `similarity_cutoff` being defaultless correctly preserves O-3's deferral.

Five comments covering eight findings. Two change emitted labels and are worth resolving before this merges, because the negative-control and adjudication gates in tasks 1.3/1.4 will be calibrated against whatever this emits:

1. **`generic_token_guard` bypassed on the exact path**, and a guard failure folded into Order 3's measured-negative — a 0.99 match that fails only the guard yields `SUBCONTRACT`. That's the leakage negative control #2 targets.
2. **`D4MoneyTrail` collapses two separately-scored directions**, so a non-grant instrument suppresses the unrelated Form D spinout signal.

Also: a dead D3 corroboration branch with a test that appears to cover it but doesn't; the D4 self-corroboration hole inherited from the frozen spec (commented on #620 — the spec is the right place to fix it); person-name parsing that drops credentialed names below the cutoff; and an unvalidated cutoff that fails open on a placeholder `0.0`.

All exploratory-tier, so this isn't a request for exhaustive hardening — the two above are flagged because they change labels, not because they're untidy.

Happy to push fixes for these directly to this branch if useful.

---

PR #643, opened 2026-08-16
Prototype an evidence-aware SBIR program terminal

hollomancer:
## Description

Adds a dependency-free, exploratory product prototype for an SBIR program decision terminal. The interface demonstrates program metrics, technology portfolio ranking, review signals, an organization screener, universal search, and firm event timelines.

The committed fixture is deliberately synthetic and non-citable. The UI and loader enforce that boundary and expose source/status metadata as the intended contract for a future versioned production export.

No new dependencies are required.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Versioning Impact

- [x] No release impact (internal or unreleased work)
- [ ] PATCH: backward-compatible fix or documentation correction
- [ ] MINOR: new capability, deprecation, or breaking change during `0.x`
- [ ] MAJOR: incompatible public change after `1.0.0`

## Testing

- `node --check tools/sbir-program-terminal/app.js`
- `python3 -m json.tool tools/sbir-program-terminal/data/demo.json`
- direct execution of the new test assertions (the cloud image lacks `uv` and `pytest`)
- browser smoke test at desktop and mobile widths covering period/lens controls, filters, search, keyboard shortcut, and profile dialogs

- [x] Unit tests added/updated
- [ ] Integration tests added/updated
- [x] Focused checks pass locally

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new application warnings
- [x] I have added tests that prove the feature's boundaries
- [x] Focused checks pass locally with my changes

[...]

PR #644, opened 2026-08-16
Add interactive style guide prototype

hollomancer:
## Description

Adds `tools/style-guide/` — a static, dependency-free page documenting the repository's visual language in one place: the network explorer's token system, the matplotlib figure palette, the markdown prose conventions, and the epistemic tier ladder.

The organizing claim is that visual form here encodes **epistemic standing rather than brand**. A dashed edge is not a stylistic variant of a solid one — it means the relationship is a name candidate rather than an identifier-verified observation. Node shape carries record kind, column position carries funding-lineage stage, edge width is log-scaled dollars, and the guardrail component exists specifically to make an interpretive limit impossible to scroll past.

While extracting the values, the more useful finding was that this is not one visual language but **four**, with no shared value between any pair:

| System | Where | Standing |
|---|---|---|
| Material Design | `scripts/data/` — 5 matplotlib scripts, every color inlined at the call site | Contested |
| Explorer tokens | `tools/sbir-dib-network-explorer/` — the only tokenized, dual-theme system | Conventional |
| Flat UI 1.0 | `sbir_etl/utils/reporting/formats/` — legacy HTML reports, light-only | Contested |
| Tailwind gray | `sbir_etl/reporting/procurement_transition/` — a fourth neutral set | Conventional |

So rather than presenting the explorer's palette as though it were house style, the page applies the repository's own discipline to its design system: every convention is labeled **enforced** (a script fails CI when you break it), **conventional** (consistent, but nothing checks it), or **contested** (the codebase actively disagrees with itself). Three labeling rules are machine-enforced: the `**Target epistemic tier:**` declaration (`scripts/ci/check_epistemic_tiers.py`), module `EPISTEMIC_TIER` constants and the import lattice between tiers (`check_tier_boundaries.py`), and reserved Status words in `docs/research-questions.md` (`check_research_question_status.py`). No visual convention is enforced.

### Conflicts recorded

These are surfaced as decisions to be made rather than quietly patched, because picking a winner changes published figures:

- `--technology` and `--critical` resolve to the same value (`#047857` / `#6ee7b7`); a CET-area diamond and a critical-supply-chain review ring are separated only by shape and stroke.
- The per-agency figure panel is orange `#FF9800` in the M&A-signal and Form-D-temporal figures but teal `#00897B` in prime-acquisitions — same role, same position.
- `#2196F3` and `#FF9800` each carry two unrelated meanings: cohort identity in the cohort builder, panel position in three other figures.
- `html_templates.py` renders a *positive* delta red while `html_processor.py` renders good status green, and the spec-status-survey skill instructs new figures to avoid red/green opposition entirely.
- One synced skill (`sbir-lit-refresh`) calls `apply_figure_style()` from a `figure-style` skill that does not exist on disk; `sbir-spec-status-survey` states that skill is not in its loadout and applies the conventions by hand. A shared figure style is already assumed by the tooling and was never written; this page is a candidate specification for it.

### Interactive behavior

Tri-state theme switch (light / auto / dark), click-to-copy tokens, live WCAG contrast recomputed against the current theme, an expandable tier ladder, and a canvas reproducing the explorer's node shapes, column-position lineage encoding, and log-scaled edge weights.

### Declared tier

`exploratory`, stated in the README along with the reason: every value is transcribed by hand, which is the exact failure mode the repository warns about elsewhere — a number that left its source and stopped being checkable. Promotion to `pipelines` means generating the token table from `styles.css` and the conflict list from a linter, so the page fails when the code drifts. Until then the stylesheet remains authoritative.

No existing file is modified; this is additive.

## Type of Change

- [x] Documentation update
- [ ] Bug fix (non-breaking change which fixes an issue)

## Versioning Impact

Adds a standalone static page under `tools/`; no packaged code, config, or public interface changes.

## Testing

Rendered in headless Chromium at 1280px in both light and dark themes, with no console errors. Verified the canvas draws all four node shapes and re-resolves its palette on theme change, the live contrast calculation reports all seven record hues at or above 3:1 against their surface in both themes, the tier ladder expands, and the token rows copy their `var()` reference.

Repository guards run locally:

- `make docs-check` — passed
- `make lint-boundaries` — passed, including `check_epistemic_tiers.py`, `check_tier_boundaries.py`, and `check_file_sizes.py`

- [ ] Unit tests added/updated
- [x] All tests pass locally

No tests added. The artifact is `exploratory` tier and static, and the tier contract does not require them; the meaningful verification is regeneration from source, which is the promotion criterion described above.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

---
_Generated by [Claude Code](https://claude.ai/code/session_01Em9voSGzWZ8dp2uQTm4Br1)_


hollomancer:
CI is red on all four `Fast Tests` shards, but the failure is pre-existing on `main` and not caused by this PR. Flagging rather than fixing, since it belongs to whoever owns the D4 refactor.

**Failure:** `TypeError: D4MoneyTrail.__init__() got an unexpected keyword argument 'status'` — 6 tests in `tests/unit/sttr_spinout_linkage/test_d4_scorer.py::TestScoreD4MoneyTrail`.

**Verified pre-existing.** Checked out `origin/main` at `5d2b750` with none of this PR's changes present and ran the same file:

```
6 failed, 38 passed
```

Identical six failures. This PR adds only `tools/style-guide/README.md` and `tools/style-guide/index.html` — two static files that no Python test imports.

**Root cause**, for whoever picks it up: `D4MoneyTrail` in `scripts/sttr_spinout_linkage/kernel.py:413` was split into two independent statuses, and its docstring explains the intent —

> The two directions carry their own status and absence reason because design.md scores them separately […] A single shared status would let one direction's typed absence suppress the other's measured signal.

so the dataclass now declares `subaward_status` and `form_d_status` with no shared `status` field. But `scripts/sttr_spinout_linkage/d4_scorer.py` still constructs it the old way at several call sites, including `d4_scorer.py:631` and `:616`:

```python
return D4MoneyTrail(status=form_d_status, reason=form_d_reason)
```

`kernel.py` was updated for the split; its caller was not. Both files are `exploratory` tier, so the fix is a caller update rather than anything contract-bound.

I'll re-run once `main` is green.

---

hollomancer:
Pushed `4bd85887` to unblock `Lint, Types, and Guards` again — same cause as `d26f8bef`: `ruff format --check` failing on the terminal test files. Cosmetic only both times (a trailing newline and a quote normalization); no assertions changed. Verified `make lint`, `make lint-boundaries`, and the 6 terminal/exporter tests locally before pushing.

This is the second identical break in ~15 minutes, so flagging the two-second preventative rather than just re-fixing it. Either of these before pushing catches it:

```bash
make format          # or:
uv run ruff format .
```

There's also a `ruff-format` hook already configured in `.pre-commit-config.yaml`, so `pre-commit install` makes it automatic.

Worth noting the repository formats to 941 files clean, so `ruff format --check` is effectively a tripwire on new files only — it'll keep catching each new terminal test file until the hook is installed.

Unrelated and still outstanding: the four `Fast Tests` shards remain red on the pre-existing `D4MoneyTrail` failure described [above](https://github.com/hollomancer/sbir-analytics/pull/644#issuecomment-5307916230), which reproduces on `main` and is not caused by anything in this branch.

---

hollomancer:
Scope split done in `8357bacb`. This PR is the style guide again — net diff against `main` is `tools/style-guide/` alone, 2 files.

**Nothing was lost.** Before removing anything here, the complete current terminal state was cherry-picked onto `main` as [`claude/sbir-program-terminal-split`](https://github.com/hollomancer/sbir-analytics/tree/claude/sbir-program-terminal-split), with original authorship preserved and verified byte-identical — matching blob SHAs across all nine files. `ruff check`, `ruff format --check`, and all 6 terminal/exporter tests pass there.

**#643 is two commits behind.** It tracks the terminal at `fcf903f4`, which predates:

- `Align terminal with visual language prototype`
- `Back terminal with fail-closed capital event export` — the 233-line `scripts/data/export_sbir_program_terminal.py`, which also swapped committed `data/demo.json` for generated data behind a `.gitignore`
- both `ruff format` fixes

The split branch has all of it and is based directly on `main`, so #643 can fast-forward or repoint to it. I deliberately didn't push to `cursor/sbir-terminal-prototype-ef4c` since it isn't my branch.

**On method:** I removed the terminal files in a normal forward commit rather than force-pushing the branch to drop those commits. History here therefore shows the terminal arriving and leaving, which is slightly noisier than a rewrite, but it doesn't rewrite a branch someone else is actively pushing to. Happy to squash it clean instead if you'd prefer.

Unchanged: the four `Fast Tests` shards are still red on the pre-existing `D4MoneyTrail` failure, which reproduces on `main` and is unrelated to either prototype.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1672 @@
...
+<div class="copied" id="copied" role="status" aria-live="polite"></div>
+
+<script>
+  (function () {
+    "use strict";
+
+    var TOKENS = [
+      { name: "--bg", light: "#f5f7f8", dark: "#0c1117", means: "Page ground" },
+      { name: "--surface", light: "#ffffff", dark: "#151c24", means: "Cards, inputs, graph field" },
+      { name: "--surface-raised", light: "#ffffff", dark: "#1b2530", means: "Detail panel, tooltip" },
```
hollomancer, line 1361: **Token name doesn't match the actual CSS custom property — breaks the swatch and the copy-to-clipboard reference for this row.**

The stylesheet declares the surface-raised token as `--raised` (see line 11 / 42 / 70), but this JS entry names it `--surface-raised`, which is never declared anywhere in the CSS.

Concrete failure: on page load, `paintChips()` sets `chip.style.background = "var(--surface-raised)"` for this row's swatch — since that custom property doesn't exist, the background resolves to nothing and the "Detail panel, tooltip" swatch renders with no fill (just its border), unlike every other row. Worse, clicking the row calls `copy("var(" + t.name + ")")`, so a reader trying to use the documented token copies `var(--surface-raised)` to their clipboard — pasting that into real CSS silently resolves to nothing, defeating the entire point of a page whose stated purpose is being an accurate, copy-pasteable source for these tokens.

```suggestion
      { name: "--raised", light: "#ffffff", dark: "#1b2530", means: "Detail panel, tooltip" },
```

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1672 @@
...
+      </div>
+    </section>
+
+    <section id="tokens">
+      <div class="section-head">
+        <h2>Interface tokens</h2>
+        <p>
+          Defined once with the native <code>light-dark()</code> function under
+          <code>color-scheme: light dark</code>. Both values are listed whatever theme you are in,
+          because the pair is the unit of design. Select a row to copy its <code>var()</code> reference.
```
hollomancer: **Prose describes a theming mechanism the page doesn't actually use.**

This paragraph says the tokens are "Defined once with the native `light-dark()` function under `color-scheme: light dark`." Neither `light-dark()` nor `color-scheme` appears anywhere in this file's `<style>` block — the actual mechanism (documented correctly in the CSS comment at the top of the file, lines 3-7) is three separate blocks: a base `:root`, an `@media (prefers-color-scheme: dark)` override, and a `:root[data-theme="dark"]` override, each redefining the same custom-property names with different hex values.

Since this page's entire premise is being an accurate transcription of the codebase's actual visual mechanisms (and it explicitly calls out elsewhere that transcription errors are exactly the failure mode it's trying to avoid), a paragraph that misdescribes its own file's CSS is a real inaccuracy, not just imprecise wording — a reader would come away believing this codebase uses `light-dark()`, which it does not, here or (per the "Four systems" survey) anywhere else in the repo.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1672 @@
+<title>Evidence-Gated Visual Language</title>
```
hollomancer, line 6: **Missing `<!DOCTYPE html>` / `<html>` / `<head>` / `<body>` triggers Quirks Mode when opened the way the README instructs.**

The README tells the reader to serve this with `python -m http.server` or open it directly via `file://`. Without a doctype, browsers render the page in Quirks Mode rather than Standards Mode, which changes box-model and table-layout algorithms (e.g. the `<table>` elements under `.table-wrap`, and percentage/auto sizing in the token/contrast grids) in ways that can differ from what would render in a normal document — and differently across browsers, since Quirks Mode compatibility isn't as tightly specified as Standards Mode. This is lower-confidence than the two token bugs above since `* { box-sizing: border-box }` mitigates the classic box-model discrepancy, but it's still a real deviation from how this page would render if opened normally, and is easy to fix by adding a standard doctype/html/head/body wrapper.

---

hollomancer (commented):
Single-pass review at high effort (Agent tool unavailable, so no multi-agent fan-out or subagent verify pass ran — one pass through all 8 angles in this context). 3 findings, most-severe first. This is a static, additive, exploratory-tier prototype (no packaged code touched), so nothing here blocks merge on its own terms, but the token-name bug undermines the page's stated purpose of being an accurate, usable transcription of the real tokens.

---

hollomancer:
Fresh review pass at head `9eb21695`. **All three earlier findings are genuinely fixed**, and a systematic sweep found no second instance of the token bug — but there is one new blocking issue.

### The three earlier findings — confirmed fixed

**Token name mismatch** — fixed by renaming the *CSS* property to `--surface-raised` (3 declarations at lines 16/47/75, usages at 326/1299) rather than changing the JS. No occurrence of `--raised` remains. That was the better direction: it matches `tools/sbir-dib-network-explorer/styles.css:5`, which this page transcribes.

**False `light-dark()` prose** — fixed. Lines 787-795 now attribute the technique to the source stylesheet and describe what this page actually does (three restated blocks). Verified true in both directions.

**Missing document wrapper** — fixed. Lines 1-5 carry doctype/html/head/meta; a full HTMLParser nesting check over all 1687 lines returns zero mismatched and zero unclosed tags.

**Token audit — clean.** I checked every token reference mechanically rather than by eye: all 18 `TOKENS` rows, 7 `RECORD_TOKENS`, 4 `TIERS[].token`, and 5 `NODES[].token` resolve to a property declared in this page's CSS. Zero undeclared names, and every light/dark pair matches the explorer's `styles.css` byte-for-byte across all 18 tokens.

### Blocking: the emoji-policy claim at lines 1097-1103 is checkably false

> all 168 emoji occurrences across 17 markdown files sit in archive directories or pre-2026 docs. Live steering, spec, and research docs have zero.

The count is right (I measure 169/17). The second sentence is not. Live files carrying emoji include `specs/company-categorization/tasks.md` (6 — listed as **Maintenance** in `specs/status.md:39`), `specs/phase3-match-benchmark/mse-dark-phase3.md` (4) and `cross-agency-scope.md` (2), and `docs/research/dod-form-d-leverage.md:121`. All were last touched 2026-08-12, so "pre-2026" doesn't cover them either.

Either soften to "concentrated in archive directories, with a handful of live specs and research docs still carrying them", or cite the actual exceptions. On a page whose whole thesis is that unverified transcription is the failure mode, a false countable claim is the one defect that undercuts the artifact itself.

### Non-blocking, worth folding in while the file is open

- Same paragraph: "The only sanctioned emoji are emitted by code as a three-value quality scale." The scale is right (`markdown_processor.py:186-191`), but `scripts/performance/detect_performance_regression.py:91` also emits ℹ️ and `scripts/data/run_neo4j_smoke_checks.py:171` emits a two-value ✅/❌.
- The trailing sentence credits archive emoji survival to `.markdownlint.yaml` exclusions. Those `ignoreGlobs` are real, but markdownlint has no emoji rule — the mechanism is invented.
- Line 1064 cites `publication-format.md` by bare filename; it's at `specs/tech-area-transition-report/publication-format.md`.
- Line 1212 paints `#555555` while labeling it `#555`. Cosmetic.
- `renderContrast()` hardcodes `var surface = dark ? "#151c24" : "#ffffff"` (line 1455) instead of reading the token — mildly at odds with line 804's praise for resolving at paint time, though the note does disclose the ground it measured against.
- `ctx.roundRect` (line 1596) has no fallback; on Safari <16.4 / Firefox <112 `drawLineage()` throws. Chips and contrast render first, so degradation is partial. Prototype-acceptable.
- README overstates a claim the HTML gets right: it says both `sbir-lit-refresh` and `sbir-spec-status-survey` "call an `apply_figure_style()` helper". Only the former does — and index.html lines 700-705 already say this correctly.

### Fit with repo conventions

Tier is correct and correctly declared — the README opens with `**Target epistemic tier:** exploratory` and argues the case CLAUDE.md would argue. Note `check_epistemic_tiers.py` only walks `specs/`, so nothing enforces the declaration; stating it anyway is what CLAUDE.md asks. `tools/` is the right home, with direct precedent in `tools/sbir-dib-network-explorer/`. No CI change needed — no Python, and 70 KB is well under `check_file_sizes.py`'s ceiling.

Two conventions it sits outside: nothing ties it to an entry in `docs/research-questions.md` (defensible for a design artifact, but it means the "does this serve a real question" test was never applied), and the notebook-first rule doesn't apply. Neither blocks; both are worth a sentence in the PR body.

**Verdict: mergeable once the emoji claim is corrected.** Given the page is entirely hand-transcribed, it's worth saying that everything else I spot-checked is accurate — the four-system palettes, the `#FF9800`/`#00897B` agency-panel conflict, the red-positive-delta conflict, the mermaid confidence palette, the lineage strip's exact match to `kindDefinitions` at `app.js:58-62`, and the "only `**Target epistemic tier:**` is enforced" claim.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1687 @@
...
+            </div>
+          </div>
+        </div>
+
+        <div class="plate">
+          <h3>Emoji policy <span class="flag conventional">CONVENTIONAL</span></h3>
+          <p class="note">
+            Unwritten but perfectly observed: all 168 emoji occurrences across 17 markdown files sit
+            in archive directories or pre-2026 docs. Live steering, spec, and research docs have
+            zero. The only sanctioned emoji are emitted by code as a three-value quality scale —
```
hollomancer: **The countable claim is false.** The count (168/17) is in the right neighborhood; the next sentence is not. Live files still carrying emoji include `specs/company-categorization/tasks.md` (Maintenance), `specs/phase3-match-benchmark/mse-dark-phase3.md` and `cross-agency-scope.md`, and `docs/research/dod-form-d-leverage.md`. Those were last touched 2026-08-12, so "pre-2026" does not cover them either.

On a page whose thesis is that unverified transcription is the failure mode, this is the defect that undercuts the artifact. Soften to "concentrated in archive directories, with a handful of live specs and research docs still carrying them," or cite the exceptions.

Secondary: markdownlint has no emoji rule, so the trailing sentence invents the enforcement mechanism. And the "only sanctioned emoji" claim misses `scripts/performance/detect_performance_regression.py` (ℹ️) and `scripts/data/run_neo4j_smoke_checks.py` (✅/❌).

hollomancer (commented):
Earlier token/doctype/`light-dark()` findings are fixed and those threads stay resolved. New blocking issue: the emoji-policy paragraph is checkably false. Mergeable once that claim is corrected.

tools/style-guide/index.html
hollomancer: Confirmed on all three counts and fixed in `7ee2520a`. Recounted from scratch rather than trusting either number.

**The count.** 169 occurrences across 17 markdown files — so 168/17 was off by one, as you said, in the right neighborhood.

**The claim after it.** 9 of those 17 files are live and carry **68 of the 169 — 40 percent**, not zero. The distribution:

| Occurrences | File |
|---|---|
| 42 | `docs/architecture/duckdb-cet-analysis.md` |
| 6 | `specs/company-categorization/tasks.md` |
| 6 | `docs/development/exception-handling.md` |
| 3 | `specs/phase3-match-benchmark/mse-dark-phase3.md` |
| 3 | `docs/development/logging-standards.md` |
| 3 | `docs/architecture/asset-naming-standards.md` |
| 2 | `specs/phase3-match-benchmark/cross-agency-scope.md` |
| 2 | `docs/nano-phase3-methodology.md` |
| 1 | `docs/research/dod-form-d-leverage.md` |

Your four examples are all in there, plus five more you did not have to name. `duckdb-cet-analysis.md` alone falsifies "Live steering, spec, and research docs have zero" by more than the entire non-archive remainder.

**Your two secondary points, both confirmed and both worse than "secondary".** `.markdownlint.yaml` has no emoji rule — its `ignoreGlobs` exclude archives from linting *generally*, which is not why archives keep their emoji, and nothing enforces the convention anywhere. And the "only sanctioned emoji" claim: **18** live Python files print emoji, and the glyph set is `ℹ️ ⚠️ ✅ ❌ 📊 🔍`, not the three of the quality scale. You named two of those files; there are sixteen more.

Rewrote the plate to state the archive/live split, name the live files, and say outright that nothing enforces this. I took your framing as the standard rather than softening to "a handful": on a page arguing that unverified transcription is the failure mode, a hedge would still be a claim nobody checked.

---

hollomancer:
Re-checked at `7ee2520a` and pushed a follow-up (`4ce2261c`). Your emoji rewrite fixed the totals, but **the enumeration is still wrong in the same way**, so I corrected that plus three transcription slips.

**The residual defect.** The sentence said the named files "account for the rest". They don't. Counting only unambiguous ✅/❌/emoji-plane characters, the named set — `duckdb-cet-analysis.md` (42), `company-categorization/tasks.md` (6), `phase3-match-benchmark/` (3+2), `dod-form-d-leverage.md` (1) — totals **54**, not the 66–68 the same sentence claims for the live set. Three live files carry the missing 12 and were unnamed:

- `docs/development/exception-handling.md` (6)
- `docs/development/logging-standards.md` (3)
- `docs/architecture/asset-naming-standards.md` (3)

Those sit in exactly the `docs/development/` and `docs/architecture/` trees the *original* claim said were clean — so omitting them understated the point the plate is making. The paragraph now names them and drops the "account for the rest" framing.

On your headline numbers: I measure 167 across 16 files (66 live in 8) versus your 169/17 and 68/9. That gap is a counting-rule artifact — whether `⚠️`/`ℹ️` variation selectors and dingbats like `✗` count — and I left your totals alone rather than churn on ±2. The enumeration was wrong under *any* rule, which is why I only changed that part.

**Three other slips, all in the "unverified transcription" category the page is about:**

- The null-result swatch painted `#555555` while labelling `#555`. The source at `docs/sbir-identification-methodology.md:444` is `fill:#555`, so your label was right — the swatch now paints what it documents, matching the two rows above it.
- `publication-format.md` was cited by bare filename; it's at `specs/tech-area-transition-report/publication-format.md`.
- The README said *two* synced skills call `apply_figure_style()`. Only `sbir-lit-refresh` does (`SKILL.md:165-166`); `sbir-spec-status-survey:213` states `figure-style` is **not** in its loadout and applies the conventions by hand — which is what index.html lines 700-705 already said correctly.

Verified after the edit: all tags balanced, zero unclosed elements, and every `var(--token)` reference still resolves to a declared custom property.

Also confirmed your earlier fixes are genuine, not papered over: the `--surface-raised` rename went the right direction (matching `styles.css:5`), the `light-dark()` prose now describes what the page actually does, and the document wrapper is complete.

That clears the blocker I raised. From my side this is mergeable.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1696 @@
...
+    function copy(text) {
+      if (navigator.clipboard && navigator.clipboard.writeText) {
+        navigator.clipboard.writeText(text).then(
+          function () { flash("Copied " + text); },
+          function () { flash(text); }
+        );
+      } else {
+        flash(text);
+      }
+    }
```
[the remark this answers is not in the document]
hollomancer: Fixed in c4e09195: `legacyCopy` selects the text in an off-screen textarea and runs `document.execCommand("copy")`, and the toast says "Copy failed" when neither path works. Also merged `main` (36 commits behind) so CI runs on a current base; the previous head had no CI run at all, only CodeQL.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1696 @@
...
+      var lb = luminance(hexToRgb(b));
+      return (Math.max(la, lb) + 0.05) / (Math.min(la, lb) + 0.05);
+    }
+
+    var contrastBox = document.getElementById("contrast-list");
+
+    function renderContrast() {
+      var dark = isDark();
+      var surface = dark ? "#151c24" : "#ffffff";
+      contrastBox.textContent = "";
```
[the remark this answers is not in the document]
hollomancer: Fixed in c4e09195: `renderContrast` reads `cssVar("--surface")`; the two literals stay only as a fallback for a stylesheet that fails to load.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1696 @@
...
+      <div class="plate">
+        <div class="tok-head"><span></span><span>Token</span><span>Light</span><span>Dark</span><span>Means</span></div>
+        <div class="tokens" id="token-list"></div>
+      </div>
+
+      <p class="note" style="margin-top: 1rem;">
+        The canvas never hardcodes these. It resolves them at paint time by mounting a hidden probe
+        span, reading <code>getComputedStyle().color</code>, and discarding it — so the graph
+        re-themes with the OS and the stylesheet stays the single source of truth. It is the best
+        pattern in the repository and nothing else uses it.
```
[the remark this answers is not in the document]
hollomancer: Fixed in c4e09195: the note now says the page reads each custom property with `getComputedStyle(document.documentElement).getPropertyValue()` through the `cssVar` helper, which is what the canvas code does.

---

hollomancer:
Fresh review of head `c4e09195`, rechecking every transcription claim against the files it cites, then pushed `bbcf3ff`.

**Seven claims were wrong on this head and are corrected:**
- "Two synced skills instruct agents to call `apply_figure_style()`": only `sbir-lit-refresh` does; `sbir-spec-status-survey` says `figure-style` is not in its loadout. Both are user-level synced skills, not files in this repository.
- "This is the only piece of the visual language a machine will stop you from breaking": `check_tier_boundaries.py` enforces module `EPISTEMIC_TIER` constants and the import lattice, and `check_research_question_status.py` enforces reserved Status words. Nothing runs `.markdownlint.yaml`.
- "Line 3 of a spec": `check_epistemic_tiers.py` uses a multiline regex over the whole file.
- Weight 650 "appears seven times": five; the other hits are `min-height: 650px`.
- "1rem/700 with tabular figures": the stylesheet sets no `font-variant-numeric`.
- Figure conventions stated as universal: `nano_form_d_temporal.py` uses `(12, 5)` and alpha `0.8`; `nano_survival_analysis.py` never closes its figure.
- "Material Design 500/700/800": `#00897B` is Teal 600.

The emoji paragraph now states its counting rule and its numbers are recomputed from this head under that rule. The in-flight table is dated 2026-08-16 and notes all four PRs merged on 2026-08-18.

**Verified clean:** all 18 token rows match `styles.css` in name, light, and dark value; the four-systems locations and every named color conflict trace to the cited files and lines; headless Chromium render in both themes has zero console errors, the canvas paints, and the contrast table populates.

The PR body's "Exactly one convention is currently enforced" sentence is edited to match.

---

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+          <p class="note">
+            Node level, tinted <code>--supplier-soft</code> with a solid <code>--entity</code> rule.
+            Rendered only when a specific record has an evidentiary caveat, so its presence is itself
+            information.
+          </p>
+        </div>
+
+        <div class="plate stack">
+          <div>
+            <h3>Controls</h3>
```
hollomancer: **[bug]** The typography specimens do not match the explorer stylesheet they claim to document, and two of them contradict their own meta labels. Explorer `h3` is `0.9rem`, `letter-spacing: 0.06em`, and not monospace; this specimen is `0.72rem` / `0.09em` / `var(--mono)` / `color: var(--muted)` (the style guide’s own chrome `h3`), while the meta still says “tracking 0.06em”. The `.eyebrow` specimen uses this page’s class (`0.68rem`, mono) while claiming explorer’s `0.74rem` (explorer `.eyebrow` is also not mono). The `h1`/`h2` specimens apply `font-weight: 650`, but explorer headings do not set 650 — that weight is used five times on interface labels, which the following plate correctly states.

**Suggestion:** Restyle each specimen with the explorer values (inline, as the `h1`/`h2` size/tracking already attempt), drop 650 from headings, and make each `.meta` line match the rendered CSS. Point the eyebrow specimen at explorer’s `0.74rem` sans stack rather than this page’s `.eyebrow` class.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+            Not a rounding of 600 or 700. It appears five times in the shipped stylesheet, always on
+            interface labels that must out-rank body text without reading as a heading. Inter carries
+            it as a variable axis; the fallback stack degrades to 700, which is acceptable — the
+            hierarchy survives, only the finesse is lost. No webfont is loaded, by design: the
+            explorer is dependency-free and must render offline from a local
+            <code>http.server</code>. Matplotlib figures set no font family at all and render in
+            DejaVu&nbsp;Sans, so charts and interface do not currently share a typeface.
+          </p>
+        </div>
+      </div>
```
hollomancer: **[bug]** The “Bar alpha” row claims `0.8` in `nano_form_d_temporal.py`. That file’s bars are still `alpha=0.85`; `0.8` is only the left-panel histogram (`ax.hist(..., alpha=0.8)`). The other four matplotlib scripts’ bars are `0.85`.

**Suggestion:** Either drop the parenthetical exception, or relabel the row so the `0.8` is attributed to the histogram fill rather than bar alpha.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+    padding: 0.5rem 0.7rem 0.5rem 0;
+    border-bottom: 1px solid var(--border);
+    color: var(--muted);
+    font-family: var(--mono);
+    font-size: 0.63rem;
+    font-weight: 700;
+    letter-spacing: 0.08em;
+    text-align: left;
+    text-transform: uppercase;
+    white-space: nowrap;
```
hollomancer: **[bug]** The metric-pair specimen copies explorer `.detail-metrics dd` (`1rem` / `font-weight: 700`) but adds `font-variant-numeric: tabular-nums`. Explorer `styles.css` does not set that property — which is exactly what the note under the specimen says, “so columns of dollars align only where the digits happen to be the same width.” With tabular nums on the specimen, the documented limitation is not visible.

**Suggestion:** Remove `font-variant-numeric` from `.detail-metrics dd` (keep it on this page’s own `.hex` / `.ratio` / `.readout` if desired) so the component matches the explorer rule the note describes.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+            hierarchy survives, only the finesse is lost. No webfont is loaded, by design: the
+            explorer is dependency-free and must render offline from a local
+            <code>http.server</code>. Matplotlib figures set no font family at all and render in
+            DejaVu&nbsp;Sans, so charts and interface do not currently share a typeface.
+          </p>
+        </div>
+      </div>
+    </section>
+
+    <section id="components">
```
hollomancer: **[suggestion]** “Grid / spines Untouched” is overstated. `nano_survival_analysis.py` calls `ax.grid(alpha=0.3)` on the Kaplan–Meier figure. No seaborn, despine, or font-family setting is still true for the five matplotlib scripts.

**Suggestion:** Qualify the grid claim (e.g. “no seaborn/despine; `nano_survival_analysis.py` enables a default grid”) or drop “untouched.”

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+  .tok, .tok-head {
+    display: grid;
+    grid-template-columns: 2.1rem minmax(0, 1.3fr) repeat(2, minmax(0, 0.75fr)) minmax(0, 1.5fr);
+    gap: 0.7rem;
+  }
+
+  .tok {
+    align-items: center;
+    width: 100%;
+    padding: 0.48rem 0.6rem;
```
hollomancer: **[suggestion]** Light-theme `.flag.contested` is `#b45309` on `#fae8d5`, about 4.20:1. The flags are `0.61rem` (~10px), so WCAG AA’s 4.5:1 normal-text bar applies; CONTESTED is used heavily on this page. Dark theme of the same pair is fine (~8:1).

**Suggestion:** Darken the light-theme contested foreground (or use `--text` on `--supplier-soft`) until the pair clears 4.5:1.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+    function luminance(rgb) {
+      var a = rgb.map(function (v) {
+        var s = v / 255;
+        return s <= 0.03928 ? s / 12.92 : Math.pow((s + 0.055) / 1.055, 2.4);
+      });
+      return 0.2126 * a[0] + 0.7152 * a[1] + 0.0722 * a[2];
+    }
+
+    function ratio(a, b) {
+      var la = luminance(hexToRgb(a));
```
hollomancer: **[suggestion]** The second `legacyCopy` comment restates the following statements (`createElement("textarea")`, `select()`, `execCommand("copy")`) instead of adding a constraint.

**Suggestion:** Keep the why (“non-secure contexts have no `navigator.clipboard`”) and drop the line that narrates the textarea fallback.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+      var dark = isDark();
+      // Read the live token so the ratios track the stylesheet; the literals
+      // are only a fallback for a stylesheet that fails to load.
+      var surface = cssVar("--surface") || (dark ? "#151c24" : "#ffffff");
+      contrastBox.textContent = "";
+
+      RECORD_TOKENS.forEach(function (name) {
+        var t = TOKENS.find(function (x) { return x.name === name; });
+        if (!t) return;
+        var value = dark ? t.dark : t.light;
```
hollomancer: **[suggestion]** The contrast section is titled “measured live” and the comment at 1488–1489 says hues are read from the stylesheet with literals only as fallback, but only `--surface` goes through `cssVar()`. Record hues are taken from the hardcoded `TOKENS` array (`dark ? t.dark : t.light`). Token chips already paint with `var(--name)`, so editing `:root` updates the table swatches while the PASS/BELOW ratios and the generated “Measured against --surface #…” note stay on the JS copy.

**Suggestion:** Resolve each contrast hue with `cssVar(name)` (keep `t.light`/`t.dark` only as fallback), and quote the live surface hex in the note rather than `#151c24` / `#ffffff`.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+          <div class="type-spec">
+            <div style="font-size: 1.25rem; line-height: 1.25; font-weight: 650;">Select an organization</div>
+            <p class="meta">h2 · 1.25rem / 1.25</p>
+          </div>
+          <div class="type-spec">
+            <div style="font-family: var(--mono); font-size: 0.72rem; font-weight: 700; letter-spacing: 0.09em; text-transform: uppercase; color: var(--muted);">Evidence and interpretation</div>
+            <p class="meta">h3 · uppercase · tracking 0.06em</p>
+          </div>
+          <div class="type-spec">
+            <p class="eyebrow">Evidence-gated funding lineage</p>
```
hollomancer: **[suggestion]** Typography specimens are captioned with the explorer scale but do not render it. `.eyebrow` on this page is 0.68rem (line 143) while the caption says 0.74rem, which is the explorer value; the specimen uses the page class so it cannot match. The h3 specimen (line 914) is inline-styled at 0.72rem / tracking 0.09em (this page’s `h3`) while its caption says tracking 0.06em (explorer `h3` is 0.9rem / 0.06em).

**Suggestion:** Give explorer specimens dedicated classes (or inline the explorer values) so chrome for this page can keep its denser rail type without lying about the source scale.

tools/style-guide/index.html
```diff
@@ -0,0 +1,1722 @@
...
+      var pts = NODES.map(function (n) {
+        return { x: pad + n.column * (w - pad * 2), node: n };
+      });
+
+      for (var i = 0; i < pts.length - 1; i++) {
+        var last = i === pts.length - 2;
+        ctx.beginPath();
+        ctx.moveTo(pts[i].x, y);
+        ctx.lineTo(pts[i + 1].x, y);
+        ctx.strokeStyle = last && candidate ? cssVar("--candidate") : cssVar("--edge");
```
hollomancer, line 1722: **[suggestion]** The lineage note documents two independent encodings — log width and `0.34 + min(0.36, fiscal_years × 0.07)` opacity — but `drawLineage()` never sets `globalAlpha` and there is no fiscal-year control. Width and dash are interactive; persistence is not shown. In the explorer both channels are applied on the same stroke.

**Suggestion:** Apply the explorer alpha on the edges (even with a fixed fiscal-year sample) or drop the opacity formula from the canvas caption so the demo only claims what it draws.

tools/style-guide/README.md
```diff
@@ -0,0 +1,61 @@
...
+Static, dependency-free browser page documenting the repository's visual
+language: the network explorer's token system, the matplotlib figure palette,
+the prose conventions, and the epistemic tier ladder.
+
+## Run locally
+
+From this directory:
+
+```bash
+python -m http.server 8080
```
hollomancer: **[suggestion]** `python -m http.server 8080` is the same command and port as `tools/sbir-dib-network-explorer/README.md`. Running both locally fails on bind. This page also opens from `file://`, so the collision is unnecessary for the documented happy path.

**Suggestion:** Pick another port (or say “any free port”) and note that `file://` is enough when no server is wanted.

hollomancer (commented):
## Summary

This is a self-contained exploratory prototype and a reasonable neighbor to `tools/sbir-dib-network-explorer/`. It fetches nothing, interpolates no untrusted data (`textContent` / `createElement` only), and the claimed explorer tokens, edge formulas, matplotlib hexes, HTML-report red/green split, and emoji census all match the current tree. The remaining issues are demo/copy mismatches inside the page itself — contrast and type specimens do not fully track the sources they describe — plus a local-server port collision with the explorer README.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

hollomancer:

### Status: 🟡 Close to Merge-Ready — 4 Suggestions

- **Draft:** No
- **Age:** 24 days — **oldest open PR**
- **Size:** +1,783/-0 across 2 files

Adds a self-contained, dependency-free interactive style guide prototype under `tools/style-guide/`.

**History:** 
- Initially blocked on token-name accuracy (2026-08-18) — **resolved**
- Then blocked on emoji-policy paragraph being checkably false (2026-08-31) — **resolved**
- Copilot: 🟡 Changes recommended (copy-to-clipboard fallback, contrast background source) — **status unclear**

**Most recent hollomancer review (2026-09-10): 0 bugs, 4 suggestions:**
> *"The remaining issues are demo/copy mismatches inside the page itself — contrast and type specimens do not fully track the sources they describe — plus a local-server port collision with the explorer README."*

All original blockers are cleared. The remaining items are suggestions, not bugs. This is close to merge-ready.

**Recommended action:** Address the 4 suggestions (contrast/type specimen accuracy, port collision), then merge. This is the longest-open PR at 24 days and would be a good candidate to close.

hollomancer:
## Consolidated review — small responsive fix requested

Reviewed at `bbcf3ffe`.

The desktop implementation looks good: light/dark/auto themes, live contrast, token copy, tier expansion, canvas interaction, and controls all worked without console errors.

- **Important — the mobile breakpoint still overflows horizontally.** At a 390px viewport, `document.documentElement.scrollWidth` is 600px. `.grid-2` at `tools/style-guide/index.html:255` uses `minmax(290px, 1fr)` inside padded plates, while the `max-width: 880px` media query does not constrain grid children. The prose plates expand to roughly 581px, and fixed 11px lineage-canvas labels overlap until the diagram is unreadable.

Suggested fix: use `minmax(min(290px, 100%), 1fr)`, add `min-width: 0` to grid/plate children, and provide a mobile canvas treatment (label truncation/wrapping or an alternate stacked legend). A 390px regression check would close this out.

tools/style-guide/README.md
hollomancer: Fixed in ee608be4. README now leads with `file://` (enough, the page fetches nothing) and documents `python -m http.server 8765` as a suggestion. Any free port works; 8080 stays the explorer's.

tools/style-guide/index.html
hollomancer: Fixed in ee608be4. Specimens use `.explorer-h1` / `.explorer-h2` / `.explorer-h3` / `.explorer-eyebrow` matching `tools/sbir-dib-network-explorer/styles.css` (h3 0.9rem / tracking 0.06em, eyebrow 0.74rem / weight 700). Page chrome `.eyebrow` and `h3` stay denser.

tools/style-guide/index.html
hollomancer: Fixed in ee608be4. `renderContrast` now resolves each record hue with `cssVar(name)` (`t.light`/`t.dark` only as fallback), paints the swatch with `var(--name)`, and quotes the live `--surface` hex in the generated note.

tools/style-guide/index.html
hollomancer: Fixed in ee608be4. `drawLineage` sets `globalAlpha` to `0.34 + min(0.36, fiscal_years × 0.07)` on every stroke, same formula as the explorer. A fiscal-years control (1–6) sits next to the obligation slider so both channels are interactive.

hollomancer:
Also in ee608be4: the 390px overflow is gone. `.grid-2` is now `minmax(min(290px, 100%), 1fr)` with `min-width: 0` on grid/plate children; the 880px query stacks grids and the canvas uses short labels. Headless Chrome at a 390px viewport: `document.documentElement.scrollWidth` is 390 (was 600).

hollomancer:
Closing this hand-transcribed prototype because its inventory already disagrees with current main. The design concept remains useful, but a successor should generate tokens and conflict inventories from the authoritative explorer stylesheet and current report generators so drift is detectable rather than documented manually.

PR #645, opened 2026-08-16
Report each D4 direction on its own status

hollomancer:
## Description

Fixes a `TypeError` on `main` that fails 6 tests and turns the `Fast Tests` shards red on **every open PR in the repository**.

`score_d4_money_trail` still constructed `D4MoneyTrail` with a single `status=`/`reason=` pair, but the dataclass in `scripts/sttr_spinout_linkage/kernel.py` had been split into `subaward_status`/`form_d_status` with per-direction reasons. `kernel.py` was updated for the split; its caller was not:

```
TypeError: D4MoneyTrail.__init__() got an unexpected keyword argument 'status'
```

### Why this isn't a mechanical rename

Translating `status=X` into `subaward_status=X, form_d_status=X` would compile and go green while preserving the exact defect the split was made to remove.

`kernel.classify_linkage` already reads the two directions independently — Order 2's corroboration checks `form_d_status is MEASURED and form_d_officer_ri_affiliated`, Order 3 checks `subaward_status is MEASURED and ri_subaward_share > 0`. The scorer's four-step "combining policy" collapsed both into one status, which is what the dataclass docstring warns against:

> A single shared status would let one direction's typed absence suppress the other's measured signal.

The scorer's own docstring already argued the same point, describing the collapse as a workaround for a kernel it noted was "not modified by this PR". The kernel has since been modified, so the workaround is obsolete. Each direction is now passed straight through and the combining policy is deleted.

### Behavior changes

Two cases get strictly more accurate:

- **Contract-instrument award with a Form D officer match** now reports `subaward_status = NOT_APPLICABLE` / `NON_GRANT_INSTRUMENT` *and* `form_d_status = MEASURED`. Previously the whole trail was relabelled `MEASURED`, erasing the fact that the subaward direction was never applicable.
- **An evaluation failure on one direction no longer discards the other.** The old `return D4MoneyTrail(status=EVALUATION_FAILED)` dropped `ri_subaward_share` and `form_d_officer_ri_affiliated` entirely; both are now retained.

Net effect is a 74→50 line reduction in the scorer — the policy was the bulk of it.

### Tests

Updated to assert per-direction status and reason. Two were renamed because the behavior they named no longer exists:

| Before | After |
|---|---|
| `test_pi_name_only_match_drives_the_combined_trail` | `test_pi_name_only_match_is_measured_on_the_form_d_direction` |
| `test_neither_direction_evaluated_prefers_the_more_specific_status` | `test_each_direction_reports_its_own_typed_absence` |

Every test's original intent is preserved. `test_contract_instrument_with_form_d_match_is_measured_not_suppressed` — the case the module exists to protect — gets stronger: no-suppression is now structural rather than dependent on a policy branch ordering.

Both `scripts/sttr_spinout_linkage/` and its tests are `exploratory` tier, so nothing contract-bound is affected. No production module imports the scorer.

## Type of Change

- [ ] Documentation update

## Versioning Impact

`exploratory`-tier code under `scripts/`, not reachable from any packaged module or public interface.

## Testing

Confirmed the failure reproduces on `main` at `5d2b750b` before changing anything (6 failed, 38 passed in the affected file), then verified the fix:

- `make test-unit` — **6241 passed, 7 skipped, 0 failed** (was 6 failed)
- `tests/unit/sttr_spinout_linkage/` — 141 passed, including the 11 `classify_linkage` cases in `test_kernel.py` that already used the two-status form
- `make lint` — ruff check, `ruff format --check` (939 files), UP042, and mypy over 316 source files all clean
- `make lint-boundaries` — all seven guards pass

Also grepped for other consumers of the old contract: the only production construction site is the one fixed here, and the remaining `trail.status` references belong to `D2PersonTrail` and `D5TextTrail`, which legitimately still carry a single status.

- [ ] Unit tests added/updated
- [x] All tests pass locally

Existing tests updated rather than added; the file already covered these cases and could not compile against the current dataclass.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01Em9voSGzWZ8dp2uQTm4Br1)_

### DIFF
--- a/scripts/sttr_spinout_linkage/d4_scorer.py
+++ b/scripts/sttr_spinout_linkage/d4_scorer.py
@@ -567,31 +567,23 @@ def score_d4_money_trail(
     subaward_client: SubawardClient | None = None,
     form_d_index: dict[str, tuple[str, ...]] | None = None,
 ) -> D4MoneyTrail:
-    """Score both D4 directions and combine them into one `D4MoneyTrail`.
+    """Score both D4 directions and report each one on its own status.
 
-    `kernel.D4MoneyTrail` (prerequisite code, not modified by this PR) has a
-    single `status`/`reason` pair covering both directions, but
-    `kernel.classify_linkage` reads each direction's own field
-    (`ri_subaward_share`, `form_d_officer_ri_affiliated`) independently --
-    Order 3 checks `status == MEASURED and ri_subaward_share > 0`, Order 2's
-    corroboration checks `status == MEASURED and form_d_officer_ri_affiliated`.
-    Setting the shared `status` from only one direction would wrongly block
-    the cascade from crediting the *other* direction's real result (e.g. a
-    contract-instrument award, correctly `NOT_APPLICABLE` for the subaward
-    share, must not also suppress a real Form D officer match) -- exactly
-    the class of directional bug this module's docstring opens with.
+    `kernel.D4MoneyTrail` carries a separate status and absence reason per
+    direction, and `kernel.classify_linkage` reads them independently:
+    Order 2's corroboration checks `form_d_status is MEASURED and
+    form_d_officer_ri_affiliated`, Order 3 checks `subaward_status is
+    MEASURED and ri_subaward_share > 0`. Each direction is therefore passed
+    straight through, with no attempt to reconcile the two into a single
+    shared status.
 
-    Combining policy, in priority order:
-    1. `MEASURED` if **either** direction produced a real measured value --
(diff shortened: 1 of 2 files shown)


hollomancer:
Closing as a duplicate — this same fix (splitting `D4MoneyTrail` construction into independent `subaward_status`/`form_d_status` and dropping the obsolete combining policy) landed on `main` via #647, which picked it up incidentally while getting its own CI green. No action needed here; thanks for catching and fixing this independently.

---

PR #654, opened 2026-08-18
Require study contracts for reserved inventory Status ranks

hollomancer:
## Why

The research-question inventory was a public API without a backstop. Status lines could say **Computable** (or imply an answer) from a research note, and the policymaker start-here box sent readers to A3 and D2 — neither of which has a study contract or a Status line a staffer can trust.

## What changed

- Reserved ranks in `docs/research-questions.md` now map to study `evidence_status`:
  - `Computable` / `Partially computable` → `reproducible` or higher
  - `Validated` → `validated` or higher
  - `Citable` → `citable`
- Exploratory studies do not authorize `Computable`. Negations (`Not computable`, `non-citable`) stay legal without a study.
- CI and `make lint-boundaries` run `scripts/ci/check_research_question_status.py`.
- Status lines that used reserved ranks without a study were rewritten as exploratory / inventory-target prose.
- The B2/B3/E1 census questions keep `Computable` / `Partially computable` because `studies/phase-iii-census` is `reproducible`.
- Policymakers now start at B2/B3 (study-backed). A3 and D2 are labeled inventory targets and stay out of that box until they have study-backed Status.

## Out of scope

- Promoting Form D (separate PR).
- Census memo drift (separate PR).
- Changing any study `evidence_status`.

### DIFF
--- a/scripts/ci/check_research_question_status.py
+++ b/scripts/ci/check_research_question_status.py
@@ -3,9 +3,14 @@
 
 ``docs/research-questions.md`` is a public API. The reserved ranks
 ``computable``, ``validated``, and ``citable`` may appear as Status claims only
-when a ``studies/*/study.yaml`` lists that question at the matching
-``evidence_status`` (or higher). Negations (``not computable``, ``non-citable``)
-are allowed without a study: they are refusals, not ranks.
+when a ``studies/*/study.yaml`` lists that *section* ID (``B2``, ``F3``, …)
+at the matching ``evidence_status`` (or higher). Authorization is per section,
+not per question bullet: a study listing ``B2`` authorizes every reserved-rank
+Status under ``### B2``.
+
+Negations (``not computable``, ``never computable``, ``non-citable``) are
+refusals, not ranks, and do not need a study. The verb ``validates`` is not
+the ``validated`` rank.
 
 This is admission control for the inventory, not proof that a study's result
 is correct. Exploratory studies do not authorize ``computable``.
@@ -24,31 +29,29 @@ from sbir_etl.quality.study_manifest import EvidenceStatus, load_study_manifest
 
 REPOSITORY_ROOT = Path(__file__).resolve().parents[2]
 INVENTORY_PATH = Path("docs/research-questions.md")
-STUDIES_ROOT = REPOSITORY_ROOT / "studies"
 
+ANY_HEADING = re.compile(r"^#{2,4}\s+")
 SECTION_HEADING = re.compile(r"^#{2,4}\s+([A-F]\d+)\b")
 STATUS_MARKER = re.compile(r"\*\*Status:\*\*")
 STATUS_END = re.compile(r"^(\s*\*Deps:|#{1,6}\s|-\s+\*\*)")
 
 # Strip these before looking for a positive rank so denials do not count.
+# A short negation window covers "not currently computable" / "never
+# computable" without exempting the noun phrase "citable claim".
 DENIAL_PHRASES = (
-    re.compile(r"\bnot\s+yet\s+computable\b", re.IGNORECASE),
-    re.compile(r"\bnot\s+computable\b", re.IGNORECASE),
-    re.compile(r"\bnon-computable\b", re.IGNORECASE),
-    re.compile(r"\bnot\s+yet\s+validated\b", re.IGNORECASE),
-    re.compile(r"\bnot\s+validated\b", re.IGNORECASE),
+    re.compile(
+        r"\b(?:not|no|never|cannot)\W+(?:\w+\W+){0,3}?(?:computable|validated|citable)\b",
+        re.IGNORECASE,
+    ),
+    re.compile(r"\bnon-(?:computable|validated|citable)\b", re.IGNORECASE),
     re.compile(r"\bunvalidated\b", re.IGNORECASE),
-    re.compile(r"\bremains\s+unvalidated\b", re.IGNORECASE),
     re.compile(r"\bnot\s+approved\s+for\s+citation\b", re.IGNORECASE),
-    re.compile(r"\bnot\s+citable\b", re.IGNORECASE),
-    re.compile(r"\bnon-citable\b", re.IGNORECASE),
-    re.compile(r"\bcitable\s+study(?:\s+contract|\s+manifest)?\b", re.IGNORECASE),
-    re.compile(r"\bcitable\s+claim\b", re.IGNORECASE),
+    re.compile(r"\bany\b[^.]*\bcitable\b[^.]*\bblocked\b", re.IGNORECASE),
 )
 
 RANK_PATTERNS: tuple[tuple[str, re.Pattern[str]], ...] = (
     ("computable", re.compile(r"\bcomputable\b", re.IGNORECASE)),
-    ("validated", re.compile(r"\bvalidat(?:ed|es)\b", re.IGNORECASE)),
+    ("validated", re.compile(r"\bvalidated\b", re.IGNORECASE)),
     ("citable", re.compile(r"\bcitable\b", re.IGNORECASE)),
 )
 
@@ -111,9 +114,9 @@ def iter_status_blocks(markdown: str) -> Iterable[tuple[int, str | None, str]]:
     index = 0
     while index < len(lines):
         line = lines[index]
-        heading = SECTION_HEADING.match(line)
-        if heading:
-            section_id = heading.group(1).upper()
+        if ANY_HEADING.match(line):
+            heading = SECTION_HEADING.match(line)
+            section_id = heading.group(1).upper() if heading else None
         marker = STATUS_MARKER.search(line)
         if marker is None:
             index += 1
@@ -153,7 +156,11 @@ def collect_status_claims(markdown: str) -> list[StatusClaim]:
 def load_question_study_ranks(
     *, repository_root: Path = REPOSITORY_ROOT
 ) -> dict[str, EvidenceStatus]:
-    """Map each study-listed question ID to the highest live evidence status."""
+    """Map each study-listed section ID to the highest live evidence status.
+
+    A study that lists ``B2`` authorizes reserved ranks for every Status
+    block under ``### B2``, not only the bullet whose estimand it covers.
+    """
 
     best: dict[str, EvidenceStatus] = {}
     for path in sorted((repository_root / "studies").glob("*/study.yaml")):
--- a/scripts/ci/check_research_question_status.py
+++ b/scripts/ci/check_research_question_status.py
@@ -8,9 +8,10 @@ at the matching ``evidence_status`` (or higher). Authorization is per section,
 not per question bullet: a study listing ``B2`` authorizes every reserved-rank
 Status under ``### B2``.
 
-Negations (``not computable``, ``never computable``, ``non-citable``) are
-refusals, not ranks, and do not need a study. The verb ``validates`` is not
-the ``validated`` rank.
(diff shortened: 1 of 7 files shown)


scripts/ci/check_research_question_status.py
```diff
@@ -0,0 +1,249 @@
...
+    re.compile(r"\bnon-computable\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+yet\s+validated\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+validated\b", re.IGNORECASE),
+    re.compile(r"\bunvalidated\b", re.IGNORECASE),
+    re.compile(r"\bremains\s+unvalidated\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+approved\s+for\s+citation\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+citable\b", re.IGNORECASE),
+    re.compile(r"\bnon-citable\b", re.IGNORECASE),
+    re.compile(r"\bcitable\s+study(?:\s+contract|\s+manifest)?\b", re.IGNORECASE),
+    re.compile(r"\bcitable\s+claim\b", re.IGNORECASE),
```
hollomancer: These two denial patterns strip `citable` from *any* Status text, including affirmative claims, so the rank this PR exists to gate can be asserted with no study at all.

Verified against this branch:

```
claimed_ranks("Citable claim: the DoD follow-on multiplier is 4.1:1.")   -> ()
claimed_ranks("Citable study result approved for external reporting.")  -> ()
```

So `**Status:** Citable claim: ...` under A3 (no study contract at all) passes `check_research_question_status.py` and CI. Only the bare adjective `Citable` is actually gated, and "citable claim" is the phrasing an author is most likely to reach for when making the claim.

The denial these entries were added for (B2 line 552, "any citable claim is blocked until ...") is a *negated* use — anchoring on the negation (`no|not|any ... blocked`) rather than exempting the noun phrase outright would keep that case legal without opening the bypass.

---

scripts/ci/check_research_question_status.py
```diff
@@ -0,0 +1,249 @@
...
+    """Yield ``(line_number, section_id, status_text)`` for each Status block."""
+
+    section_id: str | None = None
+    lines = markdown.splitlines()
+    index = 0
+    while index < len(lines):
+        line = lines[index]
+        heading = SECTION_HEADING.match(line)
+        if heading:
+            section_id = heading.group(1).upper()
```
hollomancer: `section_id` is only ever overwritten by a heading matching `[A-F]\d+` and never cleared when the document leaves the numbered question sections, so Status lines further down inherit the last question ID.

`docs/research-questions.md` ends its question sections at `### F4.` (line 1087) and then has `## Output products & audiences` containing `### Form D fundraising analysis (published)` (1126) and `### Commercialization-benchmark methodology` (1144). Verified on this branch:

```
md = "### F4. Predictive (Tier 4)\n\n## Output products & audiences\n\n### Form D fundraising analysis (published)\n\n**Status:** Citable for the 2024 vintage.\n"
list(iter_status_blocks(md))  -> [(7, 'F4', 'Citable for the 2024 vintage.')]
```

A Form D Status claim is therefore authorized (or rejected) by whatever study happens to list **F4**, and the violation message misdirects the author to F4. The PR body names Form D promotion as the very next PR, so this is the next Status line that will be added.

Side effect: the `claim.section_id is None` branch (L194) is unreachable for anything after line 157 of the inventory. `test_claim_outside_numbered_section_is_rejected` only passes because its fragment has no preceding A–F heading, so the test does not cover the real document.

Clearing `section_id = None` on any `^#{2,4}\s` heading that does *not* match `[A-F]\d+` would make the None branch do its job.

---

scripts/ci/check_research_question_status.py
```diff
@@ -0,0 +1,249 @@
...
+    re.compile(r"\bnot\s+yet\s+validated\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+validated\b", re.IGNORECASE),
+    re.compile(r"\bunvalidated\b", re.IGNORECASE),
+    re.compile(r"\bremains\s+unvalidated\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+approved\s+for\s+citation\b", re.IGNORECASE),
+    re.compile(r"\bnot\s+citable\b", re.IGNORECASE),
+    re.compile(r"\bnon-citable\b", re.IGNORECASE),
+    re.compile(r"\bcitable\s+study(?:\s+contract|\s+manifest)?\b", re.IGNORECASE),
+    re.compile(r"\bcitable\s+claim\b", re.IGNORECASE),
+)
```
hollomancer: This is a fixed enumeration of negations, but the docs added in this same PR promise negations *generally* are refusals that need no study ("Negations (`Not computable`, `not validated`, `non-citable`) are refusals, not ranks" — `docs/research-questions.md`; same wording in `studies/README.md`). Ordinary refusals outside the list are scored as positive claims and fail the build.

Verified on this branch:

```
"Not currently computable."                          -> ('computable',)
"No longer computable after the source was withdrawn." -> ('computable',)
"Never computable from public data."                 -> ('computable',)
"Not fully validated."                               -> ('validated',)
"Cannot be validated without a hand-labelled sample." -> ('validated',)
"This is not yet a citable result."                  -> ('citable',)
```

An author writing any of these truthful refusals for a question with no study gets `make lint-boundaries` / CI failing with `Status claims 'computable' ... but the highest matching study is 'none'` — the guard blocking an honest downgrade. Note the arbitrariness: `not citable` is legal but `not yet a citable result` is not.

A single negation-window rule (`\b(?:not|no|never|cannot|non-)\W+(?:\w+\W+){0,3}?(?:computable|validated|citable)\b`) would cover the documented contract instead of chasing phrasings.

---

scripts/ci/check_research_question_status.py
```diff
@@ -0,0 +1,249 @@
...
+
+RANK_PATTERNS: tuple[tuple[str, re.Pattern[str]], ...] = (
+    ("computable", re.compile(r"\bcomputable\b", re.IGNORECASE)),
+    ("validated", re.compile(r"\bvalidat(?:ed|es)\b", re.IGNORECASE)),
```
hollomancer: `validat(?:ed|es)` also matches the ordinary verb *validates*, which is not a rank claim.

This already cost content in this PR: F3's Status was reworded from "The Phase 1 NSF review **validates** the Phase I→II cohort component" to "**reports** the ..." solely to satisfy the guard — a factual statement weakened to appease a lint rule. Any future Status prose like "the preregistered placebo check validates the timing pattern" will fail CI for a question with no `validated` study, even though nothing is being claimed.

The reserved rank is the adjective. Dropping the `|es` alternation keeps `Validated` gated without capturing the verb:

```suggestion
    ("validated", re.compile(r"\bvalidated\b", re.IGNORECASE)),
```

---

scripts/ci/check_research_question_status.py
```diff
@@ -0,0 +1,249 @@
...
+    for path in sorted((repository_root / "studies").glob("*/study.yaml")):
+        manifest = load_study_manifest(path)
+        if manifest.evidence_status is EvidenceStatus.RETIRED:
+            continue
+        for question in manifest.research_questions:
+            key = question.strip().upper()
+            current = best.get(key)
+            if current is None or STATUS_RANK[manifest.evidence_status] > STATUS_RANK[current]:
+                best[key] = manifest.evidence_status
+    return best
```
hollomancer, line 199: Authorization is keyed on the heading-level ID, but a `### B2.` section holds several distinct question bullets, so one study authorizes the reserved rank for all of its siblings.

`studies/phase-iii-census/study.yaml` lists `B2` for the unlabeled-follow-on census estimand. Section `### B2. Relational (Tier 2)` also contains the Phase 0 bullet at `docs/research-questions.md:552`, currently `**Status:** Not computable. Phase 0 design only ... classification criteria are not frozen.` Rewriting *that* bullet to `**Status:** Computable.` passes the guard, because `question_ranks["B2"]` is `reproducible` from a study whose estimand does not cover it.

Not necessarily a blocker for this PR — the inventory's ID granularity is section-level — but it is worth stating in the docstring / `studies/README.md` that authorization is per-section, not per-question, so the guarantee is not read as stronger than it is.

---

scripts/ci/check_research_question_status.py
```diff
@@ -0,0 +1,249 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+
+from sbir_etl.exceptions import ConfigurationError
+from sbir_etl.quality.study_manifest import EvidenceStatus, load_study_manifest
+
+
+REPOSITORY_ROOT = Path(__file__).resolve().parents[2]
+INVENTORY_PATH = Path("docs/research-questions.md")
+STUDIES_ROOT = REPOSITORY_ROOT / "studies"
```
hollomancer: `STUDIES_ROOT` is never referenced — `load_question_study_ranks` builds `repository_root / "studies"` from its parameter instead. It is a stale second definition of the studies path: updating it after a layout change would have no effect on the glob, leaving the checker scanning the old location and silently passing every reserved rank. Suggest deleting it.

```suggestion
```

---

hollomancer (commented):
Reviewed the guard script and the inventory rewrite. The check passes on the current tree and the 9 new unit tests are green, but the regex-based admission control has two holes and one over-broad rule that I verified by running `claimed_ranks` / `iter_status_blocks` against this branch.

Highest-impact: `Citable claim: ...` and `Citable study result ...` are stripped by the denial patterns and bypass the guard entirely, and `section_id` never resets outside the A–F question sections, so a Status line under `### Form D fundraising analysis` (named in the PR body as the next PR) is attributed to F4. Details inline.

---

scripts/ci/check_research_question_status.py
hollomancer: Addressed in 5a89185a83e6e44f35fc7b50d4382b35c45b551d: dropped the `citable claim` / `citable study` exemptions. Those phrases are now rank claims. The leftover STTR wording is covered by an `any … citable … blocked` denial.

scripts/ci/check_research_question_status.py
hollomancer: Addressed in 5a89185a83e6e44f35fc7b50d4382b35c45b551d: any heading that is not `[A-F][digit]` now clears `section_id`. Added a regression that puts a Citable Status under Form D output products after F4.

scripts/ci/check_research_question_status.py
hollomancer: Addressed in 5a89185a83e6e44f35fc7b50d4382b35c45b551d: replaced the fixed denial list with a 0–3-word negation window (`not|no|never|cannot`) plus `non-citable` / `unvalidated` / `not approved for citation`.

scripts/ci/check_research_question_status.py
hollomancer: Addressed in 5a89185a83e6e44f35fc7b50d4382b35c45b551d: the rank pattern is now `\\bvalidated\\b` only. `validates` is no longer a claim.

scripts/ci/check_research_question_status.py
hollomancer: Agreed this is section-level, not bullet-level. Documented that in the checker docstring and `studies/README.md` rather than inventing per-question IDs in this PR.

scripts/ci/check_research_question_status.py
hollomancer: Addressed in 5a89185a83e6e44f35fc7b50d4382b35c45b551d: deleted the unused `STUDIES_ROOT`.

PR #662, opened 2026-08-19
Add a pluggable M&A SearchTool and pick a default backend

### LINKED ISSUE #446: Evidence stage: M&A discovery and verification
Parent: #442
Donor PR: #371

Rebuild the paused M&A discovery concepts as an importable, tested evidence stage integrated with the capital-events pipeline.

Acceptance criteria:
- real, configurable search backend behind a typed interface
- typed candidate and verified-event records
- evidence URLs/snippets, extraction provenance, dates, acquirer, and value when available
- calibrated confidence policy rather than keyword-only confirmation
- cache/retry/rate-limit behavior through shared source mechanics
- unit and integration tests
- explicit integration into MAEvent/capital_events outputs

Do not revive the donor branch's untested root-level scripts as the production interface.

hollomancer:
## Description

Stacked on #661 / #446 step 2 of 3. Adds a real pluggable `SearchTool` on top of the relocated M&A discovery module.

**Recommended default backend: Tavily.** It returns extracted `content` snippets (not just SERP blurbs) that the keyword verifier and later LLM extractor can actually read, stays inside the ~$5 / 200-candidate cap on basic search, and is one API key. Brave is implemented as a second same-shape client. Comparison note: `specs/ma-discovery-integration/search-backends.md` (exploratory, non-citable — not a measured recall benchmark).

Runtime default is `none` and fails closed. `mock` is opt-in via `--search-backend mock`, because `MockSearchTool` returns a fixture hit that the verifier confirms and writes as evidence. A live backend selected without a key raises `ConfigurationError`. Tests mock httpx and never call the network.

Fixes the PR 1 mock bug: `MockSearchTool` now matches `physical optics` / `mercury systems` case-insensitively so a `RECIPIENT_V1` query string still hits the fixture.

The `ma-discovery-integration` spec is **Active for this bounded slice**: M&A recall was selected as a research priority on 2026-09-06 and `specs/status.md` records the scope. The LLM extractor eval is #663. No C3 collision layer, no Dagster job.

Fixes #446 (partial — search-tool slice only)

## Type of Change

- [x] Documentation update

## Versioning Impact

## Testing

- [x] All tests pass locally

```
uv run pytest tests/unit/enrichers/ma_discovery/ tests/unit/config/test_schemas.py -n0 -q
uv run ruff check sbir_etl/enrichers/ma_discovery tests/unit/enrichers/ma_discovery
uv run python scripts/ci/check_identity_boundaries.py
uv run python scripts/ci/check_config_boundaries.py
```

117 ma_discovery and config-schema unit tests pass on the current head. The fail-closed test clears `SBIR_ETL__MA_DISCOVERY__*` env overrides and the config cache so it pins the shipped YAML default. No `@pytest.mark.requires_api`. No live search API calls.

## Checklist


sbir_etl/enrichers/ma_discovery/search.py
```diff
@@ -0,0 +1,234 @@
...
+    if backend == "mock":
+        return MockSearchTool()
+
+    key = _resolve_api_key(api_key, cfg)
+    if not key:
+        logger.warning(
+            "M&A search backend '{}' selected but no API key is set; using MockSearchTool",
+            backend,
+        )
+        return MockSearchTool()
```
hollomancer: A live backend selected without credentials must not fall back to the fixture search tool. `MockSearchTool` can return a synthetic, confirmed Physical Optics acquisition, so a production/configuration mistake can write fabricated fixture evidence to `discovered_acquisitions.jsonl` while the command exits successfully. Keep `mock` as the explicit default, but if `tavily` or `brave` was selected, fail closed with `ConfigurationError` when its key is absent (and add a regression test for that selection path).

hollomancer (commented):
One functional fail-open issue is present. Separately, the current `Lint, Types, and Guards` failure is attributable to this PR: `ruff format --check` reports `tests/unit/enrichers/ma_discovery/test_press.py` would be reformatted. The branch also currently conflicts with `main`.

sbir_etl/enrichers/ma_discovery/search.py
hollomancer: Resolved: `tavily`/`brave` without a key now raise `ConfigurationError`. `mock` remains the explicit no-credentials default. Rebased onto main so the branch no longer conflicts.

sbir_etl/enrichers/ma_discovery/orchestrator.py
```diff
@@ -72,14 +73,37 @@ def write_verified_jsonl(path: Path, rows: Sequence[dict[str, Any]]) -> None:
...
+    parser.add_argument(
+        "--search-api-key",
+        default=None,
+        help="API key for a live backend. Falls back to config/env.",
+    )
     args = parser.parse_args(argv)
 
     queries = load_query_csv(args.input)
-    verified = asyncio.run(process_batch(queries, MockSearchTool()))
+    search_tool = build_search_tool(args.search_backend, api_key=args.search_api_key)
```
hollomancer, line 116: **The default CLI can still write synthetic fixture evidence.** `config/base.yaml` selects `mock`, so a no-argument run constructs `MockSearchTool`; a Physical Optics/Mercury query then writes the hard-coded fixture to `data/discovered_acquisitions.jsonl` and exits successfully. The missing-key fix protects explicitly selected live backends but not this default path. Make the fixture backend explicitly opt-in (or use a disabled/null default) before this operated CLI is mergeable.

hollomancer (commented):
Two merge decisions remain. First, the fixture backend must not be the no-argument operated path. Second, `specs/status.md` still classifies `ma-discovery-integration` as Deferred while this stack implements its next two steps. My recommendation is to keep the stack unmerged until M&A recall is explicitly selected, then mark the spec Active for the bounded slice and make mock execution explicitly opt-in.

sbir_etl/enrichers/ma_discovery/orchestrator.py
hollomancer: Re-verified against the current head `12c1271`. **This still holds** — the finding is accurate as written.

The reason the missing-key guard doesn't cover this path: in `build_search_tool` (`search.py:196-212`) the `if backend == "mock": return MockSearchTool()` early return sits *above* the credential check at `search.py:214-219`, so the check is never reached on the default path.

Traced end to end:

1. `orchestrator.py:88-95` — `--search-backend` has `default=None`.
2. `build_search_tool(None)` → `backend_raw` falls through to `cfg.search_backend`, which is `"mock"` in `config/base.yaml:779-784`, in `MADiscoveryConfig` (`domain.py:625-628`), and in `.env.example`. `MA_DISCOVERY_SEARCH_BACKENDS` is `frozenset({"mock", "tavily", "brave"})` — there is no disabled option.
3. `MockSearchTool.search` (`search.py:58-70`) matches the Physical Optics / Mercury Systems query that `queries.py` itself generates, returning a hard-coded hit with `link: "http://example.com"`.
4. `verifier.py:26-34` confirms it (both names present, `"acquisition"` in the keyword list).
5. `write_verified_jsonl` writes the fixture to `data/discovered_acquisitions.jsonl`, prints "Found 1 verified acquisitions", exits 0.

`tests/unit/enrichers/ma_discovery/test_orchestrator.py:16-29` asserts this exact output shape, so the behavior is pinned rather than incidental.

**Suggested minimal fix** — add a disabled default so the fixture becomes explicitly opt-in, leaving `--search-backend mock` as the only route to it:

- `domain.py:614` — add `"none"` to `MA_DISCOVERY_SEARCH_BACKENDS`; `domain.py:625` — `default="none"` (and correct the docstring at 619-628, which says "Runtime default is the in-memory mock").
- `config/base.yaml:780` — `search_backend: none`.
- `search.py` `build_search_tool` — after the membership check (line 204) and **before** the `mock` return, raise `ConfigurationError` for `"none"` pointing at `ma_discovery.search_backend`.
- `.env.example` and `specs/ma-discovery-integration/search-backends.md:41-42` both advertise the mock default and would go stale.

Regressions: `main([...])` with no `--search-backend` raises `ConfigurationError` and never creates the output file; `build_search_tool(None, config=MADiscoveryConfig())` raises. `test_main_uses_mock_backend` and `test_factory_mock_name_ignores_key` pass `mock` explicitly and stay green — which is the point. `tests/unit/config/test_schemas.py` asserts the `"mock"` default in two places and needs updating to `"none"`.

Leaving this thread open since the defect is unfixed.

---

hollomancer:
Two further items from this review pass, separate from the open thread on the fixture-evidence default.

**1. Merge conflict against `main`.** This branch no longer merges cleanly — `.env.example` and `config/base.yaml` have both changed on both sides since the branch point. #663 is stacked on this branch and inherits the conflict, so this one needs resolving first.

**2. `process_batch` emits duplicate rows per pair** — `orchestrator.py:44-58`. The comment reads `# one hit per (company, acquirer) is enough`, but the `break` exits only the inner `for res in results` loop, not the outer iteration over query rows. `queries.generate_queries` (`queries.py:44-49`) emits **four** CSV rows per pair, so a confirmed pair is appended up to four times to `data/discovered_acquisitions.jsonl`. A `seen: set[tuple[str, str]]` tracked across rows is what the comment is actually describing.

**3. `build_search_tool` ignores configured client settings when a backend is named explicitly** — `search.py:196-223`. When `name` is truthy, the `if not backend_raw:` block that assigns `cfg = _load_ma_discovery_config()` is skipped, so `cfg` is still `None` at lines 221-223 and `timeout` / `rate_limit_per_minute` / `max_results` silently fall back to module constants. `_resolve_api_key` does load the config, but rebinds only its own local. So `--search-backend tavily` never honors `config/base.yaml`'s `timeout_seconds`, `rate_limit_per_minute`, or `max_results`. This is invisible today only because the YAML values happen to equal the module defaults — it would bite the first time someone tunes the rate limit. Loading the config once at the top of the function when `config is None` fixes it.

Item 2 is a data-correctness bug in the written artifact; item 3 is latent. Both are small and local to this PR's own new code.

---

sbir_etl/enrichers/ma_discovery/search.py
```diff
@@ -1,30 +1,233 @@
...
+    key = _resolve_api_key(api_key, cfg)
+    if not key:
+        raise ConfigurationError(
+            f"M&A search backend {backend!r} requires an API key",
+            config_key="ma_discovery.search_api_key",
+        )
+
+    timeout = cfg.timeout_seconds if cfg is not None else DEFAULT_TIMEOUT_SECONDS
+    rate = cfg.rate_limit_per_minute if cfg is not None else DEFAULT_RATE_LIMIT_PER_MINUTE
+    max_results = cfg.max_results if cfg is not None else DEFAULT_MAX_RESULTS
```
hollomancer: **Explicit `--search-backend` skips configured client settings.** When `name` is truthy, the `if not backend_raw:` block that loads `cfg` is skipped, so `cfg` is still `None` here and `timeout` / `rate_limit_per_minute` / `max_results` fall back to module constants. `_resolve_api_key` does load config, but rebinds only its own local.

`--search-backend tavily` therefore never honors `config/base.yaml`. Invisible today only because the YAML values equal the defaults. Load config once at the top when `config is None`.

sbir_etl/enrichers/ma_discovery/orchestrator.py
```diff
@@ -72,14 +73,37 @@ def write_verified_jsonl(path: Path, rows: Sequence[dict[str, Any]]) -> None:
             handle.write(json.dumps(row) + "\n")
 
 
+async def _run_batch(
+    queries: list[dict[str, str]], search_tool: SearchTool
+) -> list[dict[str, Any]]:
+    """Run ``process_batch`` and close the search client when it supports it."""
+    try:
+        return await process_batch(queries, search_tool)
```
hollomancer, line 89: **`process_batch` still writes duplicate rows per pair.** This new helper calls `process_batch`, whose `break` (`orchestrator.py:58`) exits only the inner `for res in results` loop. `queries.generate_queries` emits four CSV rows per pair (`queries.py:44-49`), so a confirmed pair is appended up to four times to `data/discovered_acquisitions.jsonl`. The comment on the `break` says "one hit per (company, acquirer) is enough."

A `seen: set[tuple[str, str]]` tracked across rows is what that comment describes. Separate from the still-open mock-default thread on `main()`.

hollomancer (commented):
The mock-default thread remains open and is still the merge blocker. Two additional defects on this head:

1. `process_batch` writes up to four JSONL rows per pair because `break` only exits the inner loop.
2. Explicit `--search-backend` skips YAML timeout/rate-limit/max_results.

The branch also still conflicts with `main` on `.env.example` and `config/base.yaml`; #663 inherits that. Recommendation: keep unmerged until M&A recall is explicitly selected, mark `ma-discovery-integration` Active for the bounded slice, and make mock execution opt-in.

sbir_etl/enrichers/ma_discovery/orchestrator.py
hollomancer: Fixed in `a7f24703`, implementing your proposal essentially as written.

- `"none"` added to `MA_DISCOVERY_SEARCH_BACKENDS` and made the schema default; docstring at `domain.py:619` corrected (it claimed "Runtime default is the in-memory mock").
- `config/base.yaml` → `search_backend: none`; the stale mock-default prose in `.env.example` and `specs/ma-discovery-integration/search-backends.md:41-42` updated too.
- The `ConfigurationError` for `"none"` sits **after** the membership check and **before** the `mock` early return, which is the ordering that made the credential guard unreachable in the first place.

The error text names the actual hazard rather than just the missing setting, since that is what a reader hitting it needs to know:

> No M&A search backend is selected. MockSearchTool returns fixture hits that verify and get written as discovered acquisitions, so it is opt-in only: pass `--search-backend mock` deliberately, or set a live backend (tavily, brave) with an API key.

Regressions added: `main([...])` with no `--search-backend` raises and **creates no output file**; `build_search_tool` raises from both `("none")` and `(None, config=MADiscoveryConfig())`. Updated the two `test_schemas.py` assertions that pinned `"mock"`. `test_main_uses_mock_backend` and `test_factory_mock_name_ignores_key` pass `mock` explicitly and stay green, which is the point.

Verified both new tests fail against the pre-fix source. `make lint` (ruff, format, UP042, mypy 329 files) and `make lint-boundaries` pass; 117 tests pass across `tests/unit/enrichers/ma_discovery` and `tests/unit/config/test_schemas.py`.

---

sbir_etl/enrichers/ma_discovery/search.py
hollomancer: Confirmed and fixed in `a7f24703`. `cfg` is now loaded once at the top:

```python
cfg = config if config is not None else _load_ma_discovery_config()
```

Your diagnosis of why it was invisible is exactly right, and it is the part worth keeping: `config/base.yaml` sets `timeout_seconds: 30`, `rate_limit_per_minute: 60`, `max_results: 5`, which are the `DEFAULT_*` constants verbatim. Nothing would have surfaced this until someone tuned the YAML and watched it do nothing.

`_load_ma_discovery_config` already swallows a missing config and returns `MADiscoveryConfig()`, so hoisting it does not introduce a new failure mode on the explicit-name path.

The regression pins the settings *and* asserts they differ from the constants, so the test cannot pass by coincidence the way the old code did:

```python
assert tool.rate_limit_per_minute == 11
assert tool._max_results == 3
assert tool._client.timeout.read == 7
assert (11, 3, 7) != (DEFAULT_RATE_LIMIT_PER_MINUTE, DEFAULT_MAX_RESULTS, DEFAULT_TIMEOUT_SECONDS)
```

Verified it fails against the pre-fix source.

---

sbir_etl/enrichers/ma_discovery/orchestrator.py
hollomancer: Confirmed and fixed in `a7f24703`. The `seen: set[tuple[str, str]]` you described is now tracked across rows, and the `break` comment moved onto the `seen.add(pair)` line where it is actually true.

Worth noting the duplication was up to 4x, not merely possible: `generate_queries` returns exactly four strings per pair (`queries.py:44-49`), and `query_rows_from_events` emits a CSV row for each, so every confirmed pair was appended four times to `data/discovered_acquisitions.jsonl` — and the file is opened `"w"` and rewritten wholesale, so the duplicates are the record, not an append artifact.

The check is at the top of the row loop rather than after verification, so a pair that already hit also skips its remaining three search calls. That is a 4x reduction in live API calls against a rate-limited paid backend, which matters once this stops running on the fixture.

Regression added: build the real four rows via `generate_queries("Physical Optics Corporation", "Mercury Systems")`, assert `len(queries) == 4`, then assert `process_batch` returns exactly one row. Fails against the pre-fix source with 4.

---

sbir_etl/enrichers/ma_discovery/orchestrator.py
```diff
@@ -4,8 +4,9 @@
 through a pluggable ``SearchTool``, and feeds snippets into
 ``verify_acquisition``. Confirmed hits are written as JSONL.
 
-This PR ships only ``MockSearchTool``. A live search-API client is a
-later step.
+The CLI constructs the search tool via ``build_search_tool``. Runtime
+default is ``mock``. Selecting ``tavily`` or ``brave`` without a key
+fails closed.
```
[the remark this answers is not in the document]
hollomancer: Fixed in 8176cdcc: the module docstring and the `--search-backend` help now say the config default is `none` (fail-closed) and `mock` is opt-in via `--search-backend mock`. Also merged `main` (a03f0647) to clear the `.env.example` conflict; both blocks kept. 117 ma_discovery and schema tests pass; ruff and docs-check clean.

---

specs/ma-discovery-integration/design.md
```diff
@@ -99,7 +99,7 @@ These are choices the design intentionally does *not* pin, because they're indep
 
 | Decision | Default I'd lead with | Why deferred |
 |---|---|---|
-| Search backend (Tavily / Brave / Bing / Serper) | Tavily — purpose-built for snippet-focused agentic search; pluggable interface stays put | Pricing changes; better resolved with a real cost estimate |
+| Search backend (Tavily / Brave / Bing / Serper) | **Tavily** is the default production client; Brave is a second same-shape backend. Runtime default stays mock until a key is set. Comparison: [`search-backends.md`](search-backends.md) (exploratory, non-citable — not a recall benchmark). | Comparison note landed; revisit only if the sample run shows snippet quality is the bottleneck |
```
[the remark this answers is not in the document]
hollomancer: Fixed in 8176cdcc: the open-decisions row and sequencing step 3 now say `none` is the fail-closed default and `mock` is opt-in.

---

specs/status.md
```diff
@@ -79,9 +79,10 @@ bypassing lifecycle review; the status and rationale still require human judgmen
   Per-source adapters stay split (#443 NIH RePORTER, then SAM/PatentsView).
   Tasks 6.1–6.2 remain optional Phase 2 expansion.
 - **`ma-discovery-integration` — Deferred.** The #371 toolkit now lives at
-  `sbir_etl/enrichers/ma_discovery/` (issue #446, toolkit relocation). Search
-  backend, LLM extractor, and collision policy remain unbuilt. Revisit only
-  when M&A recall becomes a selected research priority.
+  `sbir_etl/enrichers/ma_discovery/` (issue #446). SearchTool has a real
+  Tavily client (Brave optional) behind config; runtime default remains mock
+  without a key. LLM extractor and collision policy remain unbuilt. Revisit
+  only when M&A recall becomes a selected research priority.
```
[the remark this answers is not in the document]
hollomancer: Fixed in 8176cdcc: the registry entry now says the runtime default is `none` (fail-closed) and `mock` is opt-in only.

---

hollomancer:
Fresh review of head `8176cdcc`, then pushed `47e7c05`.

**Fixed.** `test_main_without_backend_fails_closed_and_writes_nothing` ran `main()` against the real `config/base.yaml` through `get_config()`, which also applies `SBIR_ETL__*` env overrides and caches. Reproduced: with `SBIR_ETL__MA_DISCOVERY__SEARCH_BACKEND=mock` in the shell, the exact line `.env.example` documents, the test failed and wrote the fixture hit. The test now clears both env overrides and the config cache, so it pins the shipped YAML default. Verified it passes with and without the override set.

Also: `_load_ma_discovery_config` swallowed `ConfigurationError` at debug level, so a typo in the backend env var surfaced as the unrelated "No M&A search backend is selected". It now logs the real cause at warning level before falling back. Tavily and Brave responses with no results key now log a warning instead of silently mapping to "Found 0 verified acquisitions".

**Not changed, flagging for the design pass:**
- `orchestrator.py:84-120` has no per-query error handling. One `APIError` on query N discards the confirmed rows from queries 1..N-1 and exits via traceback. At up to 800 queries on a paid backend that is expensive.
- `base_client.py:115-118, 175-181` (pre-existing): connection errors are re-raised as `APIError` before tenacity sees them, so DNS and reset failures are never retried, only timeouts and 5xx.
- `specs/status.md:93-97` still labels the spec Deferred while describing active implementation. Registry placement is the author's call.

**Verified clean:** Tavily and Brave request/response mapping and auth headers match the documented APIs; client closed on every path; 4xx/429 raise rather than fail open; the API key is never logged or written to JSONL; only `sbir_etl.config.loader.get_config` reads config; no test touches the network; the dedupe and explicit-backend tests both fail with their fixes reverted. 117 tests pass; mypy clean.

---

hollomancer:
Superseded by #699, which rebases this slice onto current `main` (including #698) and adds C3, the confirmatory harness, and the held-out miss. Closing so the two PRs do not compete for the same commits.

PR #663, opened 2026-08-19
Evaluate an M&A snippet LLM extractor without cutting over

### LINKED ISSUE #446: Evidence stage: M&A discovery and verification

hollomancer:
## Description

Stacked on #662 / #446 step 3 of 3. Adds a typed snippet extractor and a fixture-only eval of keyword vs a structured LLM JSON prompt. **Keyword path remains the orchestrator default.** Eval is exploratory / non-citable.

- `KeywordExtractor` adapts existing `verify_acquisition` onto `{confirmed, matched_company, matched_acquirer, acquisition_date, value_usd, citation_url, reason}`. It lives in `extractor.py` (pipelines tier) with the types and the deterministic payload parsing and pair validation.
- `LlmExtractor` lives in `llm_extractor.py` (exploratory tier), because the pipelines contract forbids inference. It takes an injected chat callable / `OpenAIClient`. Prompt asks for that JSON schema, marks the snippet as untrusted data, and caps `reason`. Factory returns `None` without `OPENAI_API_KEY`. The package does not re-export it; `check_tier_boundaries.py` fails a future pipelines-tier import.
- A confirmation is accepted only when the returned pair matches the requested pair under `RECIPIENT_V1`; the citation is always the search result's own URL.
- Frozen synthetic snippets in `tests/fixtures/ma_discovery/` (slam-dunk, negative, rumor, suffix/case, missing date/value).
- Harness reports precision/recall on `confirmed` plus date/value fill rates from fixtures. No network.
- Notebook `notebooks/explorations/ma_discovery_extractor_eval.ipynb` (README puts new investigations here, not `examples/`). Live OpenAI is behind `MA_DISCOVERY_LIVE_LLM=1`.
- Write-up: `specs/ma-discovery-integration/extractor-eval.md`.

**Recommendation for the next PR:** keep keyword as the orchestrator default; if `LlmExtractor` is wired, make it opt-in on the existing `gpt-4.1-mini` client until a labeled Form-D-missing sample exists. Do not treat fixture scores as a cutover warrant.

The `ma-discovery-integration` spec is **Active for this bounded slice** (M&A recall selected 2026-09-06; see `specs/status.md`). No C3 collision, no `MAEvent.confidence` rewrite, no capital-events join.

Fixes #446 (partial — extractor-eval slice only)

## Type of Change

- [x] Documentation update

## Versioning Impact

## Testing

- [x] All tests pass locally

```
uv run pytest tests/unit/enrichers/ma_discovery/ -n0 -q
uv run ruff check sbir_etl/enrichers/ma_discovery tests/unit/enrichers/ma_discovery
uv run python scripts/ci/check_tier_boundaries.py
uv run python scripts/ci/check_removed_src_references.py
```

60 ma_discovery unit tests pass on the current head. No live OpenAI or search API calls. LLM tests inject a mock that returns JSON.

## Checklist


sbir_etl/enrichers/ma_discovery/extractor.py
```diff
@@ -0,0 +1,287 @@
...
+        return ExtractionVerdict(confirmed=False, reason=parse_failure_reason)
+
+    confirmed = payload.get("confirmed")
+    if not isinstance(confirmed, bool):
+        return ExtractionVerdict(
+            confirmed=False,
+            reason="LLM JSON missing boolean 'confirmed'",
+        )
+
+    date = _as_optional_str(payload.get("acquisition_date"))
```
hollomancer: This accepts any nonblank string as `acquisition_date`, even though the prompt and verdict contract require ISO `YYYY-MM-DD`. A model response such as `"March 2024"` therefore becomes a confirmed pipeline verdict with an invalid date. Please parse with `date.fromisoformat()` (and require canonical round-tripping), then either null the field or fail the verdict on invalid schema; add a test for a confirmed payload with a non-ISO date. The same validation boundary should reject non-finite/negative `value_usd` values rather than passing them downstream.

hollomancer (commented):
The extractor is well isolated and remains opt-in, but its claimed structured schema is not enforced at the model boundary.

sbir_etl/enrichers/ma_discovery/extractor.py
hollomancer: Resolved: `acquisition_date` must be canonical `YYYY-MM-DD` via `date.fromisoformat()` plus round-trip; `"March 2024"` is nulled. `value_usd` rejects negatives, NaN, and infinities. Confirmed payloads still confirm; only the bad fields are dropped.

sbir_etl/enrichers/ma_discovery/extractor_eval.py
```diff
@@ -0,0 +1,264 @@
...
+def score_extractor(
+    extractor: SnippetExtractor,
+    fixtures: Sequence[FixtureCase],
+    *,
+    name: str | None = None,
+) -> ExtractorScores:
+    """Run ``extractor`` over ``fixtures`` and score ``confirmed`` plus fill rates."""
+    cases = tuple(_score_case(extractor.extract(item.as_input()), item) for item in fixtures)
+    return scores_from_cases(
+        cases, name=name or getattr(extractor, "name", extractor.__class__.__name__)
```
hollomancer: The current head still fails the required MyPy gate here: `getattr(extractor, "name", ...)` is inferred as `str | Any | None`, but `scores_from_cases()` requires `str`. Normalize this to a guaranteed non-empty `str` before the call (or make the protocol guarantee `name: str`) and add/retain a test for an extractor without a usable name. GitHub Actions run 32288038497 reports this exact `arg-type` error.

hollomancer (commented):
One current CI blocker remains after the schema-validation fix.

sbir_etl/enrichers/ma_discovery/extractor_eval.py
hollomancer: Fixed in `6b0f0668`: `score_extractor()` now narrows the optional extractor name to a guaranteed non-empty string and falls back to the extractor class name. Added a nameless-extractor regression test. GitHub’s MyPy/lint gate and all four fast-test shards now pass.

sbir_etl/enrichers/ma_discovery/extractor.py
```diff
@@ -0,0 +1,303 @@
...
+        model: str = DEFAULT_MODEL,
+        temperature: float = 0.0,
+    ) -> None:
+        self.model = model
+        self.temperature = temperature
+        self._chat = _bind_chat(client, model=model, temperature=temperature)
+
+    def extract(self, item: ExtractionInput) -> ExtractionVerdict:
+        raw = self._chat(EXTRACTOR_SYSTEM_PROMPT, build_user_prompt(item))
+        return verdict_from_payload(parse_llm_payload(raw), citation_fallback=item.source_url)
```
hollomancer: **A model can confirm an unrelated pair and replace citation provenance.** `verdict_from_payload()` accepts `confirmed=true` without checking that `matched_company` and `matched_acquirer` correspond to this `ExtractionInput`, and it prefers the model's `citation_url` over the search-result URL. A response about Foo/Bar with a fabricated URL therefore becomes a confirmed verdict for the requested A/B pair. Validate the returned names against the requested pair using an existing deterministic normalization, fail closed on missing/mismatched names, and keep `item.source_url` authoritative; add regressions for both cases.

hollomancer (commented):
One model-boundary provenance issue remains before this typed extractor is safe to wire into the orchestrator.

sbir_etl/enrichers/ma_discovery/extractor.py
hollomancer: Re-verified against the current head `6b0f066`. **Both halves of this still hold.**

`verdict_from_payload` (`extractor.py:151-179`) never receives `item`, so it structurally cannot compare the returned names against the requested pair — the only guard on `confirmed` is `isinstance(confirmed, bool)` at line 161. And line 178 puts the payload's `citation_url` first with `citation_fallback` only as an `or` fallback.

Concrete path: for `ExtractionInput(company="Aether Photonics", acquirer="Helios Defense", source_url="https://newswire.example/real-article")`, a payload of `{"confirmed": true, "matched_company": "Foo Industries", "matched_acquirer": "Bar Holdings", "citation_url": "https://fabricated.example/no-such-page"}` yields a confirmed verdict carrying a URL that appeared in no search result. Callers key verdicts by the *requested* pair — `extractor_eval.py:205-215` records `predicted_confirmed` against the fixture's own pair and never inspects `matched_company` — so the Foo/Bar response becomes an Aether/Helios confirmation.

The system prompt does ask for `confirmed=false` on "a different pair of firms" (line 51), but that is an instruction, not a validation, and it is precisely what fails under hallucination or a prompt injection carried in the snippet.

**Suggested minimal fix:**

1. Change the signature to `verdict_from_payload(payload, *, item: ExtractionInput, parse_failure_reason=...)` and drop `citation_fallback`; update the single call site at line 226.
2. After the bool check, fail closed when `confirmed` and the pair doesn't match, via a helper built on the canonical primitive:

```python
from sbir_etl.identity import CompanyNameProfile, normalize_company_name
```

Use **`CompanyNameProfile.RECIPIENT_V1`** (`identity/company_names.py:38` → `_matching_v1(..., remove_suffixes=True)`). It makes `"Aether Photonics, Inc."` equal `"Aether Photonics"` — the prompt explicitly permits suffix/case variation — while keeping `"Foo Industries"` distinct, and it is already what the sibling `queries.py:33` uses to build the query strings these snippets come from, so extractor and query generator agree on identity. `MATCHING_V1` would over-reject, since it keeps suffixes. Treat an empty normalization on either side as a mismatch.

3. Make provenance authoritative: line 178 becomes `citation_url=item.source_url`.

Regressions worth adding: confirmed-but-unrelated pair rejected; confirmed with null matched names rejected; model `citation_url` cannot override `source_url`; and a suffix/case variant (`"AETHER PHOTONICS, INC."` for requested `"Aether Photonics"`) still confirms — that last one pins `RECIPIENT_V1` rather than raw equality and keeps the fix from over-rejecting.

Note `test_llm_parses_schema_from_plain_json` (line 81) asserts the payload URL wins over `source_url`; that assertion inverts under this fix. The four direct `verdict_from_payload(payload)` calls at lines 189/207/219/231 need an `item=` argument. `gold_replay_chat` echoes the fixture's own pair, so `test_extractor_eval.py` stays green.

Leaving this thread open since the defect is unfixed.

---

hollomancer: Fixed in `5feca011`, implementing your three-step proposal as specified.

1. `verdict_from_payload(payload, *, item: ExtractionInput, ...)`; `citation_fallback` dropped; single call site updated.
2. A confirmation fails closed unless both returned names match the requested pair under `CompanyNameProfile.RECIPIENT_V1` — your reasoning for that profile over `MATCHING_V1` is right and I kept it: it is what `queries.py:33` already uses to build these query strings, so extractor and query generator now agree on identity rather than each having their own. Empty normalization on either side is a mismatch, which is what makes confirmed-with-null-names reject.
3. `citation_url=item.source_url` unconditionally.

One addition beyond the proposal: the rejected verdict carries the returned names through in `matched_company` / `matched_acquirer` and names both pairs in `reason` —

> `LLM confirmed a different pair than requested: asked 'Aether Photonics'/'Helios Defense', got 'Foo Industries'/'Bar Holdings'`

A silent `confirmed=False` would erase the evidence that the model went off-target, which is the signal worth keeping in an eval harness.

All four regressions you listed are in, and your prediction about the existing suite held exactly: `test_llm_parses_schema_from_plain_json` asserted the payload URL wins, and that assertion is now inverted with a comment saying why. The four direct calls take `item=`. `gold_replay_chat` echoes the fixture's own pair, so `test_extractor_eval.py` stayed green untouched.

One consequence worth flagging that follows from your fix rather than from it being wrong: four existing tests (`nulls_non_iso_date`, `nulls_noncanonical_iso_date`, `nulls_non_finite_or_negative_value`, `ignores_unknown_placeholder_date`) sent `confirmed: true` with **no** matched names, and now correctly reject. They were testing date and value coercion, not pair validation, so I gave them a `_payload()` helper that names the requested pair and added a separate test for the null-names rejection. Otherwise the pair check would have silently swallowed four coercion tests.

Also merged `feat/ma-discovery-search-tools` in, since the fail-closed-backend fixes landed there and this branch stacks on it.

56 `ma_discovery` tests pass; `make lint` (mypy, 331 files) and `make lint-boundaries` pass.

---

sbir_etl/enrichers/ma_discovery/extractor.py
```diff
@@ -0,0 +1,340 @@
...
+        temperature: float = 0.0,
+    ) -> None:
+        self.model = model
+        self.temperature = temperature
+        self._chat = _bind_chat(client, model=model, temperature=temperature)
+
+    def extract(self, item: ExtractionInput) -> ExtractionVerdict:
+        raw = self._chat(EXTRACTOR_SYSTEM_PROMPT, build_user_prompt(item))
+        return verdict_from_payload(parse_llm_payload(raw), item=item)
+
```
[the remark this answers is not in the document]
hollomancer: Fixed in 6f0e7630. `LlmExtractor` keeps the injected client, exposes `close()`, and works as a context manager; a bare chat callable has nothing to close and is tolerated. Two tests pin the close path. Also merged the updated `feat/ma-discovery-search-tools` (which carries `main`) so this branch no longer conflicts on `.env.example`. 149 tests, ruff, mypy, docs-check pass.

---

tests/unit/enrichers/ma_discovery/test_extractor.py
```diff
@@ -0,0 +1,295 @@
...
+    assert build_llm_extractor() is None
+
+
+def test_build_llm_extractor_uses_openai_client_when_key_present(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setenv("OPENAI_API_KEY", "sk-test-not-used")
+    extractor = build_llm_extractor()
+    assert isinstance(extractor, LlmExtractor)
+    assert extractor.model == "gpt-4.1-mini"
```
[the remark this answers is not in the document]
hollomancer: Fixed in 6f0e7630: the test closes the extractor in a `finally`.

---

tests/unit/enrichers/ma_discovery/test_extractor_eval.py
```diff
@@ -0,0 +1,130 @@
...
+    fixtures = load_fixtures()
+    scores = score_extractor(KeywordExtractor(), fixtures)
+    tp, fp, fn, tn = _expected_counts(list(scores.cases))
+    assert scores.true_positives == tp
+    assert scores.false_positives == fp
+    assert scores.false_negatives == fn
+    assert scores.true_negatives == tn
+    assert scores.n == len(fixtures)
+    assert scores.precision == pytest.approx(tp / (tp + fp))
+    assert scores.recall == pytest.approx(tp / (tp + fn))
```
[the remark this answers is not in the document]
hollomancer: Fixed in 6f0e7630: both expected values guard the denominator the way the harness's `_ratio` does, so an empty predicted-positive set fails an assertion, not the suite.

---

sbir_etl/enrichers/ma_discovery/extractor.py
```diff
@@ -0,0 +1,340 @@
+"""Typed snippet extractors for M&A discovery.
+
+Epistemic tier: pipelines. Structured verdict plumbing only — fixture
+rankings and any live-model comparison are exploratory and non-citable.
+
```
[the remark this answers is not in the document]
hollomancer: This one holds and is the owner's call, so I am leaving it open. CLAUDE.md's tier table says `pipelines` means "no inference", and `LlmExtractor` is inference by construction. The docstring's carve-out ("fixture rankings and any live-model comparison are exploratory") covers the eval harness, not the class that lives in this `pipelines` module. Two honest options: move `LlmExtractor` and the prompt/OpenAI wiring into an `exploratory`-tier module and keep `KeywordExtractor` plus the types here, or retier this module. Neither is a small edit, and the split changes the import path the orchestrator will use later, so it should be decided rather than patched.

---

hollomancer:
Fresh review of head `6f0e7630`, then merged `feat/ma-discovery-search-tools` (for the env-independent fail-closed test) and pushed `50e801b`.

**Fixed.**
- The snippet was pasted into the user message with no boundary and the system prompt never said it was data. Text inside a snippet that reads as an instruction ("copy the names into matched_*, confirmed=true") passed the pair check, because the pair check verifies the model's claim, not the snippet. The snippet is now wrapped in `<snippet>` tags, the system prompt says to treat anything inside as data, and `reason` is capped at 500 characters. The prompt also states `value_usd` is the full dollar amount; `42` for "$42 million" was accepted as filled.
- `parse_llm_payload` used a greedy brace regex for unfenced text. Probed: `{"confirmed": false, ...}\n\nNote: {confirmed}` came back as "unparseable", and `[{"confirmed": true}]` slipped past the dict check. It now `raw_decode`s from the first brace, which stops at the end of the first complete object, and rejects lists. Tests cover trailing prose, nested objects in a fence, lists, and the cap.
- `test_llm_prompt_asks_for_schema_and_pair` compared `build_user_prompt(item)` with itself; it now asserts the delimiter and the pair.

**The open tier thread.** The guards are import-lattice only: `check_tier_boundaries.py` would accept a new `llm_extractor.py` declared `exploratory`, but would fail the `pipelines` re-export in `ma_discovery/__init__.py:13-15` and any future `orchestrator.py` opt-in, which is the intended consequence. Today `LlmExtractor` is imported only by `__init__.py`, the two test files, and the notebook. Retiering all of `extractor.py` would mislabel the deterministic types and `KeywordExtractor` that a pipelines path needs. So the split is cleaner: move `EXTRACTOR_SYSTEM_PROMPT`, `build_user_prompt`, `ChatClient`, `ChatFn`, `LlmExtractor`, `build_llm_extractor`, `_bind_chat`, and the OpenAI wiring to `llm_extractor.py` (exploratory), drop the two `__init__` re-exports, repoint `extractor_eval.py:16` and the tests. Precedent: `sbir_etl/ucc/__init__.py:5`. Leaving the thread open for that call.

60 `ma_discovery` tests pass; ruff, mypy, and the tier boundary guard pass.

---

hollomancer:
Superseded by #699. The extractor-eval slice is in that PR, still keyword-default, still exploratory. Closing so this stack is not merged beside the confirmatory branch.

PR #666, opened 2026-08-19
Add a weekly literature-map refresh (OpenAlex + grey RSS)

hollomancer:
## Description

The literature map in `docs/research/literature-map/` was a one-shot OpenAlex snapshot with no refresh path. This adds:

- `scripts/data/update_literature_map.py` (exploratory): re-search OpenAlex, pull National Academies Press books by source id, ingest recent GAO / NAP / CRS / ITIF items from public RSS/Atom feeds, refresh citation/OA metadata for existing OpenAlex IDs, classify **new** titles with deterministic keywords, write `sbir_literature_map.csv` + `refresh_status.md`.
- `.github/workflows/literature-map.yml`: Monday 09:17 UTC (and `workflow_dispatch`). Opens `chore/literature-map-refresh` when those files change. Does **not** push to `main`.
- `OpenAlexClient.search_works` for the works API (polite-pool `mailto` via optional `OPENALEX_MAILTO`).

Grey rows without an OpenAlex work id use synthetic keys (`gao:…`, `nap:…`, `crs:…`, `itif:…`). A later OpenAlex hit with the same title or DOI is merged onto that row. RSS is a recency window, not a backfill of `[L#]`. CSIS's public RSS is mostly old events and is not ingested; those `[L#]` entries stay hand-tracked.

Existing rows keep their stored relevance/area labels. Authored memos (`sbir_literature_map.md`, `citation_gap_memo.md`) are not rewritten. This does not promote the map, edit `[L#]`, or change study Status.

**Operator setup:** optional repo secret `OPENALEX_MAILTO` for OpenAlex's polite pool. No API key. Grey feeds are unauthenticated.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes the issue)
- [x] Documentation update

## Versioning Impact

## Testing

- [x] All tests pass locally

`uv run pytest tests/unit/scripts/test_update_literature_map.py tests/unit/test_ci_test_contract.py -n0 -q` — 17 passed.

## Checklist

### DIFF
--- a/sbir_etl/enrichers/openalex_client.py
+++ b/sbir_etl/enrichers/openalex_client.py
@@ -200,13 +200,12 @@ class OpenAlexClient(BaseAsyncAPIClient):
         """GET ``/works`` with caller-supplied query params.
 
         Used by the exploratory literature-map refresh. Returns the raw JSON
-        object (``results``, ``meta``). Empty ``results`` on 4xx; propagates
-        :class:`APIError` on 5xx.
+        object (``results``, ``meta``). A 404 is an empty page; other 4xx/5xx
+        errors propagate so malformed filters are not reported as zero hits.
         """
         try:
             return await self._make_request("GET", "works", params=self._with_mailto(params))
         except APIError as e:
-            status = e.details.get("http_status")
-            if status and 400 <= status < 500:
+            if e.details.get("http_status") == 404:
                 return {"results": [], "meta": {}}
             raise
--- a/sbir_etl/enrichers/openalex_client.py
+++ b/sbir_etl/enrichers/openalex_client.py
@@ -195,3 +195,18 @@ class OpenAlexClient(BaseAsyncAPIClient):
             return None
 
         return _parse_author(profile)
+
+    async def search_works(self, params: dict[str, Any]) -> dict[str, Any]:
+        """GET ``/works`` with caller-supplied query params.
+
+        Used by the exploratory literature-map refresh. Returns the raw JSON
+        object (``results``, ``meta``). Empty ``results`` on 4xx; propagates
+        :class:`APIError` on 5xx.
+        """
+        try:
+            return await self._make_request("GET", "works", params=self._with_mailto(params))
+        except APIError as e:
+            status = e.details.get("http_status")
+            if status and 400 <= status < 500:
+                return {"results": [], "meta": {}}
+            raise
--- a/scripts/data/update_literature_map.py
+++ b/scripts/data/update_literature_map.py
@@ -474,6 +474,20 @@ def merge_rows(
     return list(by_id.values()), added
 
 
+async def resolve_doi_work_id(client: OpenAlexClient, doi: str) -> str:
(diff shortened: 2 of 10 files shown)


scripts/data/update_literature_map.py
```diff
@@ -0,0 +1,662 @@
...
+            )
+            parsed = [parse_work(w) for w in works]
+            query_counts[f"source:{label}:{source_id}"] = len(parsed)
+            existing, added = merge_rows(existing, parsed, hint_area=hint)
+            added_total += added
+
+        for doi in ANCHOR_DOIS:
+            cited = await _works_pages(
+                client,
+                {"filter": f"cites:doi:{doi},{year_filter}"},
```
hollomancer: OpenAlex’s `cites` filter expects an OpenAlex Work ID (`W…`), not a `doi:` value. I verified this exact request shape against the API; it returns HTTP 400 with `'doi:10.1257/aer.20150491' is not a valid OpenAlex ID.` Because `search_works()` converts that error to an empty page, every anchor DOI currently contributes zero citing works without failing the workflow. Please resolve each anchor DOI to its Work ID first, then query `cites:W…`, and cover that request shape with a regression test.

sbir_etl/enrichers/openalex_client.py
```diff
@@ -195,3 +195,18 @@ async def lookup(self, name: str) -> OpenAlexRecord | None:
...
+
+        Used by the exploratory literature-map refresh. Returns the raw JSON
+        object (``results``, ``meta``). Empty ``results`` on 4xx; propagates
+        :class:`APIError` on 5xx.
+        """
+        try:
+            return await self._make_request("GET", "works", params=self._with_mailto(params))
+        except APIError as e:
+            status = e.details.get("http_status")
+            if status and 400 <= status < 500:
```
hollomancer: Returning an empty result for every 4xx turns malformed filters (including the invalid `cites:doi:…` request in this PR) and exhausted 429s into successful-looking zero-hit queries. That makes the weekly job unable to distinguish “no literature” from “the query was rejected.” Please propagate 400/401/403/429 (or at least all errors other than an explicitly expected 404) so the refresh fails visibly and does not publish a misleading status report.

hollomancer (commented):
Two correctness issues need attention before the scheduled refresh can be relied on. The current anchor-citation path is invalid at the API boundary, and the client then masks that failure as an empty result set.

scripts/data/update_literature_map.py
hollomancer: Resolved: each ANCHOR_DOI is looked up with `filter=doi:…` and we query `cites:W…`. A missing DOI now fails the refresh instead of contributing zero hits.

sbir_etl/enrichers/openalex_client.py
hollomancer: Resolved: `search_works` now returns empty only on 404. 400/401/403/429 and other client errors propagate as `APIError` so the weekly job fails visibly.

PR #668, opened 2026-08-20
Add exploratory headcount-at-award readout

hollomancer:
## Summary

- add an exploratory, non-citable headcount-at-award generator over the canonical SBIR.gov bulk materialization
- report schema and agency-year coverage, the cap-slackness distribution, top firms, near-cap firms, mechanical >500 anomaly buckets, and multi-award trajectories
- use the existing `PRELOAD_V1` canonical firm merge policy without introducing another matching implementation
- add Phase I-gated agency repeat-award proxies and an auditable all-firm repeat-award cohort

## Validated repeat-award result

The all-firm output contains 25,509 unique firms anchored on their first chronological award with parsed headcount. Headcount has a positive but modest relationship with later award incidence (Pearson 0.0516; Spearman 0.2202), total observed award count (0.0878; 0.3203), and Phase II/III award share (0.0510; 0.2067).

The relationship survives two chronology/censoring checks:

- requiring headcount on the actual first award: Pearson 0.0518; Spearman 0.2206
- restricting to firms first observed through 2020: Pearson 0.0468; Spearman 0.2081
- residualizing within first-headcount award year: Pearson 0.0418; Spearman 0.1860

These are award-history proxies, not application win rates. The source contains no unsuccessful proposals.

## Cap-removal interpretation

The readout does not extrapolate a numerical policy effect from winner-only data. With a fixed award count, adding newly eligible applicants increases the denominator and mechanically lowers the aggregate applicant win rate. Estimating effects for newly eligible or incumbent firms requires proposal-level applicants and rejections, affiliate-adjusted size, entrants above the cap, and a budget/allocation response.

## Verification

- full generator run over 219,497 retained SBIR.gov rows
- 25,509 unique firm-proxy rows; band totals reconcile; repeat flags reconcile to later award counts
- `ruff check` and `ruff format --check`
- `pytest tests/unit/phase_transition/test_sbir_gov_source.py tests/unit/identity/test_canonical_merge.py` (54 passed)

Generated CSV, Markdown, parquet, and SVG outputs remain ignored runtime artifacts under `data/reports/`; this PR adds the deterministic generator only.

### DIFF
--- a/scripts/data/build_headcount_at_award_readout.py
+++ b/scripts/data/build_headcount_at_award_readout.py
@@ -1,0 +1,1227 @@
+#!/usr/bin/env python3
+"""Build an exploratory headcount-at-award readout from SBIR.gov awards.
+
+Epistemic tier: exploratory / non-citable.
+
+This script:
+1. Materializes the canonical SBIR.gov source parquet from the checked-in
+   ``data/raw/sbir/award_data.csv`` snapshot.
+2. Profiles the employee-count field coverage in the actual materialized schema.
+3. Rolls awards to firms using the repository's existing canonical merge policy
+   (UEI primary, DUNS fallback, normalized-name last).
+4. Writes CSV artifacts, a Markdown readout, and a two-panel histogram figure.
+"""
+
+from __future__ import annotations
+
+import argparse
+import math
+import re
+from collections import Counter
+from dataclasses import dataclass
+from html import escape
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import pandas as pd
+
+from sbir_analytics.assets.phase_transition.sbir_gov_source import (
+    materialize_sbir_gov_history,
+    verify_sbir_gov_materialization,
+)
+from sbir_etl.identity import (
+    CanonicalMergePolicy,
+    CompanyNameProfile,
+    build_canonical_company_map,
+    normalize_company_name,
+)
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+RAW_SBIR_PATH = REPO_ROOT / "data/raw/sbir/award_data.csv"
+MATERIALIZED_SBIR_PATH = REPO_ROOT / "data/processed/phase_iii_census_sbir_awards.parquet"
+DEFAULT_OUTPUT_DIR = REPO_ROOT / "data/reports/headcount_at_award"
+FIGURE_NAME = "headcount_at_award_distribution.svg"
+MARKDOWN_NAME = "headcount_at_award_readout.md"
+SOURCE_URL = "https://data.www.sbir.gov/mod_awarddatapublic/award_data.csv"
+ANALYSIS_DATE = "2026-08-20"
+
+NONPROFIT_NAME_TOKENS: tuple[str, ...] = (
+    "UNIVERS",
+    "COLLEGE",
+    "FOUNDATION",
+    "HOSPITAL",
+    "INSTITUTE",
+    "INSTITUTION",
+    "LABORATORY",
+    "LABORATORIES",
+    "NONPROFIT",
+    "NON PROFIT",
+    "SCHOOL",
+)
+
+PHASE_SORT_ORDER = {"I": 1, "IB": 2, "II": 3, "IIB": 4, "III": 5}
+PHASE1_FAMILY = frozenset({"PHASE I", "I", "PHASE IB", "IB"})
+PHASE23_FAMILY = frozenset({"PHASE II", "II", "PHASE IIB", "IIB", "PHASE III", "III"})
+
+
+@dataclass(frozen=True)
+class OutputPaths:
+    root: Path
+
+    @property
+    def analysis_dir(self) -> Path:
+        return self.root / "analysis"
+
+    def ensure(self) -> None:
+        self.root.mkdir(parents=True, exist_ok=True)
+        self.analysis_dir.mkdir(parents=True, exist_ok=True)
+
+    def csv(self, name: str) -> Path:
+        return self.root / f"{name}.csv"
+
+    @property
+    def figure(self) -> Path:
+        return self.analysis_dir / FIGURE_NAME
+
+    @property
+    def markdown(self) -> Path:
+        return self.root / MARKDOWN_NAME
+
+
+def _parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--raw-sbir", type=Path, default=RAW_SBIR_PATH)
+    parser.add_argument("--materialized-sbir", type=Path, default=MATERIALIZED_SBIR_PATH)
+    parser.add_argument("--output-dir", type=Path, default=DEFAULT_OUTPUT_DIR)
+    return parser
+
+
+def _clean_text(value: object) -> str:
+    if value is None or pd.isna(value):
+        return ""
+    text = str(value).strip()
+    return "" if text.lower() in {"", "nan", "none", "<na>"} else text
+
+
+def _parse_headcount(value: object) -> int | None:
+    """Mirror the repository's employee-count coercion deterministically."""
+
+    text = _clean_text(value)
+    if not text:
+        return None
+    cleaned = text.replace(",", "")
+    try:
+        parsed = float(cleaned)
+    except ValueError:
+        digits = "".join(ch for ch in cleaned if ch.isdigit())
+        return int(digits) if digits else None
+    if parsed < 0:
+        return None
+    return round(parsed)
+
+
+def _parse_money(value: object) -> float | None:
+    text = _clean_text(value)
+    if not text:
+        return None
+    cleaned = re.sub(r"[^0-9.\-]", "", text)
+    if cleaned in {"", "-", ".", "-."}:
+        return None
+    try:
+        return float(cleaned)
+    except ValueError:
+        return None
+
+
+def _parse_year(value: object) -> int | None:
+    text = _clean_text(value)
+    if not text:
+        return None
+    match = re.search(r"\b(19|20)\d{2}\b", text)
+    return int(match.group(0)) if match else None
+
+
+def _normalize_matching_name(value: object) -> str:
+    return normalize_company_name(value, profile=CompanyNameProfile.MATCHING_V1)
+
+
+def _original_company_key(row: pd.Series) -> str:
+    uei = _clean_text(row.get("company_uei"))
+    if uei:
+        return f"UEI:{uei}"
+    duns = _clean_text(row.get("company_duns"))
+    if duns:
+        return f"DUNS:{duns}"
+    return f"NAME:{_normalize_matching_name(row.get('company_name'))}"
+
+
+def _canonical_firm_frame(frame: pd.DataFrame) -> pd.DataFrame:
+    canonical_map = build_canonical_company_map(frame, policy=CanonicalMergePolicy.PRELOAD_V1)
+    enriched = frame.copy()
+    enriched["firm_original_key"] = enriched.apply(_original_company_key, axis=1)
+    enriched["firm_key"] = (
+        enriched["firm_original_key"].map(canonical_map).fillna(enriched["firm_original_key"])
+    )
+    enriched["firm_key_basis"] = enriched["firm_key"].str.split(":", n=1).str[0]
+    enriched["firm_canonical_uei"] = enriched["firm_key"].where(
+        enriched["firm_key"].str.startswith("UEI:")
+    )
+    enriched["firm_canonical_uei"] = enriched["firm_canonical_uei"].str.removeprefix("UEI:")
+    enriched["firm_canonical_duns"] = enriched["firm_key"].where(
+        enriched["firm_key"].str.startswith("DUNS:")
+    )
+    enriched["firm_canonical_duns"] = enriched["firm_canonical_duns"].str.removeprefix("DUNS:")
+    enriched["firm_canonical_name_key"] = enriched["firm_key"].where(
+        enriched["firm_key"].str.startswith("NAME:")
+    )
+    enriched["firm_canonical_name_key"] = enriched["firm_canonical_name_key"].str.removeprefix(
+        "NAME:"
+    )
+    return enriched
+
+
+def _phase_sort_key(value: object) -> tuple[int, str]:
+    text = _clean_text(value).upper()
+    return (PHASE_SORT_ORDER.get(text, 999), text)
+
+
+def _joined_unique(values: pd.Series) -> str:
+    unique = sorted({_clean_text(value) for value in values if _clean_text(value)})
+    return "; ".join(unique)
+
+
+def _phase_mix(values: pd.Series) -> str:
+    counter = Counter(_clean_text(value).upper() for value in values if _clean_text(value))
+    ordered = sorted(counter.items(), key=lambda item: (_phase_sort_key(item[0]), item[0]))
+    return "; ".join(f"{phase}={count}" for phase, count in ordered)
+
+
+def _representative_name(values: pd.Series) -> str:
+    counter = Counter(_clean_text(value) for value in values if _clean_text(value))
+    if not counter:
+        return ""
+    max_count = max(counter.values())
+    return sorted(name for name, count in counter.items() if count == max_count)[0]
+
+
+def _summarize_firms(frame: pd.DataFrame) -> pd.DataFrame:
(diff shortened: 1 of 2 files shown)


scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+    CanonicalMergePolicy,
+    CompanyNameProfile,
+    build_canonical_company_map,
+    normalize_company_name,
+)
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+RAW_SBIR_PATH = REPO_ROOT / "data/raw/sbir/award_data.csv"
+MATERIALIZED_SBIR_PATH = REPO_ROOT / "data/processed/phase_iii_census_sbir_awards.parquet"
```
hollomancer: **Default run overwrites the SHA-pinned evidence-tier census parquet.** `_materialize_source()` calls `materialize_sbir_gov_history()` unconditionally, so a default run of this exploratory script rewrites `data/processed/phase_iii_census_sbir_awards.parquet` and its `.checks.json` — the frozen Phase III census study's verified input (SHA `b46c552f…`, 219,500 rows pinned in `studies/phase-iii-census/materialization-2026-02-06.md`). The PR's own run retained 219,497 rows, so the author's `award_data.csv` already differs from the pinned snapshot; and even with an identical CSV, `to_parquet` bytes vary across pyarrow versions, so the pinned SHA drifts.

Suggestion: default this readout to its own scratch/output path (e.g. under `data/processed/headcount_readout/`) and reuse or verify an existing artifact instead of always re-materializing the shared census input.

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+    "INSTITUTE",
+    "INSTITUTION",
+    "LABORATORY",
+    "LABORATORIES",
+    "NONPROFIT",
+    "NON PROFIT",
+    "SCHOOL",
+)
+
+PHASE_SORT_ORDER = {"I": 1, "IB": 2, "II": 3, "IIB": 4, "III": 5}
```
hollomancer, line 68: **`PHASE_SORT_ORDER` keys never match the real phase values, so the phase tiebreaker is inert.** The materialized parquet keeps raw SBIR.gov phase strings like `'Phase I'` / `'Phase II'` (`sbir_gov_source._normalize_phase` only strips a `'PHASE '` prefix on the uppercased comparison; the stored values and the tests use the `'Phase I'` form), but this dict is keyed by `'I'`, `'IB'`, `'II'`, … — so `_phase_sort_key` returns 999 for every real row.

Consequence: for a firm with a Phase I and a Phase II award in the same year and no `award_date`, ordering falls through to `award_record_key` (an `award_id|phase|sha256` string), so the Phase II can arbitrarily sort as the "first" award — corrupting `first_headcount_phase`, `first_award_has_headcount`, the first-Phase-I selection in `_phase1_proxy_by_agency`, and the trajectory first/last rows. Note `PHASE1_FAMILY`/`PHASE23_FAMILY` correctly include both `'PHASE I'` and `'I'` forms — this dict needs the same treatment (or normalize the phase before lookup).

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+            observed["headcount"].gt(500)
+            & observed["prev_headcount"].lt(100)
+            & observed["next_headcount"].lt(100)
+        )
+        if mask.any():
+            result.loc[observed.index[mask]] = True
+    return result
+
+
+def _build_crosscheck_summary() -> pd.DataFrame:
```
hollomancer: **The crosscheck "table absent" result is a hardcoded constant, not a check.** `_build_crosscheck_summary` states "No materialized FPDS/SAM employee-count table was present in this checkout" and the markdown says joins "were attempted", but the function never touches the filesystem — no `Path.exists()`/glob anywhere. Run this in a checkout that *does* contain a materialized FPDS/SAM employee-count table and the readout still claims it was absent and that joins were attempted — both false.

Either perform the actual existence check and report its result, or reword the output to say the cross-check was not implemented in this readout.

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+    """Mirror the repository's employee-count coercion deterministically."""
+
+    text = _clean_text(value)
+    if not text:
+        return None
+    cleaned = text.replace(",", "")
+    try:
+        parsed = float(cleaned)
+    except ValueError:
+        digits = "".join(ch for ch in cleaned if ch.isdigit())
```
hollomancer: **The digit-concatenation fallback fabricates large headcounts that land unflagged in the >500 tail this analysis is about.** A value like `'10-50'` fails `float()`, so all digits are concatenated → 1050 (verified empirically); `'5 to 10'` → 510. Such records enter `anomalies_over_500` as "unclassified" and inflate the >500 share, top-tail counts, and the histogram range. Mirroring `sbir_etl/models/award.py`'s validator is a fine default for the ETL, but for a tail-focused readout these rows need at least a parse-method flag (or their own anomaly bucket) so fabricated values are distinguishable from genuinely reported >500 headcounts.

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+                "firm": _representative_name(ordered["company_name"]),
+                "uei": _clean_text(ordered["firm_canonical_uei"].iloc[0]),
+                "award_count": int(len(ordered)),
+                "headcount_observation_count": int(len(observed)),
+                "first_award_year": ordered["award_year"].iloc[0],
+                "first_award_headcount": ordered["headcount"].iloc[0],
+                "max_headcount": int(observed["headcount"].max()),
+                "latest_award_year": ordered["award_year"].iloc[-1],
+                "latest_award_headcount": ordered["headcount"].iloc[-1],
+                "crossed_350": bool(
```
hollomancer: **`crossed_350` ignores ordering, so a shrinking firm is flagged as having crossed 350.** The check is `observed['headcount'].lt(350).any() and .ge(350).any()` with no ordering constraint — a firm with headcounts [420, 300, 90] over time gets `crossed_350=True` and sorts to the top of `trajectory_multi_award_firms.csv` as if it grew through the near-cap band, when it actually contracted. Either require the ≥350 observation to come after a <350 one (the frame is already time-sorted here), or rename the flag to reflect "spanned the 350 band in either direction".

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+RAW_SBIR_PATH = REPO_ROOT / "data/raw/sbir/award_data.csv"
+MATERIALIZED_SBIR_PATH = REPO_ROOT / "data/processed/phase_iii_census_sbir_awards.parquet"
+DEFAULT_OUTPUT_DIR = REPO_ROOT / "data/reports/headcount_at_award"
+FIGURE_NAME = "headcount_at_award_distribution.svg"
+MARKDOWN_NAME = "headcount_at_award_readout.md"
+SOURCE_URL = "https://data.www.sbir.gov/mod_awarddatapublic/award_data.csv"
+ANALYSIS_DATE = "2026-08-20"
```
hollomancer: **`ANALYSIS_DATE` is hardcoded, so every future run misdates its output.** Re-run this in 2027 against a refreshed `award_data.csv` and the readout header still claims "Generated on 2026-08-20" (the crosscheck notes repeat the same frozen date). The manifest already carries a truthful `generated_at` timestamp (`materialize_sbir_gov_history` sets `datetime.now(UTC)`) — use that in the markdown instead of a baked-in constant.

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
+#!/usr/bin/env python3
+"""Build an exploratory headcount-at-award readout from SBIR.gov awards.
```
hollomancer, line 2: **Notebook-first workflow: the still-changing statistical design belongs in a notebook.** CLAUDE.md ("Research and analysis workflow") says to use a notebook-first workflow "when the research question, cohort definition, matching rule, statistical assumption, or visualization is still changing", and to check whether a bounded notebook would answer the open questions before implementing an uncertain analysis directly as a large Python script. The repeat-award proxy design, sensitivity cohorts, and fixed-effect residualization here are exactly such still-changing choices; embedding them in a ~1,200-line monolithic generator makes every methodological iteration a code change instead of a notebook cell. The deterministic artifact-generation portion (schema/coverage CSVs) is script-appropriate — consider splitting the exploratory statistics into a notebook per the stated workflow, or noting why the notebook check was passed over.

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+    if x_residual.nunique() < 2 or y_residual.nunique() < 2:
+        return pd.NA
+    return float(x_residual.corr(y_residual, method="pearson"))
+
+
+def _all_firm_repeat_proxy(
+    frame: pd.DataFrame,
+) -> tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+    rows: list[dict[str, Any]] = []
+    for firm_key, group in frame.groupby("firm_key", sort=False):
```
hollomancer: **Efficiency: five per-firm Python loops each re-sort the same groups by the same four keys.** `_summarize_firms`, `_mark_single_award_spikes`, `_all_firm_repeat_proxy`, `_phase1_proxy_by_agency`, and the trajectory loop each run `sort_values(['award_sort_date','award_year','phase_sort_key','award_record_key'])` once per firm (~25k+ groups each), and `enriched.apply(_original_company_key, axis=1)` is a Python-level row loop over 219k rows. Sorting the frame once globally by `[firm_key, award_sort_date, award_year, phase_sort_key, award_record_key]` and grouping the pre-sorted frame with `sort=False` (order is preserved within groups) removes all per-group sorts; the original key is derivable vectorized with `np.select` over the three cleaned columns. Exploratory tier, so purely a nice-to-have.

---

scripts/data/build_headcount_at_award_readout.py
```diff
@@ -0,0 +1,1227 @@
...
+            float(headcounts.quantile(0.90)),
+            float(headcounts.quantile(0.99)),
+            float((headcounts >= 350).mean()),
+            float((headcounts >= 500).mean()),
+        ],
+    }
+    return pd.DataFrame(metrics)
+
+
+def _safe_corr(x: pd.Series, y: pd.Series, method: str) -> float | pd.NA:
```
hollomancer: **`float | pd.NA` is an invalid type expression** — `pd.NA` is a value (an `NAType` instance), not a type. It's harmless at runtime only because of `from __future__ import annotations`; anything that resolves the annotations (`typing.get_type_hints`, runtime validators, future mypy scope inclusion) raises `TypeError`. Same on `_year_fixed_effect_corr`. Spell it `float | NAType` (from `pandas._libs.missing`) or just document the NA return in the docstring.

---

hollomancer (commented):
Reviewed `scripts/data/build_headcount_at_award_readout.py` (single-pass inline review). 9 inline comments, most severe first:

1. **Evidence-tier clobbering** — the default run unconditionally re-materializes the SHA-pinned Phase III census parquet and its manifest (line 43).
2. **Inert phase tiebreaker** — `PHASE_SORT_ORDER` keys never match the `'Phase I'`-form values, so "first award" selection can arbitrarily order same-year awards (line 64).
3. **Fabricated crosscheck claims** — the "FPDS/SAM table absent / joins attempted" text is a hardcoded constant, not a check (line 272).
4. Digit-concat parse fallback turns `'10-50'` into 1050 employees and feeds it unflagged into the >500 anomaly tail (line 118).
5. `crossed_350` is order-blind — shrinking firms are flagged as crossers (line 1145).
6. Hardcoded `ANALYSIS_DATE` misdates future runs (line 48).
7. CLAUDE.md notebook-first workflow concern for the still-changing statistical design (line 2).
8. Efficiency: five per-firm loops re-sorting identical groups (line 490).
9. Invalid `float | pd.NA` annotations (line 459).

Verified against the PR head: parquet schema and phase normalization in `sbir_gov_source.py`, canonical-company-map key compatibility, and the pandas edge cases empirically (several plausible crash candidates were dropped because they don't actually fail).

---

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. The default now writes to the analysis-local `data/processed/headcount_at_award/sbir_gov_awards.parquet`, never the Phase III evidence input. If a materialization already exists, the script reads and verifies it rather than rewriting it.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. The Markdown date now derives from the verified materialization manifest's `generated_at` value, with `unknown` only if the manifest lacks it.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. Award ordering is now performed as stable whole-frame sorts before grouped iteration, and firm-key construction is vectorized instead of row-wise `apply`. A full run over 219,497 awards completed successfully and regenerated all tables and the figure.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. Parsing now accepts pure numerics or exactly one embedded numeric token, records the parse method, and rejects ranges/multiple numeric tokens as ambiguous instead of concatenating digits. Parse provenance is included on anomaly records, with regression tests for `10-50` and `5 to 10`.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. The hard-coded unavailable result is gone. The output now mechanically discovers locally materialized FPDS/SAM employee-count candidates and reports `source_not_found` or `source_detected_join_not_implemented`; agree/disagree are `NA` unless a supported join is actually run, so the readout no longer claims an attempted cross-check.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. `_phase_sort_key` now canonicalizes materialized labels such as `Phase I`/`Phase II` before applying the ordinal. The full-data rerun regenerated the agency tables; overall all-firm correlations are unchanged, while agency-level Phase I repeat rates changed where same-date phase ordering had mattered.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. `crossed_350` now requires a chronological below-350 observation followed by a later observation at or above 350; reverse trajectories and firms always above the threshold remain false. Regression tests cover all three cases.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Addressed in `aea9b278`. Both helpers now return `float | None`, matching the actual runtime sentinel and avoiding `pd.NA` as a type. A regression test resolves the annotations with `typing.get_type_hints`.

scripts/data/build_headcount_at_award_readout.py
hollomancer: Decision documented in `aea9b278`: this file remains the canonical repeatable artifact generator because the criteria in this PR are already fixed and deterministic. The module now explicitly requires future cohort/statistical-design changes to go notebook-first; I did not add a second notebook implementation that could drift from the generator.

hollomancer:
Review follow-up (`aea9b278`): all nine review threads are addressed and resolved. Validation: 8 focused tests pass; Ruff check/format and `git diff --check` pass; a full run over 219,497 materialized awards regenerated all CSVs and the figure. The all-firm correlations are unchanged. Correcting `Phase I`/`Phase II` ordering updates agency-level Phase I repeat-rate results where same-date phase ordering mattered (largest reviewed shift: DOE, 60.72% to 62.79%), so downstream quotations should use the regenerated agency table. The readout remains exploratory/non-citable.

PR #669, opened 2026-08-21
Add exploratory supplier-share census

hollomancer:
## Summary

- add an exploratory, non-citable 2x2 census of observed federal persistence versus observed venture signals across the frozen `T x N x window` grid
- fail closed on Form D/EFTS noncoverage, malformed or duplicate checkpoints, and incomplete directional M&A refinement
- add the explicitly post-result Revision 4 deterministic typed-noncoverage interval while keeping incomplete point headlines null
- keep the private 50-firm sample withheld until every required channel is measurable

## Final repaired source cut and result

**Exploratory / non-citable. The frozen point headline remains suppressed.** Revision 4 is a disclosed post-result amendment. Its lower endpoint counts `persistent_no_venture`; its upper endpoint additionally counts `persistent_unknown_venture`. This is a deterministic partial-identification interval for typed required-channel noncoverage only—not a confidence interval, imputation, preregistered result, or physical supply-chain/chokepoint claim.

The approved repair-and-freeze run is complete:

- repaired EFTS roster: 34,460 rows and 34,460 unique source labels; no duplicate, missing, or search-error rows; 132 labels retain typed filing-document noncoverage
- rebuilt base M&A artifact: 5,252 unique events (1,679 high, 2,050 medium, 1,523 low), including 3,198 direction-sensitive events
- fresh strict directional pass: exactly 3,198 typed verdicts; 3,085 complete and 113 explicitly `context_incomplete`
- exact-coverage apply: 5,252 final events (1,679 high, 1,718 medium, 1,855 low), with zero signal or confidence derivation mismatches
- central cell `(T=10, N=6, window=15)`: 20,049 mature firms; 19,853 measurable (99.02%); point firm and dollar shares remain null
- Revision 4 central interval: **24.10%-24.33% of firms** and **67.10%-67.42% of cumulative observed SBIR/STTR dollars**; 47 persistent firms carrying **$184.3M** move the upper endpoint

The validation state is `blocked_missing_required_signal_inputs`. The approved one-reviewer hand audit remains conditional: no named 50-firm packet was emitted, so no human review was started and the gate was not overridden.

## Reproducibility and provenance

Two independent census runs produced byte-identical Parquet, CSV, Markdown, and SVG outputs. Their manifests differ only in output-directory paths.

- repaired EFTS SHA-256: `c9c823e43a7f406ecf8550be9590f3ecce4f66be1b1a8866c32078f6ca200c03`
- base M&A SHA-256: `43f42633729fda4c6040aa2991dd3f76cbf0aa0a78970c6d0ff34bde08ff4311`
- strict refinements SHA-256: `a0ff6113207d7a445e5715bf1fded2bc12ab4838a1a2d629d1bb163a2d56815e`
- final M&A SHA-256: `2017bffdd51c2f201428033a174bfd23fc3dce1202f69f5e82b9a999e55792f2`
- authoritative firm-grid SHA-256: `f2ee98c9da5c2a6f6e86ee0d20ae9a1b3e114e52ab0ac5cfdb803df06c4c387b`
- summary SHA-256: `bf4a23f319092842d72127552d2d315395eace96dba875940f94148dd21b4b0c`
- readout SHA-256: `53dc332b3c6b0776eb5d73bb73061e04c5e3076e793911af878ea603fbbe6738`

The private source-cut archive retains the original and repaired checkpoints, effective run parameters, timestamps, code/input/output hashes, row audits, and the exact gate state. Its manifest SHA-256 is `0c4480de2f83fcade6b05cd552711024931cb963f46ef112518f7757427641f5`.

This is intentionally documented as a mixed retrieval window: 34,328 clean August 21 checkpoint rows were retained, 132 error/incomplete rows were retried on September 10, and the 3,198-event directional pass was run on September 10. With explicit approval, only public company names and the configured contact email were sent to SEC EFTS/document endpoints. This is not represented as a uniform historical SEC snapshot.

## What is implemented

- atomic, idempotent SEC scan and refinement checkpoints, including malformed-tail healing and typed retry state
- strict SEC client error propagation for coverage-sensitive callers
- anchored/passive-only M&A target classification with adversarial buyer, seller, comparator, and unrelated-deal coverage
- an atomic apply bridge that requires exactly one structurally valid verdict for every direction-sensitive event and recomputes the frozen confidence rule
- census quarantine for explicit and legacy-untyped direction-sensitive noncoverage
- Revision 4 firm/dollar lower and upper endpoints, widths, and mover counts/dollars for every nonempty mature grid/stratum total; matrix bound fields remain null
- stale private-sample removal on blocked runs and atomic mode-`0600` replacement on eligible runs

## Validation and merge state

- Head `2906df18` is conflict-free and mergeable.
- All routed GitHub lint, type, guard, fast-test, hermetic E2E, security, CodeQL, and version-policy checks pass; intentionally routed full/Neo4j/Docker/setup jobs are skipped.
- Revision 4 passed independent quality and specification audits. The focused census suite is 22/22; the broader repaired SEC/M&A/census neighborhood is 182/182.
- The Amendment 4 freeze hashes match the implementation, strict bound reconciliation rejects corrupted shares, and repeated census outputs are deterministic.

## Epistemic limits

The study measures an observed sustained-federal-performer construct. It does not establish commercialization failure, dependence, technical indispensability, supply-chain embeddedness, or a physical chokepoint. Revision 4 does not bound false negatives in apparently searched sources, identity error, non-Reg-D capital, prime/sub-tier undercoverage, or any of those broader concepts. `citable=false` remains mandatory; evidence promotion would require a separate approved contract.


hollomancer:
Fresh review (2026-08-23): no new blocking code findings. I reran the focused 100-test supplier/SEC/refinement suite, Ruff, format checks, and both frozen-spec SHA checks; all pass. Decision: keep this PR draft and exploratory/non-citable. The identification interval in the PR description is a supplemental bound over unresolved firms, not a promoted headline; required-channel coverage, hand adjudication, anchors, and negative-control review remain promotion gates.

hollomancer:
## Evidence audit: the census is good; the `sbir_etl/` changes riding along are not

**The census itself is well-built exploratory work.** `requirements.md:12` declares `exploratory`, and the design honors it: `citable=false` on outputs, typed absence reasons (`no_filing_found` / `not_searchable` / `window_censored`), suppression of the headline when channels are unsearchable, a full sensitivity grid rather than a single cell, and a "Exploratory and non-citable" Status in `docs/research-questions.md` that reads as a refusal, so `check_research_question_status.py` won't demand a study contract. No complaints there.

### Blocking: an undeclared behavior change in pipelines-tier M&A detection

`sbir_etl/enrichers/sec_edgar/client.py:273-274` adds to the EFTS request:

```python
"from": 0,
"size": limit,
```

Previously no `size` was sent, so the endpoint returned its server-side default page and `hits[:limit]` could only *truncate*, never extend. Now a `limit=100` call actually retrieves up to 100 hits. **This changes the mention set returned for every company**, in a module whose tier contract requires determinism and reproducibility.

`enricher.py:453` then asserts:

> `# that the optimization cannot change the retained event or its tier.`

That comment is false in combination with the client change. And `enricher.py:541-547`:

```python
if len(combined) < _EFTS_PAGE_SIZE:
    strong_mentions = [m for m in combined if m.get("form_type") in _MA_FORM_TYPES][:20]
    annual_mentions = [m for m in combined if m.get("form_type") in _ANNUAL_FORM_TYPES][:10]
```

replaces two form-restricted EFTS queries with one combined query, filtered and sliced locally. When the complete MA-form hit set exceeds 20 (or annual exceeds 10), `[:20]` selects by *combined* ordering rather than the form-restricted query's own ranking. Not the same 20 filings.

To be fair to the comment: I traced the `latest_by_filer` dedup at `enricher.py:454,480-485` against the downstream dedup at `:769-774` and that part **is** equivalent — same key, same strict `>` tie-break, same concatenation order. So half the claim is defensible. The `size` and slicing changes are not.

**Why this reaches beyond this PR:** this is the code path that produced the 1,197 high / 1,593 medium / 2,790 total figures that #672 is depositing on Zenodo. Merging this means repo `HEAD` no longer reproduces the preprint's numbers. #672's Appendix A pins commit `241fe9d`, which protects the paper — but `CITATION.cff:12` `repository-code` is unpinned, so readers following it get a method that has silently moved.

`requirements.md:23-24` says the producer "SHALL consume the existing … M&A high-plus-medium tiers **without changing their thresholds**." The thresholds are unchanged; the *inputs to the tiering* changed, so the tiers change. Satisfied in letter, violated in substance.

### Secondary: exploratory work reaching into pipelines internals

`enricher.py:349` and `scripts/archive/data/refine_ma_medium_tier.py`:

```python
callback = getattr(client, "__dict__", {}).get("_context_incomplete_callback")
```

An undeclared private attribute poked onto `EdgarAPIClient` and read via `__dict__` specifically to bypass the class interface. `epistemic-tiers.md:100` forbids exploratory work being depended on by another tier; this is the mirror image — exploratory needs driving pipelines-tier design through a hook that is in no contract and cannot be type-checked.

### Remediation
1. **Split the `sbir_etl/enrichers/sec_edgar/` changes into their own PR**, with a stated before/after effect on high/medium tier counts. A pipelines-tier behavior change shouldn't ride along with an exploratory census.
2. Correct or delete the `enricher.py:453` comment.
3. Replace the `client.__dict__` callback with an explicit optional parameter or a declared attribute.
4. Coordinate merge order with #672 — if the enricher change lands first, `study.yaml`'s `identity_policy.version: sec-edgar-ma-pr286-two-tier` no longer describes the code at `HEAD`.

Correctly left in draft. The census half is ready to go once the enricher half is extracted.

---

hollomancer:
## scope-guard second pass — TRIM

Ran this through `scope-guard` on top of the existing evidence audit and verified every claim against the diff myself. **It independently reached the same blocking conclusion**, which is worth stating: two reviewers with different briefs both landed on `client.py:273-274` as the thing that stops this merging. Not restating that finding — four *new* ones below, plus what I checked and found clean.

Confirmed the core defect directly:

```diff
             "forms": forms,
+            "from": 0,
+            "size": limit,
```

On `main` no `size` was sent, so `hits[:limit]` could only truncate. It now changes the returned mention set for **every** caller of `search_filing_mentions`.

### New findings

**1. `pyproject.toml` deliberately exempts the trees this PR reformatted.** `[tool.ruff.format] exclude = ["scripts/**", ...]`, with the comment: *"per docs/steering/epistemic-tiers.md, exploratory work is not held to a house format."* The `style: format archived EFTS scanner` commits reformat `scripts/archive/data/*` anyway — no CI gate asked for it. This is not cosmetic-only in effect: it mixes formatting churn into the same hunks as the behavior change above, which is a large part of why a two-line recall change reads as incidental. Revert the formatting-only hunks and the real diff becomes reviewable.

**2. Archived scripts are edited and, unusually, given tests.** `scripts/archive/README.md` says archived scripts exist "for reproducing published analyses" and that nothing in CI, the Makefile, or Dagster invokes them. This adds 302 lines to `scan_sbir_edgar.py`, 47 to `refine_ma_medium_tier.py`, and **88 lines of new tests** under `tests/unit/scripts/archive/`. Editing an archived script degrades the reproduction guarantee it was archived to provide, and putting the deadest code in the repo under CI is the clearest over-build here. Decide explicitly: if these are being revived, un-archive them into `scripts/data/`. If they are one-shot, run them and drop the tests.

**3. A machine-specific absolute path is a committed default.** `build_supplier_share_census.py:42`:

```python
"/Volumes/SSDmini/sbir-analytics/data/transition/contracts_ingestion.parquet"
```

A macOS external volume. `tasks.md` item 5, "Run available materialized inputs," is only verifiable on one laptop. Require `--contracts-path`, or route it through `sbir_etl/config/loader.py` like every other configured path.

**4. Over-build inside the exploratory tier, two places.** ~85 lines of hand-written SVG string concatenation at `:1425-1510` when `matplotlib>=3.10.8` is already a dependency. And the "placebo statistic" at `:1074-1106` is a single deterministic hash-ordered relabeling — no null distribution, no p-value — so it produces an uninterpretable number today. `tasks.md` P4 is literally "Approve a negative-control threshold and rerun"; defer it there, where its threshold gets defined.

**5. Lower confidence, worth an assertion.** `enricher.py:537-545` partitions the combined result by exact string membership in `_MA_FORM_TYPES` / `_ANNUAL_FORM_TYPES`. Those strings are the API's `forms=` request vocabulary, but `form_type` comes from the response's `root_forms` (`client.py:288-290`). If EDGAR normalizes `DEFM14A`, `SC TO-T` or `SC 14D9` differently there, those mentions fall out of *both* partitions on the fast path while the `else` path still finds them — a silent recall drop that varies with result-set size. I did not verify EDGAR's `root_forms` vocabulary. Cheap guard: assert the partition is exhaustive.

### Checked and clean — stating these so the finding list is not mistaken for the whole picture

- **Primitive reuse is correct.** `normalize_company_name`, `build_canonical_company_map`, `CanonicalMergePolicy` all come from `sbir_etl.identity` with named profiles (`FORM_D_JOIN_V1:183`, `MATCHING_V1:194`). No forked normalizer, no second config reader.
- **No silent promotion.** `data/reports/` is gitignored, the readout stamps `citable=false`, the notebook asserts `not summary["citable"].any()`, and the F2 Status reads "Exploratory and non-citable" — a refusal, not a rank dodge.
- **No duplicate computation path.** `nano_ma_signal.py` and `sbir_ma_signal_counts_by_fy.py` answer different questions.
- **The freeze machinery should stay**, even at exploratory tier. `_verify_freeze` plus the "Visibility at approval" sections are the one piece of evidence machinery that cannot be retrofitted — once you have seen the results, a freeze is worthless. Same for the 10 safety-property tests guarding AC4's "never a valid-looking zero"; those are not coverage-chasing.
- **The notebook is the model the other companions should follow**: zero outputs, no execution counts, canonical artifacts only, degrades to a "run the generator first" error.

### The one thing worth acting on before spending more time

Get `sbir_etl/enrichers/sec_edgar/{client,enricher}.py` out of this PR. Everything else above is trimming. That one changes M&A search recall for every existing consumer, contradicts this spec's own "without changing their thresholds" constraint in substance if not in letter, and — as the earlier audit noted — it is the code path that produced the figures #672 is depositing on Zenodo.

The census half is sound. It is being held up by a ~600-line side quest.

---

hollomancer:
Re-verified both open audits against head `829a3a73`. No code pushed here; the split decision below is the author's.

**Audit claims, checked against this head**

| Claim | Status | Evidence |
|---|---|---|
| `client.py` adds `"from": 0, "size": limit` | PARTIAL | New vs `main` (`client.py:273-274`). Default `limit=20` and the EFTS server default page is 100, so callers with `limit <= 100` get the same set given stable server ordering. The enricher now requests exactly 100 (`enricher.py:527`). A real pipelines-tier change, narrower than stated. |
| `enricher.py:453` comment is false | PARTIAL | The per-filer latest-only pre-dedup it describes does hold (`enricher.py:827-835`). The combined fast path (`:537-583`) has no such guarantee; it assumes combined-query ordering equals form-restricted ordering. |
| `__dict__` callback | HOLDS | `enricher.py:349`, `refine_ma_medium_tier.py:95`, set at `scan_sbir_edgar.py:72`. |
| Formatting churn in `scripts/archive/data/` | HOLDS | 171 of 336 changed lines are pure reformat (refine: 27 of 45; scan: 144 of 291). |
| Machine-specific default path | HOLDS | `build_supplier_share_census.py:41-43`. |
| Form-type partition not exhaustive | PARTIAL | `enricher.py:542-547` drops any hit whose `form_type` is empty or outside the two frozensets; the slow path would have kept it. Nothing logs the remainder. |
| Hand SVG and single-relabeling placebo | HOLDS | `:1444-1508`; `_placebo_supplier_share` `:1078-1104`. |

**New findings on the census script**

1. `build_supplier_share_census.py:767-791` re-derives M&A confidence with a rule no producer uses. It expects `ma_definitive` + `direction == "target"` for medium, but `detect_sbir_ma_events.py:147-162` assigns form_d/subsidiary high, acquisition medium, else low, and never writes `direction`; `refine_ma_medium_tier.py` writes `direction` to a separate file. Any real events file with an acquisition-text medium event trips `confidence_mismatches`, `derivation_consistent` goes false, every `ma_searchable` goes false, and every headline is suppressed with the manifest blaming upstream derivation. Fail-closed rather than a false zero, but the headline path is unreachable from repo-produced inputs. CONFIRMED.
2. `--ma-search-complete` (`:130-134`) claims to assert coverage "when no per-firm EFTS scan artifact is supplied", but `derivation_consistent` (`:811-825`) requires both the scan and Form D files to exist. Without a scan file the flag is a no-op. CONFIRMED.
3. With the `/Volumes/SSDmini/...` default absent, persistence silently degrades to tenure/count only. The manifest records `available: False` but classification proceeds, so the same inputs yield different matrix cells on a different host. `--contract-history-scope` is metadata-only. CONFIRMED.
4. Contract persistence (`:536-560`) counts any prime obligation after the first Phase II end, including later SBIR Phase I/II contracts that FPDS also carries, so winning another SBIR award fires "persistent" via the contract channel. Frozen in `design.md`, so a design question rather than a code defect.

**Checked and clean:** join grain (`firm_key` matches the canonical-map contract), denominators (`min_count=1`, all-NaN dollars yield None), maturity-gate window semantics, censoring order, ambiguous-label exclusion, `_validate_grid`, frozen-spec SHA test. 238 tests pass on this head.

**The one decision:** the `sbir_etl/enrichers/sec_edgar/{client,enricher}.py` change still rides along. Both prior audits and this pass land on the same recommendation: extract it into its own PR with a stated before/after on tier counts, and coordinate with #672, which deposits the counts that code path produced.

---

hollomancer:
The `sbir_etl/enrichers/sec_edgar/` change is extracted to #698 per the three audits. This PR now merges that branch (`75cc62d`) and its base is `feat/sec-edgar-efts-recall`, so the diff here is the census only: 11 files, 3,005 insertions, none under `sbir_etl/`. When #698 merges, GitHub retargets this PR to `main`.

#698 also carries two fixes to the extracted code: a public `EdgarAPIClient.context_incomplete_callback` attribute replaces the private hook read through `__dict__`, and the combined-query fast path now falls back to the separate per-tier searches when any hit's root form falls outside both tier vocabularies instead of dropping it. 101 census and sec_edgar tests pass on this head.

Still open on the census script itself, unchanged from the review above: the M&A confidence re-derivation that matches no producer (`build_supplier_share_census.py:767-791`), the `--ma-search-complete` no-op without a scan file, and the machine-specific default contract path that silently changes classification on another host. Those are census-side and stay with this draft.

---

hollomancer:
#698 is merged to `main` (squash `ff425bf2`), including the Copilot nits on the archived scanner regex and the 100-hit page comment. This draft should now retarget to `main`. Census-side leftovers from the earlier audit are unchanged: M&A confidence re-derivation, `--ma-search-complete` no-op without a scan file, and the machine-specific contracts path. Keep draft.

sbir_etl/enrichers/sec_edgar/enricher.py
```diff
@@ -519,6 +533,60 @@ async def _search_inbound_ma_mentions(
...
+    # request vocabulary (an empty root_forms, or a variant such as 8-K12B),
+    # so any hit that lands in neither tier also falls back to the separate
+    # searches rather than being dropped.
+    if len(combined) < _EFTS_PAGE_SIZE and all(
+        mention.get("form_type") in _MA_FORM_TYPES or mention.get("form_type") in _ANNUAL_FORM_TYPES
+        for mention in combined
+    ):
+        strong_mentions = [
+            mention for mention in combined if mention.get("form_type") in _MA_FORM_TYPES
+        ][:20]
```
hollomancer, line 552: **[bug]** When the combined EFTS page has fewer than 100 hits, `_search_inbound_ma_mentions` treats that page as the complete strong+annual set, then still keeps only `[:20]` 8-K-tier hits and `[:10]` 10-K-tier hits in combined-ranking order. Combined ordering can interleave annual filings ahead of 8-Ks, so a definitive M&A 8-K that an 8-K-only `limit=20` search would have returned can be dropped even though it was already in the downloaded page. The page-boundary fallback exists specifically to stop combined ranking from hiding a tier's top hits; this slice reintroduces that hiding on every complete page with more than 20 strong or 10 annual hits. `enrich_company` uses this path, so production inbound-M&A recall changes.

**Suggestion:** If the combined page is complete, partition and classify the full hit set (or re-run the original per-tier searches whenever either tier exceeds its old limit). Keep the 20/10 caps only for the truncated 100-hit fallback, which already restores separate searches.

scripts/archive/data/scan_sbir_edgar.py
```diff
@@ -62,16 +116,26 @@ def unregister(self, company_name: str) -> None:
         self._active_companies.discard(company_name)
 
     def write(self, message):
-        if "HTTP 5" not in message:
+        match = re.search(
+            r"EDGAR filing mention search failed for '(.*)':",
```
hollomancer: **[bug]** `_ServerErrorTracker.write` switched from non-greedy `'(.*?)':` to greedy `'(.*)':`, and the unit test that required the match to stop at the first closing quote was deleted. The client logs `EDGAR filing mention search failed for '{company_name}': {e}`. If `{e}` contains a later `':` (quoted detail, nested message), group 1 becomes `Acme Labs': later quoted 'detail` rather than `Acme Labs`, the active-company check misses, and the scan row is written without `had_server_errors`. The census then treats that label as a clean searchable negative. This is a two-dot regression versus current main.

**Suggestion:** Restore a bounded parse (non-greedy, or split on the known `failed for '` / `': ` envelope) and put back a test for both apostrophes in company names and later quoted error text. Do not delete the closing-quote regression test to paper over greedy matching.

specs/supplier-share-census/design.md
```diff
@@ -0,0 +1,234 @@
...
+The normalized summary reports:
+
+- overall;
+- agency groups `DoD`, `HHS`, `NSF`, and `other`, using each award's source agency;
+- each first-award year;
+- full-history award-count strata `1`, `2-5`, `6-20`, `21+`.
+
+Cumulative-dollar deciles are assigned over headline-eligible firms by descending firm cumulative
+award dollars, deterministic ties broken by `firm_id`. The concentration statistic is the share
+of persistent-no-venture dollars held by the first `ceil(n/10)` firms in that supplier cell. It is
```
hollomancer, line 149: **[suggestion]** Frozen design text still defines the concentration statistic as the share of supplier-cell dollars held by the first `ceil(n/10)` firms in that supplier cell, and the header still says Revision 0. Revision 1 in `amendments.md` and `_supplier_concentration` / `_cumulative_dollar_deciles` assign denominator-wide row-count deciles and use D01 of the mature population. The hashed freeze therefore documents a superseded formula next to the operative amendment.

**Suggestion:** Update the design body to Revision 1's denominator-wide D01 definition (amendments stay append-only) so the frozen prose matches the producer, then refresh `DESIGN_SHA256`.

sbir_etl/enrichers/sec_edgar/client.py
```diff
@@ -130,6 +131,12 @@ def __init__(
         )
         self.timeout = cast(int, self.api_config.get("timeout_seconds", 30))
         self.rate_limit_per_minute = cast(int, self.api_config.get("rate_limit_per_minute", 600))
+        # Optional hook the enricher calls when a mention carries an incomplete
```
hollomancer, line 134: **[suggestion]** The new `context_incomplete_callback` comment narrates why the attribute was made public (`not a private one read back through __dict__`) rather than stating the hook's contract. That is design-history leakage on a production client used by live enrichers.

**Suggestion:** Replace with one sentence on when the enricher invokes the hook (incomplete `doc_id` / accession / CIK) and that it defaults to no-op.

hollomancer:

### Status: 🔴 Blocked — Merge Conflicts

- **Draft:** Yes
- **Merge status:** Merge conflicts with `main` (DIRTY)
- **Age:** 19 days — second-oldest open PR
- **Size:** +3,602/-109 across 19 files

This is still in draft and has accumulated merge conflicts with `main`. The PR body notes the headline result is suppressed pending resolution. Needs rebase onto `main` and any open design questions closed before it can proceed.

**Recommended action:** Rebase onto main and bring out of draft when ready for review.

hollomancer:

Reviewed at `75cc62d1`.

- **Blocker — the documented rescan path creates duplicates that permanently fail coverage.** `scripts/archive/data/scan_sbir_edgar.py:241-265,551` always opens the checkpoint in append mode. A fresh rerun appends every row; `--rescan-errors` appends replacements beside stale failures. `scripts/data/build_supplier_share_census.py:693-707` then marks any label with more than one row as unclean. Truncate/atomically replace on fresh runs and replace failed records during rescans, then regenerate the pinned scan.
- **Blocker — the claimed context-incomplete refinement cannot be reproduced by this head.** `refine_ma_medium_tier.py:199-236` only invokes a callback for thrown/malformed cases, but the real EDGAR client swallows search/fetch failures into `[]`/`None`, and the CLI creates it with no callback. No path emits the 90 claimed `direction="context_incomplete"` rows. Track incomplete state per event, including swallowed fetch/search failures, emit it explicitly, and rerun the final events artifact.
- **Important — blocked runs can leave a stale private validation sample.** `_write_validation_sample()` at `build_supplier_share_census.py:1355-1363` returns `False` without removing an existing path; the manifest then says no sample was written while an earlier named sample remains on disk. Remove/replace the target atomically when the gate blocks, or use run-specific output directories.

The focused tests pass and the real-data run correctly suppresses the headline, but they do not cover these recovery/reproducibility paths.

hollomancer:
Implemented and pushed the deterministic repair pass in `da2e1803` (after merging current `main`; the prior merge conflict is resolved).

What is now fixed:

- SEC scan checkpoints are atomic and idempotent. Fresh runs clear stale output; resume/rescan compacts duplicate rows, heals malformed tails, preserves out-of-scope rows, and retries every typed/legacy incomplete marker.
- Coverage-sensitive SEC calls can opt into exceptions. The refiner now persists `context_classification_complete` and stable failure reasons instead of converting transport/document failures to measured `no_filing`/`ambiguous` results.
- Direction-sensitive selection follows source signals, including low definitive-merger rows and previously demoted acquisition rows. Target rules are anchored/passive-only and no longer promote buyers, sellers, unrelated acquisitions, generic portfolio text, or role-blind merger boilerplate.
- A new atomic apply bridge requires exactly one structurally valid verdict for every direction-sensitive event, preserves provenance, recomputes the frozen tier rule, and writes no partial artifact on failure.
- The census now quarantines explicit incomplete events and every legacy direction-sensitive event without typed-complete context. Strong independent Form-D/subsidiary signals remain exempt. A blocked private-review gate removes its stale generated sample; eligible samples replace atomically with mode `0600`.

Verification:

- Final quality sweep: 160 targeted/neighbor tests pass; Ruff/format, MyPy on the pipeline client, direct CLI imports, `make lint-boundaries`, `make docs-check`, and `git diff --check` pass.
- The broader unit run passed 6,618 tests with 7 skips before the final two narrowly scoped classifier/coverage patches; both patches are covered by the final targeted suite.
- Two full producer runs emitted byte-identical CSV, readout, SVG, and 614,574-row Parquet artifacts.
- A temporary live retry of the 132 failed SEC scan rows ended at exactly 34,460 unique rows with zero duplicates and zero search failures. All 132 still had at least one unavailable filing document, so they correctly remain typed context-unknown; no temporary data artifact was committed.

Current honest analysis state:

- Central denominator: 20,049 mature firms.
- Venture-measurable after fail-closed legacy quarantine: 17,477 (87.2%).
- Headline remains suppressed and `citable=false`.
- The existing M&A artifact has 3,198 direction-sensitive rows and zero typed-complete refinement records. The old refinement file covers 2,049 names but all are legacy-invalid under the new contract; another 1,149 names are absent. A fresh strict rebuild of all 3,198 rows against an approved/pinned SEC cut is required before applying a final M&A artifact.

I did **not** add post-result deterministic bounds. That is a new analytical reporting rule after results became visible, so it needs an append-only Amendment 4 approval first. The 50-firm review, face-validity anchors, negative-control threshold, and evidence-contract/promotion gates also remain human work. This PR still measures an exploratory sustained-federal-performer proxy; it does not establish physical supply-chain embeddedness or a chokepoint.

hollomancer:
Follow-up: replacement CI is fully green at 063d4b53. The PR is conflict-free (merge state CLEAN); all routed lint, type, guard, test, E2E, security, CodeQL, and version-policy checks passed. It remains draft pending the pinned SEC source-cut decision and human validation gates described above.

hollomancer:
Approved D1-A/D2-B repair is complete and frozen.

- Strict SEC/M&A chain: 34,460 unique scan rows -> 5,252 base events -> 3,198 exact typed refinements -> 5,252 final events; derivation mismatches: 0.
- Remaining typed gaps: 132 scan rows and 113 direction-sensitive event rows. These map to 196 nonmeasurable mature firms in the central denominator.
- Central `(10,6,15)` result: 20,049 mature firms, 19,853 measurable (99.02%). The point headline remains null.
- Post-result Revision 4 interval: firms 24.10%-24.33%; cumulative observed SBIR/STTR dollars 67.10%-67.42%; 47 upper-endpoint movers / $184.3M.
- Two independent runs produced byte-identical Parquet (`f2ee98c9…`), CSV (`bf4a23f3…`), Markdown (`53dc332b…`), and SVG (`c9161bc6…`) outputs; manifests differ only by output-directory paths.
- Private source-cut manifest: `0c4480de…`; mixed August 21 / September 10 retrieval window and approved SEC disclosure are recorded explicitly.
- Validation remains `blocked_missing_required_signal_inputs`; no 50-firm packet was emitted, so D3-A remains approved but cannot start without clearing or formally amending that gate.

All routed checks are green, the independent quality/spec audits pass, and the PR body now contains only the final repaired result and its non-citable limits.

PR #683, opened 2026-08-30
docs(study): add prospective SBIR M&A signal protocol

hollomancer:
## Summary
- add an exploratory, non-citable protocol for a new SBIR M&A dated-signal study
- pin the available SBIR.gov bulk object as a private retrieval-defined source-row selection, with 2026-08-29 as the Proposal Award Date boundary
- acquire and pin the private SEC Form D quarterly index source: 71 files (2009 Q1–2026 Q3), with 784,466 filed entries through 2026-08-28
- generate a private exact-key candidate ledger: 1,800 candidate names and 5,744 Form D filings; no fuzzy tiers or resolved-identity claim
- download XML for every selected filing with per-filing provenance; all selected accessions returned HTTP 200
- add an Amendment 7 private, accession-grain observation of the Form D business-combination field, without event creation, aggregation, or result output

## Verification
- ruff checks on the study utilities
- XML manifest and predicate-ledger reconciliation: one private observation row per selected accession, with SHA-256 provenance
- make docs-check
- uv run python scripts/ci/validate_study_manifests.py

## Scope
The candidate ledger, downloaded XML, and predicate observations are ignored private artifacts. Identity resolution, predicate interpretation as an event, EDGAR/EFTS search, aggregation, and public claims remain separate gates.


hollomancer:
## Evidence audit: mergeable after one factual fix

This is the strongest manifest of the four research PRs currently open, and it's worth saying why — it's a good model for opening a study without claiming anything:

- `estimand`: "**No numerical output is currently authorized.**" An estimand that declares its own emptiness is the right answer for a pre-run protocol.
- `materialization: allowed: false` with a named, specific blocker — `studies/README.md:35-36` requires that a closed gate name its unresolved blocker, and this one does.
- `identity_policy.strategy`: "No identity resolution is authorized."
- `negative_evidence_allowed: false`.
- Explicitly refuses to reuse the unrecoverable April 2026 result as either input or benchmark.

I verified both frozen hashes independently — `design.md` → `e4e690a5…b37cfe6` and the notebook → `eb797c8f…1ee51b1` both match `study.yaml:16-20`. All five `implementation` paths resolve (`sbir_ma_signal_counts_by_fy.py` is pre-existing on `main` from #583, not missing). `validate_study_manifests.py` will pass.

### One fix

**`notebooks/BACKLOG.md:53` states a status the spec contradicts.**

The new row reads: "Blocked pre-run protocol. **No sources acquired**, no April-result reproduction, no numerical output…"

But `specs/sbir-ma-dated-signal-study/amendments.md` records four *completed* private acquisitions — Amendment 1 (SBIR.gov source acquisition), Amendment 3 (fresh-snapshot attempt, object retrieved 2026-08-30), Amendment 5 (SEC Form D index acquisition), Amendment 6 (Form D candidate ledger and XML retrieval) — and `requirements.md:5-8` says those steps "are complete."

This errs toward *under*-claiming, so it's not a citability leak. But a tracked status line saying "no sources acquired" when private SBIR.gov and SEC Form D data has in fact been retrieved is a provenance-visibility defect, and BACKLOG.md is the line a reviewer scanning for state will trust.

Suggested correction: sources acquired privately under authorization; identity resolution, aggregation, and numerical output remain blocked.

---

notebooks/BACKLOG.md
```diff
@@ -50,6 +50,7 @@ canonical generator; they stay `exploratory` and non-citable until explicitly pr
...
 | Top-10 incumbent repeat-winner displacement exposure | A-CP7 — how much Phase I capacity do the ten highest *prior*-award-count firms absorb, and what is the mechanical reallocation capacity toward first-time and low-volume incumbents? | `explorations/a_cp7_top10_repeat_winner_displacement.ipynb` | Active. Exploratory crowd-out **proxies** only; no applicant file, not causal, not a Status upgrade. |
+| Prospective SBIR M&A dated signal study | F1 — what dated observed M&A signals are present in a newly reviewed SBIR firm frame as of 2026-08-29 UTC? | `explorations/sbir_ma_dated_signal_study.ipynb` | Blocked pre-run protocol. No sources acquired, no April-result reproduction, no numerical output; exploratory and non-citable pending human source/privacy/license approval and blinded validation design. |
```
hollomancer: **Status line contradicts the spec.** "No sources acquired" is false: `specs/sbir-ma-dated-signal-study/amendments.md` records completed private acquisitions (Amendments 1, 3, 5, 6), and `requirements.md` says those steps are complete.

This under-claims rather than leaking citability, but BACKLOG.md is the line a reviewer scanning for state will trust. Suggested: sources acquired privately under authorization; identity resolution, aggregation, and numerical output remain blocked.

hollomancer (commented):
Strong pre-run protocol: empty estimand, materialization closed, April result refused. One factual fix in `notebooks/BACKLOG.md` — "No sources acquired" contradicts completed amendments.

notebooks/BACKLOG.md
hollomancer: Confirmed and fixed in `55f6c816`. Amendment 7 makes it explicit — it opens *"The operator directed continuation **after the authorized candidate XML retrieval completed**"* — so the acquisitions are not merely authorized, they happened.

New status line:

> Blocked pre-run protocol. Sources acquired privately under operator authorization (amendments 1, 3, 5, 6) and a source-field predicate observed (amendment 7); identity resolution, aggregation, numerical output, and April-result reproduction all remain blocked. Exploratory and non-citable pending human source/privacy/license approval and blinded validation design.

I added amendment 7 to your suggested wording. The predicate observation is a step past acquisition and is the thing most likely to be mistaken for a finding by someone skimming, so naming it here alongside its interpretation boundary seemed better than leaving it invisible.

Your framing of why this matters is the right one: an under-claim on the line reviewers trust for state is still a false claim, and it is the direction that makes an audit trail look thinner than it is.

`make docs-check` passes.

---

hollomancer:
The BACKLOG correction is in and accurate, verified at `55f6c816`.

The row now reads "Sources acquired privately under operator authorization (amendments 1, 3, 5, 6) and a source-field predicate observed (amendment 7); identity resolution, aggregation, numerical output, and April-result reproduction all remain blocked." I checked each amendment against `amendments.md` rather than taking the numbering on trust — 1 (private first-source acquisition), 3 (fresh-snapshot attempt), 5 (SEC Form D index), 6 (Form D candidate ledger and XML retrieval), and 7 (Form D XML predicate observation) all exist and say what the row claims. The blocked-work list still matches the manifest's `materialization: allowed: false`.

Worth noting the file has since grown an Amendment 8 (private filing-level identity review queue). That doesn't contradict the row — a review queue isn't identity resolution — but if this sits open much longer, the status line will need re-checking against the amendment list again. It's the kind of line that drifts.

**I also merged `main` in and pushed (`dfbf3c41`)**, since this had gone into conflict on `specs/status.md` after v0.12.0 and #693 both added registry entries. Clean merge, no content lost — 37 entries, with this spec's and the new `edgar-event-date-fidelity` entry both present, and the bidirectional registry guard passes.

Nothing else outstanding from my review. This remains the strongest of the four research manifests I looked at.

---

hollomancer:
Fresh review of head `dfbf3c41`, then pushed `12e0374`.

**Doc drift, fixed.** `specs/status.md` said source acquisition "remain[s] blocked" while amendments 1, 3, 5, and 6 record completed private acquisitions, and BACKLOG.md, `requirements.md`, and the `tasks.md` header each named a different amendment as the current ceiling. All four now name amendment 8 (private identity review queue) as the ceiling and list what stays blocked. The 1,800-name and 5,744-filing candidate-ledger counts in `tasks.md` are now labeled as private provenance counts at name-key and accession grain, matching the label already on the SBIR source-row diagnostics.

**Retry edge case, fixed.** `fetch_sbir_ma_form_d_xml.py` treated any HTTP 200 as completed. A 200 with an empty body is written as `bytes: 0, sha256: null` and counted as failed, but was never retried on rerun. Completion now requires a non-empty body. The User-Agent version string comes from `sbir_etl.__version__` instead of a stale `0.11.0` literal.

**Verified clean:** frozen-artifact hashes recomputed and match; `validate_study_manifests.py`, tier and research-question guards, docs-check, and ruff all pass; no ledger, XML, or name file is tracked; no predicate distribution appears anywhere tracked or in the PR body.

The frozen notebook still describes the pre-Amendment-1 state ("Inputs: none"). That is true of the notebook itself and does not need a change, but any edit to it requires re-pinning its hash.

---

scripts/data/observe_sbir_ma_form_d_business_combination.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _xml_provenance(path: Path) -> dict[str, dict[str, str]]:
+    provenance: dict[str, dict[str, str]] = {}
+    for record in _records(path):
+        if record.get("status") != 200:
+            continue
+        accession = record["accession_number"]
+        if accession in provenance:
+            raise ValueError(f"Duplicate successful XML retrieval: {accession}")
```
hollomancer: **[bug]** `_xml_provenance` treats every HTTP 200 as a unique successful retrieval and raises on a second one. `fetch_sbir_ma_form_d_xml.py` was changed so a 200 with an empty body is *not* complete and must be retried, appending another status-200 line (this time with bytes and a SHA-256) to the same manifest. A resumed fetch that actually uses that retry path produces a manifest the observer cannot read, so Amendment 7's required accession-grain ledger cannot be rebuilt from the authorized provenance file.

**Suggestion:** Ignore 200s that lack `bytes`/`sha256`. If the same accession appears more than once, keep the last record that has a non-empty SHA-256 rather than failing the run. Mirror that rule in `_completed_accessions` so fetch and observe share one definition of a successful retrieval.

scripts/data/build_sbir_ma_form_d_identity_review_queue.py
```diff
@@ -0,0 +1,106 @@
...
+
+    with args.output.open("w", encoding="utf-8") as output:
+        for observation in observations:
+            if observation["predicate_status"] != "true":
+                continue
+            accession = observation["accession_number"]
+            candidate = candidates.get(accession)
+            if candidate is None:
+                raise ValueError(f"Observation without candidate: {accession}")
+            issuer_name, issuer_cik = _issuer_fields(args.xml_dir / f"{accession}.xml")
```
hollomancer: **[suggestion]** Issuer name/CIK are parsed from whatever `{accession}.xml` is on disk, while `xml_sha256` is copied from the Amendment 7 observation. Observe only evaluates the predicate after that SHA matches; the queue does not. A replaced or truncated XML file would still emit `issuer_name_alias_agreement` (or a blank issuer) under the old hash. Missing files also raise `FileNotFoundError` instead of staying in the unreviewed queue.

**Suggestion:** Hash the XML bytes, require equality with `observation["xml_sha256"]` before calling `_issuer_fields`, and on mismatch/missing file leave issuer fields null (or record a provenance conflict) without crashing.

scripts/data/build_sbir_ma_form_d_identity_review_queue.py
```diff
@@ -0,0 +1,106 @@
...
+                "form_d_filer_name": filing["filer_name"],
+                "form_d_issuer_cik": issuer_cik,
+                "form_d_issuer_name": issuer_name,
+                "form_type": filing["form_type"],
+                "index_cik": filing["cik"],
+                "name_key": candidate["name_key"],
+                "name_key_profile": candidate["name_key_profile"],
+                "prefilled_evidence_codes": evidence_codes,
+                "queue_version": QUEUE_VERSION,
+                "review_outcome": "unreviewed",
```
hollomancer, line 127: **[suggestion]** Amendment 8's human adjudication rule requires a closed outcome, written rationale, reviewer id, timestamp, *and* source-reference IDs, plus assignment from a closed evidence-code list. The queue has `review_outcome` / `review_rationale` / `reviewer_id` / `reviewed_at_utc` and `prefilled_evidence_codes`, but no slot for the reviewer's selected codes or source-reference IDs. Exact-name agreement is prefilled as an evidence code, which is easy to misread as the confirmation Amendment 8 forbids.

**Suggestion:** Add empty `evidence_codes` and `source_reference_ids` fields (keep prefills separate), and document that `issuer_name_alias_agreement` alone must remain `unresolved`.

notebooks/BACKLOG.md
```diff
@@ -50,6 +50,7 @@ canonical generator; they stay `exploratory` and non-citable until explicitly pr
...
 | Top-10 incumbent repeat-winner displacement exposure | A-CP7 — how much Phase I capacity do the ten highest *prior*-award-count firms absorb, and what is the mechanical reallocation capacity toward first-time and low-volume incumbents? | `explorations/a_cp7_top10_repeat_winner_displacement.ipynb` | Active. Exploratory crowd-out **proxies** only; no applicant file, not causal, not a Status upgrade. |
+| Prospective SBIR M&A dated signal study | F1 — what dated observed M&A signals are present in a newly reviewed SBIR firm frame as of 2026-08-29 UTC? | `explorations/sbir_ma_dated_signal_study.ipynb` | Blocked pre-run protocol. Sources acquired privately under operator authorization (amendments 1, 3, 5, 6) a source-field predicate observed (amendment 7), and a private filing-level identity review queue prepared (amendment 8); automated identity decisions, aggregation, numerical output, and April-result reproduction all remain blocked. Exploratory and non-citable pending human source/privacy/license approval and blinded validation design. |
```
hollomancer: **[suggestion]** The question cell still asks what signals are present "in a newly reviewed SBIR firm frame as of 2026-08-29 UTC." Amendment 4 and `study.yaml` reject that reading: the date is an inclusive `Proposal Award Date` selection on one pinned object, not a complete as-of firm universe.

**Suggestion:** Replace "as of 2026-08-29 UTC" with the retrieval-defined selection wording used in the manifest (source-row cutoff in the pinned provider object, not an as-of frame).

hollomancer: **[nit]** The status cell is missing a conjunction after the first parenthetical: "authorization (amendments 1, 3, 5, 6) a source-field predicate observed".

**Suggestion:** Restore "and" (or a semicolon) so the three completed private steps read as a list.

hollomancer:

### Status: 🔴 Draft — Needs Review

- **Draft:** Yes
- **Merge status:** CLEAN, no conflicts
- **Age:** 10 days
- **Size:** +1,464/-0 across 14 files

Adds an exploratory protocol for a new SBIR M&A dated-signal study with pinned sources, candidate ledger, and provenance tracking. All CI checks pass. Well-structured body with clear scope boundaries and verification steps.

As a draft, this needs a maintainer review pass before it can land. The scope, evidence boundaries, and artifact handling are clearly documented.

**Recommended action:** Request review when ready to bring out of draft.

hollomancer:

Reviewed at `12e03748`.

- **Blocker — the hash-frozen notebook contradicts the amended protocol.** `notebooks/explorations/sbir_ma_dated_signal_study.ipynb` still says acquisition is unauthorized, inputs are absent, or no acquisition occurred, while Amendments 1-8 and the spec record bounded acquisition and observation. Because `studies/sbir-ma-dated-signal-study/study.yaml:16-20` pins the notebook as the current contract artifact, update it and refresh the hash, or explicitly classify it as an immutable historical initial protocol.
- **Blocker — current XML bytes can inherit a stale recorded hash.** `scripts/data/build_sbir_ma_form_d_identity_review_queue.py:38-100` parses the current XML file but copies `observation["xml_sha256"]` without recomputing and comparing it; missing XML raises instead of producing an unavailable record. Read once, recompute SHA-256 before parsing, require equality, and fail closed for missing/mismatched files. Add valid, missing, and tampered-XML tests.

The prospective study remains the right validation path, but human adjudication, privacy/license review, freeze, and evidence promotion should remain explicit draft gates.

scripts/data/build_sbir_ma_form_d_identity_review_queue.py
hollomancer: `51b5eb11` — the queue hashes the on-disk XML and parses issuer fields only when that SHA-256 equals the Amendment 7 observation hash. Missing or mismatched files leave issuer fields null, omit `issuer_name_alias_agreement`, and stay in the unreviewed queue.

notebooks/BACKLOG.md
hollomancer: `51b5eb11` — the question cell now uses the inclusive Proposal Award Date selection through 2026-08-29 UTC in one pinned provider object, not an as-of firm universe.

scripts/data/observe_sbir_ma_form_d_business_combination.py
hollomancer: `51b5eb11` — `_xml_provenance` now ignores HTTP 200s that lack bytes/SHA-256 and keeps the last complete retrieval for an accession instead of raising. `_completed_accessions` uses the same complete-retrieval rule, so a resumed empty-body 200 retry can rebuild the Amendment 7 ledger.

scripts/data/build_sbir_ma_form_d_identity_review_queue.py
hollomancer: `51b5eb11` — queue rows now include empty `evidence_codes` and `source_reference_ids` for the reviewer, separate from `prefilled_evidence_codes`. The script documents that `issuer_name_alias_agreement` alone must remain `unresolved`.

notebooks/BACKLOG.md
hollomancer: `51b5eb11` — restored the missing `and` after the first parenthetical in the status cell.

hollomancer:
## Follow-up on the 2026-09-10 blockers, at HEAD `b1fa0688`

Inline threads are all resolved. CI green. Mergeable.

**XML hash:** done in `51b5eb11`. The queue hashes on-disk XML, requires equality with the Amendment 7 observation SHA, and fails closed to null issuer fields on missing or mismatched bytes. Tests cover valid / missing / tampered.

**BACKLOG vs Amendment 8:** the status line now names amendments 1, 3, 5, 6, 7, and 8. That matches `requirements.md` / `tasks.md` / `amendments.md`.

**Notebook hash:** `study.yaml:19-20` still matches the file (`eb797c8f…1ee51b1`). The remaining issue is classification, not a broken pin.

The frozen notebook still says Inputs: none / acquisition unauthorized / planned as-of cutoff. Amendments 1–8 record bounded private acquisition. Need a decision on the page:

- **(a)** Update the notebook to the amended protocol and re-pin the hash, or
- **(b)** Label it in `study.yaml` as an immutable historical initial protocol and stop treating it as the current contract.

Until that choice is explicit, keep this draft. Empty estimand, `materialization: allowed: false`, and refusal to reuse the April result remain the right pre-run posture.

scripts/data/observe_sbir_ma_form_d_business_combination.py
```diff
@@ -0,0 +1,113 @@
...
+    """Return the tri-state source-field observation authorized by Amendment 7."""
+    try:
+        root = ElementTree.fromstring(xml_bytes)
+    except ElementTree.ParseError:
+        return "unavailable"
+
+    element = root.find(PREDICATE_PATH)
+    if element is None or element.text is None:
+        return "unavailable"
+    return "true" if element.text.strip().lower() == "true" else "false"
```
hollomancer: **Blocking.** `_predicate` returns `true` only for text equal to `true` after strip/lower, and `false` for any other present value. `isBusinessCombinationTransaction` is a boolean field; XML Schema boolean lexical space admits `true`, `false`, `1`, and `0`, so a filing encoding `1` is recorded as false — a positive observation flipped to a negative one in a study whose manifest sets `negative_evidence_allowed: false`. Amendment 7 mandates exactly this rule, so the defect is in the authorization text as much as the code.

Second failure mode: `main()` returns 0 and prints nothing (commit `2065dfa9` suppressed aggregate output), so if `PREDICATE_PATH` never resolves — a default XML namespace on real `primary_doc.xml`, a schema difference, a fetched HTML error page — every row is `unavailable` and the run is indistinguishable from a clean one. The only XML fixture is a three-element synthetic string with no XML declaration, no `schemaVersion`, and no namespace, and `_predicate` has no direct test: no false case, no missing-element case, no parse-error case, no SHA-mismatch case.

This field is the substance of Amendments 7 and 8; the identity review queue admits only `predicate_status == 'true'` rows, so a systematic parse failure silently produces an empty queue and a systematic `1`-encoding produces false negatives.

**Fix:** Map `{true, 1} → true`, `{false, 0} → false`, anything else → `unavailable`, and amend Amendment 7 to match. Print counts that carry no predicate information: rows total, rows with no provenance line, rows with a missing XML file, rows with a SHA mismatch, rows where the predicate element did not resolve. Add unit tests for false / missing element / malformed XML / SHA mismatch, and pin one redacted real `primary_doc.xml` as a fixture.

scripts/data/build_sbir_ma_form_d_candidates.py
```diff
@@ -0,0 +1,131 @@
...
+                "sbir_award_identifiers": [
+                    {"uei": uei, "duns": duns, "agency_tracking_number": tracking, "contract": contract}
+                    for uei, duns, tracking, contract in identifiers
+                ],
+                "form_d_index": filing,
+            }
+            output.write(json.dumps(record, sort_keys=True) + "\n")
+            written += 1
+    print(f"SBIR exact-name keys: {len(sbir):,}")
+    print(f"Candidate Form D filings: {written:,}")
```
hollomancer: **Blocking.** This prints only two counts: `SBIR exact-name keys: {len(sbir)}` (distinct name keys across all 112,949 selected SBIR rows, not the keys that matched a Form D filer) and `Candidate Form D filings: {written}` (ledger rows at `(name_key, accession)` grain). `tasks.md:76-77` reports “1,800 exact-key candidate names and 5,744 candidate filings” as if both came from here. 1,800 cannot be `len(sbir)` — that would imply about 63 award rows per distinct company name across 112,949 rows — so the figure came from unrecorded post-processing while this label is close enough to be read as its source.

A headline count in a tracked spec with a mislabeled near-homonym in stdout is how a wrong number gets quoted later. The matched-key count, the one quantity that characterises the join, is never emitted.

**Fix:** Print three labelled counts — selected-row name keys, name keys that produced at least one candidate, and distinct candidate accessions — and quote those labels verbatim in `tasks.md:76-78`.

studies/sbir-ma-dated-signal-study/study.yaml
```diff
@@ -0,0 +1,69 @@
...
+  2026-08-29T23:59:59Z. This is not a complete as-of firm universe or a
+  denominator for prevalence or exit rates. A future amendment must declare any
+  observed-signal quantity, reviewed identity policy, outcome-source coverage,
+  and failure modes before acquisition or aggregation. It must not reuse or
+  reproduce the unrecoverable April 2026 result.
+frozen_artifacts:
+  - path: specs/sbir-ma-dated-signal-study/design.md
+    sha256: e4e690a55eaa21847686ecf0fb709cc9e2e3db80e8ce3380222d8aadbb37cfe6
+  - path: notebooks/explorations/sbir_ma_dated_signal_study.ipynb
+    sha256: eb797c8f25644e2cd62864b4aa84f52f4e9967182337cca45c1a748b61ee51b1
```
hollomancer: **Blocking — owner’s still-open consolidated blocker.** I recomputed this digest as `eb797c8f25644e2cd62864b4aa84f52f4e9967182337cca45c1a748b61ee51b1`, identical to the pin, so the notebook is byte-identical to the version flagged on 2026-09-10. Cells still say “**Planned as-of cutoff:** 2026-08-29 UTC”, “**Inputs:** none. Source acquisition is conditional on separate human source/privacy/license approval.”, `PLANNED_AS_OF_CUTOFF` with `assert not REQUIRED_INPUTS`, and “No analysis is authorized.” Amendments 1/3/5/6 record completed private acquisitions, 7 a completed observation, and 8 a prepared queue.

Two of the PR’s own completion criteria are therefore false: `design.md:106-108` makes notebook/spec agreement the completion test, and `tasks.md:11-12` checks off “requirements, design, notebook, and study manifest agree.” `tasks.md:11` still says “planned as-of cutoff.” The 12:04 follow-up already asked for (a) update-and-re-pin or (b) classify as immutable historical protocol; this is still that decision, now also falsifying the checked-off T0 criterion.

**Fix:** Either update the notebook to the current state and re-pin, or relabel it in `study.yaml` and `design.md` as an immutable initial-protocol record, drop it from the current-contract framing, and reword `design.md:106-108` and `tasks.md:11-12` so the agreement criterion is satisfiable. Fix `tasks.md:11` either way.

scripts/data/observe_sbir_ma_form_d_business_combination.py
```diff
@@ -0,0 +1,113 @@
...
+        return [json.loads(line) for line in handle if line.strip()]
+
+
+def _candidate_filings(path: Path) -> dict[str, dict[str, str]]:
+    filings: dict[str, dict[str, str]] = {}
+    for record in _records(path):
+        filing = record["form_d_index"]
+        accession = filing["accession_number"]
+        if accession in filings:
+            raise ValueError(f"Duplicate candidate accession: {accession}")
```
hollomancer: **Blocking.** The builder dedupes on `(key, accession)`, so one accession can be written twice under two filer-name spellings that normalize to different keys, while both consumers raise `ValueError('Duplicate candidate accession')` on a repeated accession. `tasks.md:76-78`’s reconciliation — 5,744 candidate filings and “5,744 distinct candidate accessions” — therefore holds only because this cut happened to contain no such case; a rebuild over a wider selection aborts the pipeline.

The risk is not hypothetical: 784,466 entries over 2009-01-02..2026-08-28 is about 44,400 per year, well above commonly cited Form D offering volumes. The excess is consistent with D/A amendments plus EDGAR emitting one index line per filer on multi-filer submissions — which is exactly the duplicate case, and would also mean “candidate filings” at `(name_key, accession)` grain is not a filing count.

The declared grain and the enforced grain disagree, so the only thing between the documented reconciliation and a crash is the data.

**Fix:** State the ledger grain in the builder docstring. Either dedupe on accession and carry all matched name keys in a list field, or make observe/queue collapse duplicates deterministically and record the collapse. Print distinct accessions alongside row count so the two are never conflated again.

specs/sbir-ma-dated-signal-study/tasks.md
```diff
@@ -0,0 +1,121 @@
...
+  `Proposal Award Date` boundary.
+  - Verify: 112,949 rows have nonblank `Company` and a valid date on/before the
+    cutoff; `UEI`, `Duns`, `Agency Tracking Number`, and `Contract` are present
+    in 104,461, 109,199, 112,937, and 112,948 selected rows respectively. These
+    are source-row diagnostics, not firm or outcome counts.
+- [x] Acquire and pin the authorized private SEC Form D quarterly index source.
+  - Verify: every 2009 Q1–2026 Q3 raw index file has private provenance and
+    integrity metadata; 784,466 Form D/D-A entries span 2009-01-02 through
+    2026-08-28, with none after the cutoff. No matching, XML retrieval,
+    predicate evaluation, or result is emitted.
```
hollomancer: **Blocking.** These figures have no committed producer: 112,951 parseable Proposal Award Date values; 112,949 selected rows; UEI/Duns/Agency Tracking Number/Contract present in 104,461/109,199/112,937/112,948 of them; 784,466 Form D/D-A index entries across 71 files. No script in this PR produces any of them — there is no SBIR.gov retrieval script, no Proposal-Award-Date cutoff/schema audit script, and no Form D index acquisition script, and `study.yaml`’s implementation list registers no producer either.

Separately, no tracked file states the CSV’s total row count or the excluded count, although Amendment 1 requires the row count and Amendment 4 explicitly authorizes aggregate selected/excluded counts. `build_sbir_ma_form_d_candidates.py:34-36` drops any row whose date fails `date.fromisoformat` with a bare `except ValueError: continue` and no counter.

These are the figures a reviewer scanning for state will trust. As committed they cannot be recomputed or falsified by anyone, including the author on a fresh checkout.

**Fix:** Commit the audit script (it emits aggregates only, so no private data enters Git) and add it to `study.yaml`’s implementation. Record total / selected / excluded rows broken out by reason (unparseable date, blank Company, post-cutoff, empty name key). If committing the producer is unwanted, relabel the figures explicitly as one-off unreproducible diagnostics.

studies/sbir-ma-dated-signal-study/study.yaml
```diff
@@ -0,0 +1,69 @@
...
+    retrieval-defined source-row selection. It has no complete as-of coverage,
+    reviewed firm identity, observed event population, denominator, or numerical
+    result.
+  - >-
+    Exploratory status is non-citable and does not authorize a Computable,
+    Validated, or Citable inventory status or any external numerical claim.
+  - >-
+    A signal is not an acquisition, exit, rate, survival outcome, agency effect,
+    comparator result, or causal finding without a separately frozen and
+    validated evidence-tier design.
```
hollomancer, line 69: **Blocking.** The manifest’s estimand and limitations describe only the SBIR.gov object. Missing:

(a) Both retrievals report an August 1 snapshot while the inclusive selection boundary is August 29 (`tasks.md:47-48`, `design.md:13-15`), so the final 28 days of the declared window are systematically incomplete. This is why the cutoff check is recorded as failed, and that failure is not carried into the manifest.
(b) Amendment 5’s `date_filed <= 2026-08-29` bound and the 2009 Q1 coverage floor (“not evidence of no earlier filing”), both applied in code via the same `SELECTION_CUTOFF` constant.
(c) The differential observation window — award dates run from 1905/the 1980s to 2026 while Form D observation starts in 2009 and stops 2026-08-29, so any future per-firm signal share is a function of award vintage.
(d) The Amendment 4 input hash `efdf7ca5...a23628` (`amendments.md:104`), which appears nowhere in the manifest.

`study.yaml` is the artifact that travels and the place a later reader or promotion check looks for what the number does not cover. “Not a denominator for prevalence or exit rates” is the right disclaimer but is not the same as naming the two specific censoring mechanisms the audit has already discovered.

**Fix:** Add limitations for the August 1 / August 29 gap, the `date_filed <= 2026-08-29` bound, the 2009 coverage floor, and the vintage-dependent observation window; add the Amendment 4 SHA to the manifest.

hollomancer (commented):
## Open PR review — keep draft

Reviewed against `origin/main` at `493340ea`. Head `b1fa0688`. Tier and scope discipline in the manifest is real: both frozen SHA-256 pins recompute exactly, `docs/research-questions.md` is untouched so F1 keeps its non-citable Status, and the diff is surgical. Four of the owner’s five prior line-level asks are fixed. The XML-hash blocker from 2026-09-10 is done.

What is not fixed is the owner’s own consolidated blocker (hash-pinned notebook still describes a pre-Amendment-1 world) plus five new defects underneath the paperwork. Complements the 12:04 notebook-classification comment; this pass adds the unreproducible counts, the 1,800 figure, the predicate, the grain disagreement, and the missing coverage limits.

Cannot formally request changes on an own PR; treat this as **needs work — keep draft**. Mergeable is currently dirty on `notebooks/BACKLOG.md`.

### Blocking

1. **Hash-pinned notebook contradicts the amended protocol.** Digest identical to the 2026-09-10 pin. See line comment on `study.yaml:19-20`.
2. **Every substantive number in `tasks.md` has no committed producer and no denominator.** See line comment on `tasks.md:45-70`.
3. **Headline “1,800 exact-key candidate names” is not printed by any code path.** See line comment on the builder’s stdout.
4. **Predicate treats XML Schema `1` as false**, and a total parse failure is silent. See line comment on `_predicate`.
5. **Ledger grain disagrees with what both consumers enforce.** Builder writes `(name_key, accession)`; observe/queue require unique accession. See line comment.
6. **`study.yaml` limitations omit the censoring the audit already found** (August 1 vs August 29 gap, 2009 Form D floor, vintage-dependent window, Amendment 4 SHA). See line comment.

### Number audit

8 of 13 checked figures hold as stated. 1,800 is inconsistent (no producer). Total CSV row count and excluded-by-reason counts are omitted. 1905-07-01 sits inside the accepted selection with no lower bound.

### Do these first

1. Resolve the frozen-notebook contradiction (update and re-pin, or relabel as immutable historical protocol and reword the agreement criteria) — the owner’s only still-open blocker, and it falsifies two of this PR’s own completion checks.
2. Commit the SBIR/Form D audit producer (aggregates only); record total and excluded row counts by reason; print matched name keys and distinct accessions so 1,800 and 5,744 have labelled sources.
3. Harden the predicate (`true`/`1` / `false`/`0` / else unavailable); print provenance/parse-health counts; add false / unavailable / SHA-mismatch tests plus one redacted real `primary_doc.xml`.
4. Add the missing coverage limits to `study.yaml`; declare the ledger grain; reconcile builder vs consumer uniqueness.
5. Resolve the `notebooks/BACKLOG.md` conflict by re-applying only line 63 onto current main and confirming the “Keep automated” and “Promotion queue” sections survive.

hollomancer:
## Review findings at `c358a468`

The focused implementation is mechanically healthy—all current checks are green and the 22 new targeted tests pass—but I think the study contract still needs changes before this leaves draft.

1. **P1 — Form D Item 10 is framed more strongly than the source supports.** `freeze-packet.md:11-14,33` calls the retained field an ownership-change signal, while `amendments.md:195-198` correctly says it does not establish a target, acquisition, exit, or event. [Form D Item 10](https://www.sec.gov/info/edgar/forms/formd.pdf) records only whether an offering is connected to a business combination; it does not encode target/acquirer direction. Please use a neutral, direction-unresolved transaction-evidence contract and require independent target-side evidence before this can inform F1 exit counts. This is also the safe point of alignment with #705: Form-D-only is not exit evidence, but it is not universally acquirer evidence either.

2. **P1 — Private-source aggregates were published before aggregate-release approval, and several lack committed producers.** `tasks.md:45-85` records exact SBIR, Form D, and candidate figures. However, `freeze-packet.md:62-66` still marks public aggregate release pending, while `amendments.md:141-165` prohibits counts, aggregation, and public artifacts for Amendments 5-6 (and the later amendments retain that boundary). Calling tracked values “private provenance counts” does not keep them private in a public repository. Either record explicit approval for each released aggregate or remove the figures from tracked content and the PR description, noting the prior exposure. If retained, add aggregate-only producers for the source/cutoff/index audits so another analyst with the private inputs can recompute them.

3. **P1 — The hash-pinned notebook describes a superseded study state.** `notebooks/explorations/sbir_ma_dated_signal_study.ipynb:16-31,65-69,93-120` says there are no inputs, acquisition remains unauthorized, and no source was acquired or analyzed. `study.yaml:16-20` nevertheless pins it as a current frozen artifact after Amendments 1-8 record completed retrieval, observation, and review-queue preparation. Update and re-pin the notebook, or explicitly label it as an immutable initial-protocol snapshot and remove it from current-state agreement checks.

4. **P1 — The same valid XML value has incompatible interpretations and the semantic change was not versioned.** `observe_sbir_ma_form_d_business_combination.py:78-102` maps XML Schema `1` to `true`, while shared `form_d_scoring._bool` maps only literal `true`, so existing consumers can classify the same filing as `False`. In addition, `PREDICATE_VERSION` remains `form_d_business_combination_v1` even though `amendments.md:183-190` records a semantic correction. Move this field into one shared tri-state parser, add parity tests, bump the predicate version, and rerun/reconcile the private observation and review queue.

5. **P2 — Retrieval resume trusts the manifest without validating the saved XML.** `fetch_sbir_ma_form_d_xml.py:43-64,90-95` skips an accession based solely on a successful manifest line. A deleted/corrupt file, a manifest reused with a different output directory, or a non-XML HTTP-200 body is therefore treated as complete and cannot self-heal. Before skipping, verify the on-disk size/hash and expected XML structure; add missing/corrupt-artifact regression tests.

Recommendation: keep this draft until the authorization, current-state, source-semantics, and provenance issues are resolved. It can then land independently; #705 is not a hard commit dependency.

hollomancer:
## What is still needed before this leaves draft

CI is green and the PR is mergeable. Keep it draft. The empty estimand, closed materialization, and refusal of the April result are still the right posture. What is left is to make the *current* contract true.

**Decide, then edit:**

1. **Notebook pin.** `study.yaml` still hashes a notebook that says “Inputs: none / acquisition unauthorized.” Amendments 1–8 already acquired sources. Either update and re-pin, or label it an immutable initial protocol and drop it from current-state agreement checks. Standing blocker since 10 Sep.

2. **Item 10 is not an ownership-change or acquirer signal.** `freeze-packet.md` overclaims. The field is a direction-unresolved business-combination flag. Independent target-side evidence is required before it can inform F1 exits. Same fact as #705; this PR should keep the weaker, safer wording.

3. **Private counts in a public repo.** `tasks.md` publishes 1,800 names / 5,744 filings / SBIR row audits while `freeze-packet.md` still says aggregate release is pending. Remove the figures, or record approval *and* commit aggregate-only producers so someone else can recompute them. The 1,800 figure currently has no committed printer.

4. **One parser, versioned.** `observe_…` maps XML Schema `1` → true; `form_d_scoring._bool` maps it to `False`. Bump `PREDICATE_VERSION`, put the tri-state in one shared function, then rerun the private observation/queue.

5. **Resume must hash the file.** `fetch_sbir_ma_form_d_xml.py` skips on a successful manifest line. Check on-disk size/hash/XML before skipping.

Human adjudication and license review stay later gates for *running* the study, not for merging the protocol. #705 is not a merge dependency. The XML boolean `1`/`true` split is the only hard coupling with #705; fix it once as a shared primitive, then this can land as a pre-run protocol.

studies/sbir-ma-dated-signal-study/study.yaml
hollomancer: Fixed in `c358a468`. All four are in `study.yaml` limitations now:

- the August 1 snapshot against the August 29 inclusive boundary, naming the 28 days as systematically incomplete and the cutoff check as failed for that reason;
- the `date_filed <= 2026-08-29` bound and the 2009 Q1 coverage floor, with absence of an earlier filing explicitly not evidence that none exists;
- the differential observation window, award dates from the 1980s against Form D observation starting 2009, so a per-firm signal share is a function of award vintage;
- the Amendment 4 input SHA `efdf7ca5...a23628`, with a note that the bytes are not committed so nothing in this repository can reproduce the selection from them.

notebooks/explorations/sbir_ma_dated_signal_study.ipynb
```diff
@@ -0,0 +1,192 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "findings",
+   "metadata": {},
+   "source": [
+    "## Findings and caveats\n",
+    "\n",
+    "| Claim | Evidence/artifact | Caveat or alternative explanation |\n",
+    "|---|---|---|\n",
+    "| No finding | No inputs acquired or analyzed | Exploratory, blocked, and non-citable |\n",
```
hollomancer, line 153: **[suggestion]** `eabebd14` re-pinned this notebook as the current contract and updated the research-contract cells to say sources were acquired under amendments 1, 3, and 5–8. The findings table still states "No inputs acquired or analyzed", and the analysis placeholder at line 140 still says "no source acquisition or analysis is authorized." Those sentences contradict the rest of the notebook, `design.md`'s "updated to describe the acquired state" claim, and T0's notebook/spec agreement check. A hash-pinned findings table is what a later reader will quote.

**Suggestion:** Reword the findings row to "No finding; this notebook loads none of the private sources and performs no analysis" and drop "source acquisition" from the placeholder comment so only analysis stays unauthorized. Re-pin the notebook hash in `study.yaml`.

specs/sbir-ma-dated-signal-study/freeze-packet.md
```diff
@@ -0,0 +1,93 @@
...
+> It neither supersedes Amendment 1 nor authorizes a source acquisition, a
+> date-field scan, identity resolution, candidate creation, aggregation,
+> materialization, or a claim. It exists so the owner can approve or reject a
+> bounded contract without reconstructing the lost April analysis.
+
+## Proposed narrow question and estimand
+
+**Question.** In a reviewed SBIR firm frame observed through the planned cutoff
+of `2026-08-29T23:59:59Z`, what is the count of reviewed candidate firms with
+at least one retained, source-observed ownership-change signal from the approved
```
hollomancer, line 13: **[suggestion]** The proposed question still counts "source-observed ownership-change signal[s]", and line 33 names Form D as the "First ownership-change signal" for "records whose source record identifies a business-combination offering." Amendment 7 (amendments.md:195–198) correctly says a `true` Item 10 value is not a target, acquisition, transaction, exit, or event, and Form D Item 10 does not encode direction. If this draft is later frozen as written, F1 exit counts would rest on a direction-unresolved offering flag.

**Suggestion:** Use a direction-unresolved business-combination observation here, matching Amendment 7. Require independent target-side evidence before this field can inform an F1 exit count. Line 32's 112,949 selected-row figure should carry the same "one-off diagnostic, no committed producer" caveat now on `tasks.md`, or move to that file only.

scripts/data/observe_sbir_ma_form_d_business_combination.py
```diff
@@ -0,0 +1,170 @@
...
+
+import argparse
+import hashlib
+import json
+from pathlib import Path
+from xml.etree import ElementTree
+
+
+EPISTEMIC_TIER = "exploratory"
+PREDICATE_VERSION = "form_d_business_combination_v1"
```
hollomancer, line 19: **[suggestion]** Amendment 7 Revision 1 changed the mapping of XML Schema `1`/`0` while leaving `PREDICATE_VERSION = "form_d_business_combination_v1"`. A private ledger produced before the revision is indistinguishable from one produced after, and `sbir_etl.enrichers.sec_edgar.form_d_scoring._bool` still treats only literal `"true"` as true, so `1` is `False` in the shared parser and `"true"` here. The identity queue admits only `predicate_status == "true"`, so a pre-revision run that stored `1` as `false` would drop those filings from review with no version marker.

**Suggestion:** Bump to `form_d_business_combination_v2`, share one tri-state helper with the Form D scorer (or document that this study's parser is intentionally distinct), and rerun the private observation and queue so the version string matches the bytes.

scripts/data/fetch_sbir_ma_form_d_xml.py
```diff
@@ -0,0 +1,166 @@
...
+        help=f"Request rate (1-{MAX_REQUESTS_PER_SECOND}; default: %(default)s).",
+    )
+    args = parser.parse_args()
+    if args.limit is not None and args.limit < 1:
+        parser.error("--limit must be positive")
+    if not 1 <= args.requests_per_second <= MAX_REQUESTS_PER_SECOND:
+        parser.error(f"--requests-per-second must be between 1 and {MAX_REQUESTS_PER_SECOND}")
+
+    candidates = _load_candidates(args.candidates)
+    completed = _completed_accessions(args.manifest)
```
hollomancer, line 91: **[suggestion]** Resume skips an accession from the manifest alone (`_completed_accessions` at line 62: HTTP 200 with nonzero `bytes` and a SHA-256). It does not check that `{accession}.xml` still exists, matches that digest, or is XML. A deleted or truncated file, a manifest reused with another `--output-dir`, or an HTTP 200 HTML interstitial (SEC undeclared-tool pages) is treated as complete and will not self-heal. Observe fail-closes those rows to `unavailable`, but fetch will never retry them.

**Suggestion:** Before skipping, verify the on-disk size and SHA-256 against the last complete manifest line and reject non-XML bodies (content-type or root tag). Keep empty-body 200s retryable, as they are now.

scripts/data/build_sbir_ma_form_d_candidates.py
```diff
@@ -0,0 +1,146 @@
...
+                    (row.get("Agency Tracking Number") or "").strip(),
+                    (row.get("Contract") or "").strip(),
+                )
+            )
+            entry["source_row_count"] += 1  # type: ignore[index,operator]
+    return candidates
+
+
+def _form_d_entries(index_dir: Path):
+    for path in sorted(index_dir.glob("*.idx")):
```
hollomancer, line 66: **[suggestion]** `_form_d_entries` uses `index_dir.glob("*.idx")`, which does not recurse. SEC quarterly indexes live at `YYYY/QTRn/form.idx`. Passing that tree yields zero files, `written == 0`, and exit code 0 — a silent empty ledger. This run clearly used a layout the glob could see, but the expected flattening is not documented and there is no empty-input check.

**Suggestion:** Use `rglob("*.idx")` or fail if no index files are found, and document the expected directory shape next to `--form-d-index-dir`.

scripts/data/build_sbir_ma_form_d_candidates.py
```diff
@@ -0,0 +1,146 @@
...
+                    {"uei": uei, "duns": duns, "agency_tracking_number": tracking, "contract": contract}
+                    for uei, duns, tracking, contract in identifiers
+                ],
+                "form_d_index": filing,
+            }
+            output.write(json.dumps(record, sort_keys=True) + "\n")
+            written += 1
+            matched_keys.add(key)
+            matched_accessions.add(str(filing["accession_number"]))
+    # Three distinct quantities that were previously collapsed into two labels.
```
hollomancer, line 134: **[nit]** "Three distinct quantities that were previously collapsed into two labels" narrates the review history. The same pattern is in `observe_sbir_ma_form_d_business_combination.py:33–34` and `build_sbir_ma_form_d_identity_review_queue.py:40–41` ("Rejecting the repeat would abort the pipeline").

**Suggestion:** Keep the grain fact ("row count is not a filing count") and drop the prior-collapse / abort-the-pipeline framing.

PR #685, opened 2026-08-31
Add exploratory Navy Phase III transition readout

hollomancer:
## Summary

- add a compact exploratory/non-citable Navy Phase III transition readout, aggregate summary, and two figures
- extend the research-only FPDS Element 10Q puller with exact repeatable query clauses plus transaction/revision and funding-agency fields
- reuse canonical award-key, Phase II/III pairing, and name-normalization helpers; emit no firm names or row-level matches
- add a quality/literature crosswalk and harden shared evaluation semantics exposed by the audit

## Breaking value change in the checks schema

`checks["undercount_warning"]` keeps its key but **its `note` value changed**. It previously read "FPDS sbir_phase coding is known to undercount Phase III, especially outside DoD. Treat transition rates as lower bounds." It now reports observed coded-channel rates rather than identified bounds, because bias runs in both directions: missed codes and identity misses bias downward, while same-firm false matches bias upward.

The old text was not preserved under the deprecated key on purpose — it asserts the one-sided bias this readout retracts, so carrying it forward would keep serving a withdrawn claim to consumers that have not migrated. The alias instead carries a `note_semantics_changed` field naming the change and pointing at `coding_coverage_warning`. Nothing inside this repository reads `undercount_warning`; it exists for external readers of the checks JSON.

## Corrected results and limits

- FY2016–FY2025: 4,828 Navy Phase I awards, 2,780 Phase II awards, and 2,282 Navy-attributed strict-coded Phase III keys
- 1,884 coded keys first appear in-window; 398 originated earlier, so the annual table now separates first from latest coded action
- latest-action description median: 34 characters; 39.5% reach 40 and 7.8% reach 150
- the former 900-character diagnostic is removed: FPDS requires Description of Requirement and has capped newly entered text at 250 characters since 2019-06-28; eight later representatives above 250 trace to legacy pre-cap contracts
- 799/2,012 completed Phase II awards map to 315 distinct same-UEI coded contracts across 157 firms; 159 contracts cover 643 assignments, with maximum reuse of 36
- signed latency is event-conditional, not a survival estimate: 525 assignments precede Phase II completion and 274 are on/after completion; 1,213 awards have no qualifying coded signal by the administrative cut
- Panel C remains blocked because required-field description missingness is constant; Panel D reports 47 qualifying Form D firms, while EFTS acquisition timing and the union remain blocked
- the historical DoD comparator remains unreproduced and non-method-matched; no statutory undercount, ranking, recommendation, or affiliation claim is made

## Infrastructure finding

The 900-character error was broader than report prose but bounded to research-evaluation semantics. Shared benchmark code accepted 515/900-character cut points despite the FPDS cap; core FPDS ingestion did not truncate text or impose a 900-character rule. This revision:

- centralizes the official required/250-character FPDS constraint and rejects out-of-range benchmark thresholds
- changes character counts from “floors” to diagnostics
- removes blanket lower-bound language from reusable Phase III metadata
- marks signed completion-relative frames as not Kaplan–Meier-ready without a valid nonnegative origin and censoring assessment

## Published context

The report now links GAO and NASEM findings on incomplete Phase III tracking; the closest Navy study (Rovito, Kamp, and Etemadi, 2025); NRC evidence that sales can precede Phase II completion; Kaplan and Meier (1958); NRC commercialization surveys; Howell (2017); and SEC Form D methodology. These sources support the limitations and outcome choices, not the Navy point estimates.

## Known gap outside this diff

`docs/guides/government-policy-demo-plan.md:225` still instructs presenters to "describe [Phase III totals] as lower bounds when derived from public coding or keyword discovery." That is the same claim this PR retracts, in the document that scripts a live policy room. Left untouched here to keep the diff bounded; it needs its own change.

## Verification

- 84 focused readout, FPDS, benchmark, phase-transition, and asset-job tests passed
- full Ruff lint/format and mypy passed (`330` typed source files)
- `make lint-boundaries`, `make docs-check`, and `git diff --check` passed
- broader unit collection reached 6,525 passed / 7 skipped; two unrelated legacy tests were blocked only by the managed sandbox&#39;s refusal to write through offloaded data-directory symlinks
- repeated pinned-input builds reproduced the summary and figures


hollomancer:
## Evidence audit: mergeable after two wording fixes

Declared exploratory and genuinely built to it, with unusual rigor. `docs/readouts/navy-transition-v1.md:3-5` carries the "**Exploratory — non-citable.**" banner, and `navy-transition-v1.summary.json` carries machine-readable `_epistemic: {citable: false, tier: "exploratory", …}`, which satisfies the `citable: false` metadata requirement in `epistemic-tiers.md:131-133`. The new `sbir_etl/utils/fpds_constraints.py` inherits `EPISTEMIC_TIER = "pipelines"` and contains only source-system constants — right for its tier.

Worth calling out specifically: Panel C is **blocked rather than substituted** (`panel_c.status: "blocked_constant_input"`), Panel D's EFTS branch is blocked for lack of acquisition dates, and the historical DoD comparator is labeled "historical, unreproduced" with a disagreement between legacy notes surfaced rather than quietly resolved. Failed and negative results stay visible. That is the behavior the contract wants and it is the part of this PR most worth keeping.

**The pipelines-tier changes are a net epistemic improvement.** `phase_iii.py:1-6`, `pairs.py`, `models/phase_transition.py:136-146`, and `docs/phase-transition-latency.md:34-44,67-77` systematically replace "known undercount" / "lower bound" / "Ready for Kaplan-Meier" with "observed coded-channel rate, **not an identified bound**: coding/identity misses can bias downward while unrelated same-firm matches can bias upward," and add `km_ready: False` plus an `estimator_warning` to asset metadata. That corrects a real over-claim in shipped pipeline code. I confirmed no study manifest pins any of the touched files, so no frozen hash breaks.

### Two fixes before merge

**1. `docs/readouts/navy-transition-v1.md:107` — reserved-word misuse, contradicted three lines above.**

> "The results are a **reproducible** local map of observed public signals…"

`reproducible` is a reserved study-status rank (`studies/README.md:10`, mapped to `Computable` at `:18-22`). This readout has no study contract. And it contradicts the framing at `:10-11` of the same document: "untracked snapshots prevent a clean-checkout rebuild." Suggest "a **local** map of observed public signals" or "a locally regenerable map."

**2. `phase_iii.py:188-191` — the deprecated alias is not value-compatible.**

`undercount_warning` is retained and labeled `"deprecated_alias_for": "coding_coverage_warning"`, but its `note` string content changed — old: "Treat transition rates as lower bounds"; new: "Report observed coded-channel rates, not identified bounds". Any consumer reading `checks["undercount_warning"]["note"]` gets different semantics under the same key. The key alias is preserved; the value contract isn't. Either note it as a value change in the PR description, or leave the old note text under the deprecated key.

Neither is structural. Everything else here is in good shape.

---

packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_iii.py
```diff
@@ -168,18 +168,26 @@ def validated_phase_iii_contracts(context=None) -> Output[pd.DataFrame]:
...
         "undercount_warning": {
-            "agencies_with_zero_phase_iii": zero_p3_agencies,
-            "agencies_total": len(agency_coverage),
-            "note": (
-                "FPDS sbir_phase coding is known to undercount Phase III, especially "
-                "outside DoD. Treat transition rates as lower bounds."
-            ),
+            **coding_coverage_warning,
+            "deprecated_alias_for": "coding_coverage_warning",
         },
```
hollomancer: **The deprecated alias is not value-compatible.** The key `undercount_warning` is retained, but its `note` now says "Report observed coded-channel rates, not identified bounds" instead of "Treat transition rates as lower bounds." Consumers reading `checks["undercount_warning"]["note"]` get different semantics under the same key.

Either note the value change in the PR description, or leave the old note text under the deprecated key and put the new wording only on `coding_coverage_warning`.

docs/readouts/navy-transition-v1.md
```diff
@@ -0,0 +1,109 @@
...
+  linkage/prediction design still cannot validate this readout.
+- An [NRC NIH assessment](https://www.nationalacademies.org/read/11964/chapter/6) says first sales can
+  precede Phase II completion, making negative times plausible but not validating these matches. The
+  event-only ECDF excludes censored follow-up and is not [Kaplan–Meier](https://doi.org/10.1080/01621459.1958.10501452).
+- [NRC 2014](https://www.nationalacademies.org/read/18821/chapter/5) reports commercialization
+  near 45–50%, which is not comparable to this signal rate. [Howell (2017)](https://www.aeaweb.org/articles?id=10.1257/aer.20150808)
+  supports finance as an outcome; [SEC methodology](https://www.sec.gov/files/dera-white-paper_regulation-d_082018.pdf)
+  shows Form D is incomplete and self-reported, with no Navy-SBIR attribution here.
+
+**Bottom line.** The results are a reproducible local map of observed public signals, not a complete
```
hollomancer: **Reserved-word misuse, contradicted three lines above.** `reproducible` is a study-status rank (`studies/README.md`; mapped to Computable). This readout has no study contract, and lines 10–11 of the same document say "untracked snapshots prevent a clean-checkout rebuild."

Suggest "a **local** map of observed public signals" or "a locally regenerable map."

hollomancer (commented):
Exploratory labeling is correct and the pipelines-tier wording cleanup is a net improvement. Two wording fixes remain: reserved-word "reproducible" in the readout, and the deprecated `undercount_warning` alias changing its `note` value. Neither is structural.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_iii.py
hollomancer: Confirmed and fixed in `b282cdb7`, but I took neither option exactly — the second one is unsafe here.

The old note read *"FPDS sbir_phase coding is known to undercount Phase III, especially outside DoD. Treat transition rates as lower bounds."* That sentence asserts a **one-sided** bias, which is the claim this readout retracts: the new wording exists because missed codes bias down while same-firm false matches bias up. So leaving the old text under `undercount_warning` would keep serving a withdrawn claim to precisely the consumers who have not migrated off the deprecated key — the silent-semantics-change problem you named, inverted.

Instead the alias now declares the change in-band:

```python
"undercount_warning": {
    **coding_coverage_warning,
    "deprecated_alias_for": "coding_coverage_warning",
    "note_semantics_changed": (
        "The 'note' value under this key changed and is no longer the "
        "lower-bound wording. Coded-channel rates are observed rates, "
        "not identified bounds; bias can run in either direction. Read "
        "'coding_coverage_warning' instead."
    ),
},
```

A consumer reading the alias now finds out the value moved, without having to diff two releases of the checks file. I also noted it in the PR description per your first option.

Checked for other consumers: nothing in the repo reads `undercount_warning`, so this alias exists for external readers of the checks JSON only.

**One thing I did not touch, flagging rather than widening the PR:** `docs/guides/government-policy-demo-plan.md:225` still says *"Phase III totals: describe as lower bounds when derived from public coding or keyword discovery."* That is the same retracted claim, in the document that scripts what gets said in a live policy room. It is outside this PR's diff, so I left it — but it is a bigger exposure than the JSON key, and worth its own change.

---

docs/readouts/navy-transition-v1.md
hollomancer: Confirmed and fixed in `b282cdb7`. Now reads "a **locally regenerable** map of observed public signals".

The contradiction was as tight as you described — `:10-11` of this same document says untracked snapshots prevent a clean-checkout rebuild, which is the negation of the property the word claims. Took "locally regenerable" over "local" because it keeps the useful assertion (someone on this host can rerun it) without borrowing the study-status rank.

`reproducible` now appears nowhere else in the file.

---

hollomancer:
Both items I raised are fixed, verified at `b282cdb7`:

- **`reproducible` is gone.** Line 107 read "a reproducible local map"; it now reads "a locally regenerable map", and the word appears nowhere in the readout. That was the reserved study-status rank being used outside a study contract, contradicting the "untracked snapshots prevent a clean-checkout rebuild" framing three lines above.
- **The deprecated alias now declares its own change.** `note_semantics_changed` under `undercount_warning` states plainly that the `note` value changed and is no longer the lower-bound wording, and points readers at `coding_coverage_warning`. That's a better answer than what I suggested — I proposed either restoring the old text or noting it in the PR description; making the artifact self-describing means a consumer reading the JSON finds out, not just someone reading this thread.

**I also merged `main` in and pushed (`4d9b1278`), because this had gone into conflict.** v0.12.0 changed `phase_iii.py` and `pairs.py` — the same two files this PR edits — to write transition parquets unconditionally. Merge commit, not a rebase, so your checkout stays valid.

I verified the two changes actually coexist rather than trusting the clean auto-merge: the unconditional `to_parquet` is at `phase_iii.py:158` and twice in `pairs.py`, and your `coding_coverage_warning` / `note_semantics_changed` block is intact at `phase_iii.py:173-201`. 296 tests pass across the phase_transition, reporting, and agency_private_capital suites — including the new `test_empty_output_write.py`, which exercises the same asset your changes touch.

Green and mergeable from my side.

---

scripts/data/build_navy_transition_readout.py
```diff
@@ -0,0 +1,1110 @@
...
+    form_d_latest: date | None = None
+    for record in form_d_records:
+        key = normalize_name(record.get("company_name"))
+        tier = _tier((record.get("match_confidence") or {}).get("tier"))
+        for offering in record.get("offerings") or []:
+            filing = pd.to_datetime(offering.get("filing_date"), errors="coerce")
+            if pd.isna(filing):
+                continue
+            filing_date = filing.date()
+            form_d_latest = max(form_d_latest or filing_date, filing_date)
```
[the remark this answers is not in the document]
hollomancer: Confirmed reachable and fixed in `1da6248`, but not by capping. `--as-of` is a free CLI argument, independent of `data_cut`, and both maxima run over every record in the snapshot (any firm, any tier), so `latest_observed` can land on either side of `window_end`. The counts are unaffected either way; line 671 guards `start <= filing_date <= as_of`.

Capping would hide the more dangerous direction. For the committed v1 build, `form_d_latest_observed_filing` is 2026-08-20 and `efts_latest_observed_mention` is 2026-08-21, both *before* the 2026-08-31 `as_of`. That means the last ten days of the window are empty because the snapshots stop early, not because nothing was filed. A cap would have left that invisible.

The summary now emits `form_d_snapshot_reaches_window_end` and `efts_snapshot_reaches_window_end` (`latest >= as_of`) next to the dates, with a code comment stating the keys are snapshot-wide. The committed summary JSON gains both keys from its existing values (both `false`). Test asserts the booleans agree with the dates.

---

hollomancer:
Fresh review of head `4d9b1278`, then pushed `1da6248`. Three things, one of which needs a local rerun.

**Unsupported "all" claim.** The readout said "all 8 later representatives above 250 trace to pre-cap contracts." The builder computed only the pre-cap subset (`post_cap & over_cap & first_action < 2019-06-28`) and never the total, so the JSON could not support "all". The builder now emits `post_cap_over_cap_n` and renders "N of M". The committed readout drops the word "all"; the M value is absent from the committed summary until the readout is rebuilt against the local snapshots, which only the host with those snapshots can do.

**Snapshot reach.** See the resolved Copilot thread. Both snapshots stop before the declared `as_of`, so the window tail is empty for a data reason. Worth a sentence in Panel D on the next rebuild.

**Not changed, flagging:** Panel B's 2,012 denominator includes Phase II awards with no UEI as permanent "no signal", and awards with a missing `contract_end_date` are dropped between Panel A's 2,780 and Panel B's 2,012 with no reported count. Neither number is in the JSON. Emitting `phase_ii_award_without_uei_n` and `phase_ii_award_missing_end_date_n` would close that; sizing needs the untracked parquet.

Also: `sbir_etl/utils/fpds_constraints.py` now declares `EPISTEMIC_TIER = "pipelines"` explicitly. Readout, FPDS-constraint, and eval-validity tests pass; tier boundary and docs checks pass.

---

scripts/data/build_navy_transition_readout.py
```diff
@@ -0,0 +1,1122 @@
...
+  linkage/prediction design still cannot validate this readout.
+- An [NRC NIH assessment](https://www.nationalacademies.org/read/11964/chapter/6) says first sales can
+  precede Phase II completion, making negative times plausible but not validating these matches. The
+  event-only ECDF excludes censored follow-up and is not [Kaplan–Meier](https://doi.org/10.1080/01621459.1958.10501452).
+- [NRC 2014](https://www.nationalacademies.org/read/18821/chapter/5) reports commercialization
+  near 45–50%, which is not comparable to this signal rate. [Howell (2017)](https://www.aeaweb.org/articles?id=10.1257/aer.20150808)
+  supports finance as an outcome; [SEC methodology](https://www.sec.gov/files/dera-white-paper_regulation-d_082018.pdf)
+  shows Form D is incomplete and self-reported, with no Navy-SBIR attribution here.
+
+**Bottom line.** The results are a reproducible local map of observed public signals, not a complete
```
hollomancer: **[bug]** `render_markdown()` still writes “a reproducible local map of observed public signals”. `reproducible` is a reserved study-status rank; the committed markdown was hand-patched to “locally regenerable”, but the generator was not. A local rebuild from the committed script would restore the reserved word and contradict the same template’s “untracked snapshots prevent a clean-checkout rebuild” framing a few lines above.

**Suggestion:** Change the template string to “locally regenerable” (or equivalent non-rank language) and add a test that `render_markdown(summary)` does not contain `reproducible`.

docs/readouts/navy-transition-v1.md
```diff
@@ -0,0 +1,109 @@
...
+| FY2023 | 411 | 221 | 211 | 197 |
+| FY2024 | 472 | 303 | 184 | 283 |
+| FY2025 | 533 | 317 | 185 | 595 |
+
+Latest-action descriptions: median 34 characters; ≥40,
+902/2,282 (39.5%); ≥150,
+177/2,282 (7.8%);
+1,740 representatives are nonzero modifications.
+FPDS [requires the field and caps newly entered text at 250 characters after 2019-06-28](https://beta.fpds.gov/downloads/Manuals/FPDS_User_Manual_V1.5.pdf);
+8 later representatives above 250
```
hollomancer: **[bug]** The committed readout is stale relative to the committed builder. `description_summary()` now emits `source_constraint.post_cap_over_cap_n` and `render_markdown()` renders “{legacy_n} of {over_cap_n} later representatives above 250”, but `navy-transition-v1.summary.json` has only `post_cap_over_cap_legacy_n: 8` (no `post_cap_over_cap_n`), and the markdown still says “8 later representatives above 250 trace to pre-cap contracts.” That sentence still cannot be checked against a total, which is the same unsupported-completeness problem the `post_cap_over_cap_n` field was added to close. Passing the committed summary through `render_markdown()` would KeyError.

**Suggestion:** Rebuild the markdown, summary, and figures from the builder after fixing the reserved word in the template. If the untracked snapshots are unavailable in this checkout, do not treat the checked-in prose as generator output until that rebuild happens; at minimum assert `post_cap_over_cap_n` in `description_summary()` tests.

tests/unit/scripts/test_build_navy_transition_readout.py
```diff
@@ -0,0 +1,504 @@
...
+    description = description_summary(coded_awards)
+    assert description["source_constraint"]["max_chars"] == 250
+    assert description["source_constraint"]["post_cap_representative_n"] == 2
+    assert description["source_constraint"]["post_cap_over_cap_legacy_n"] == 1
+    assert description["representative_nonzero_mod_n"] == 1
+    assert "ge_900_count" not in description
+    assert "not a bound" in description["interpretation"]
+
+
+def test_committed_readout_artifacts_exclude_the_impossible_900_character_metric() -> None:
```
hollomancer: **[suggestion]** Several polarity/consistency locks are missing, so the tests would still pass if the generator/artifact drift remained or if the pipelines coded-channel note were reverted to “Treat transition rates as lower bounds”. `test_committed_readout_artifacts_exclude_the_impossible_900_character_metric` only checks that 900 is absent; `test_annual_panel_separates_first_from_latest_action_and_description_cap` asserts `post_cap_over_cap_legacy_n` but not `post_cap_over_cap_n`; nothing compares committed markdown to `render_markdown()`; and no phase-transition asset test asserts `coding_coverage_warning.note` / `note_semantics_changed`. By contrast, `km_ready is False` and rejection of a 900-character threshold are actually locked.

**Suggestion:** Add (1) `assert "post_cap_over_cap_n" in description["source_constraint"]` on the existing description fixture, (2) a round-trip that committed markdown equals `render_markdown(committed_summary)` once the summary schema matches, and (3) an asset-level assertion that `undercount_warning.note` does not contain “lower bound” and that `note_semantics_changed` is present.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_iii.py
```diff
@@ -170,17 +170,35 @@ def validated_phase_iii_contracts(context=None) -> Output[pd.DataFrame]:
...
+            "can bias upward."
+        ),
+    }
     checks: dict[str, Any] = {
         "ok": True,
         "generated_at": now_utc_iso(),
         "total_rows": int(len(phase_iii)),
         "coverage": coverage_dict,
+        "coding_coverage_warning": coding_coverage_warning,
+        # Key-compatible alias for consumers of the original checks schema. The
```
hollomancer: **[suggestion]** The six-line comment above `undercount_warning` narrates the key-vs-value compatibility decision and quotes the withdrawn “lower bounds” sentence. That is change history / design rationale, not a constraint the next reader needs in order to edit the dict.

**Suggestion:** Drop the block, or keep a one-liner that the alias key is retained but the `note` value is not the old lower-bound wording (the `note_semantics_changed` field already says this to consumers).

docs/readouts/navy-transition-v1.md
hollomancer, line 38: **[bug]** The committed readout is not a product of the committed builder. `description_summary` always emits `source_constraint.post_cap_over_cap_n`, and `render_markdown` interpolates `{legacy_n} of {over_cap_n} later representatives above 250`. The checked-in `navy-transition-v1.summary.json` has no `post_cap_over_cap_n` (only `post_cap_over_cap_legacy_n: 8`), and the markdown says “8 later representatives above 250 trace to pre-cap contracts” with no denominator. `render_markdown` on the committed summary therefore raises `KeyError`. The bottom-line sentence also drifted (`locally regenerable map` in the markdown vs `reproducible local map` in the generator). If `post_cap_over_cap_n > 8`, the published sentence incorrectly implies every post-cap over-cap description is grandfathered. `test_committed_readout_artifacts_exclude_the_impossible_900_character_metric` does not round-trip `render_markdown(summary)` against the committed files, so this drift is uncaught.

**Suggestion:** Rebuild `docs/readouts/navy-transition-v1.md` and `navy-transition-v1.summary.json` from the current script (or change the generator to match the committed prose, including omitting `post_cap_over_cap_n` only if that is still the intended contract). Add a unit test that `render_markdown(json.loads(committed_summary))` equals the committed markdown, and that `source_constraint` contains `post_cap_over_cap_n`.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/pairs.py
```diff
@@ -384,6 +393,18 @@ def transformed_phase_transition_survival(
         "observed_events": observed,
         "censored": censored,
         "match_rate": round(match_rate, 4),
+        "time_origin": "phase_ii_period_of_performance_end",
+        "negative_event_times": negative_event_times,
+        "negative_time_rows": negative_time_rows,
+        "nonnegative_time_origin": nonnegative_time_origin,
+        "km_ready": False,
```
hollomancer, line 409: **[suggestion]** This PR sets survival `km_ready` to False and rewrites asset/docs/model text so the completion-relative frame is not advertised as Kaplan–Meier-ready. `packages/sbir-analytics/sbir_analytics/assets/jobs/phase_transition_job.py` still describes the job as producing a “KM-ready survival frame” (line 20). That is the Dagster-visible description operators see when materializing the same assets.

**Suggestion:** Update the job description to match the new estimator warning (signed completion-relative follow-up; not KM-ready without a nonnegative origin or pre-completion stratum).

packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_iii.py
hollomancer: **[suggestion]** The six-line comment above `undercount_warning` narrates the schema migration and quotes the withdrawn lower-bound wording. The `note_semantics_changed` / `deprecated_alias_for` fields already carry that history for consumers.

**Suggestion:** Drop or shrink the comment to a single WHY note, e.g. that the alias key is kept but must not restore the old one-sided-bias sentence.

hollomancer (commented):
## Summary

The exploratory labeling, same-UEI first-signal pairing, pre-award exclusion, and FPDS 250-character cap retraction are carefully done, with no HIGH-threshold polarity change and no silent promotion of the Navy counts into an evidence contract. The main defect is that the committed readout was not regenerated from the committed generator: `render_markdown` and `description_summary` now require `post_cap_over_cap_n`, but the checked-in JSON omits it and the markdown still states an unqualified “8 later representatives… trace to pre-cap contracts.” The KM-ready retraction is also incomplete in the Dagster job description, which this PR did not touch.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

hollomancer:

### Status: 🟡 One Bug — Regenerate Readout Before Merge

- **Draft:** No
- **Age:** 9 days
- **Size:** +2,464/-49 across 22 files

Adds an exploratory Navy Phase III transition readout with careful handling of FPDS coding, the 250-character cap, and retraction of the one-sided undercount claim.

**Hollomancer review (2026-09-10) — 1 bug, 2 suggestions:**

> **Bug:** The committed readout was not regenerated from the committed generator. `render_markdown` and `description_summary` now require `post_cap_over_cap_n`, but the checked-in JSON omits it and the markdown still states an unqualified "8 later representatives… trace to pre-cap contracts."

Also noted: the KM-ready retraction is incomplete in the Dagster job description (which this PR did not touch — out of scope).

Copilot also flagged: external-signal dates after the `as_of` window create internal inconsistency.

The exploratory labeling, same-UEI first-signal pairing, pre-award exclusion, and FPDS 250-character cap retraction are all noted as carefully done.

**Recommended action:** Regenerate the readout from the committed generator so JSON and markdown match. Then merge.

hollomancer:
## Consolidated review — request changes

Reviewed at `1da62489`.

- **Blocker — checked-in artifacts were not generated by this head.** `scripts/data/build_navy_transition_readout.py:420-454` now emits `post_cap_over_cap_n`, and the renderer at 879-886 requires it, but `docs/readouts/navy-transition-v1.summary.json:96-104` lacks the field. The committed Markdown also differs from current renderer wording. Rerun the pinned build, commit all JSON/Markdown/figures, and add a schema or golden test tying artifacts to the generator.
- **Blocker — Panel D states a complete window from incomplete snapshots.** The summary declares `window_end: 2026-08-31`, while Form D ends August 20 and EFTS August 21 and both coverage flags are false. The renderer at 924-931 ignores those flags and reports 47 firms for the full window. Cap the window, suppress the count, or visibly caveat it, with a regression test.
- **Important — pre-completion events count as within-five-year transitions.** `packages/sbir-analytics/sbir_analytics/assets/phase_transition/pairs.py:375-408` checks `event_observed & time_days <= horizon`; negative times therefore qualify. Require `0 <= time_days <= horizon` and report prevalent events separately.
- **Important — the FPDS manual URL is dead.** `sbir_etl/utils/fpds_constraints.py:16-20` redirects to a generic SAM page. Link the current official data dictionary and separately source the June 2019 effective-date claim, then regenerate artifacts.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/phase_iii.py
hollomancer: Comment above `undercount_warning` is now a single WHY note: keep the alias key; do not restore the withdrawn one-sided-bias note. `8c9ae401`.

docs/readouts/navy-transition-v1.md
hollomancer: Cannot rebuild the committed readout from the committed builder in this checkout. Missing pinned inputs: `data/processed/phase_iii_census_sbir_awards.parquet`, `data/processed/phase_iii_census_sbir_awards.checks.json`, `data/form_d_details.jsonl`, `data/sbir_ma_events.jsonl`, and the four `navy-transition-sr3|st3-awarding|funding.parquet` pulls plus their manifests.

Did not invent `post_cap_over_cap_n`. `render_markdown` now raises `ValueError` when that headline count is missing rather than interpolating a `KeyError` (`8c9ae401`). The generator bottom line now says "locally regenerable". Thread left open until a pinned-input rebuild can emit both JSON and markdown.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/pairs.py
hollomancer: `phase_transition_latency_job` no longer advertises a KM-ready survival frame. The Dagster-visible description now says signed completion-relative follow-up, not KM-ready without a nonnegative origin or pre-completion stratum. `8c9ae401`.

hollomancer:
## Review pass (code-reviewer)

The unresolved thread is still a merge blocker for the **published readout**, not for the pipelines-tier edits.

Verified at `c01142d0`:

- Committed `navy-transition-v1.summary.json` still has `post_cap_over_cap_legacy_n: 8` and no `post_cap_over_cap_n`.
- `render_markdown` still requires that field and raises `ValueError` if it is missing. That is a rebuild guard, not evidence the committed files were produced by this generator.
- Committed markdown still says “8 later representatives above 250 trace to pre-cap contracts” with no denominator.
- After the thread reply, `c01142d0` also taught the renderer a Panel D snapshot-gap sentence. The committed markdown still omits it. The flags are already in the JSON (`window_end` 2026-08-31, Form D 2026-08-20 / false, EFTS 2026-08-21 / false), so that sentence does **not** need a parquet rebuild — but `render_markdown` cannot emit it until `post_cap_over_cap_n` is present.

### Codex 2026-09-10 blockers

| Item | Status |
|---|---|
| Artifacts not generated by HEAD | Still live |
| Panel D complete window from incomplete snapshots | Code fixed; published prose still live |
| Pre-completion events counted as within-five-year transitions | **Fixed** in `a09b6f21` (`time_days >= 0`) |
| `beta.fpds.gov` manual URL | Still live; `www.fpds.gov` is the live host. Cap/date claim is still documented there. Not a merge blocker by itself |

Do not invent `post_cap_over_cap_n`. Close-out is a host rebuild of JSON + markdown + figures together, plus a golden test that `render_markdown(committed_json) == committed_markdown`.

If a merge is needed before that host run, split the pipelines-tier files (Phase III warning, KM metadata, five-year floor, FPDS helper) and keep the readout artifacts off `main`.

docs/readouts/navy-transition-v1.md
hollomancer: Re-reviewed at `c01142d0`. The ValueError guard is still not a substitute for matching artifacts.

Committed JSON still lacks `post_cap_over_cap_n`; committed markdown still has no denominator on the 8 later representatives. After this thread, `c01142d0` also added a Panel D snapshot-gap sentence the committed markdown omits — those flags are already in the JSON, but `render_markdown` cannot emit the caveat until `post_cap_over_cap_n` is present.

Leave open until a host with the pinned snapshots rebuilds JSON, markdown, and figures together, plus a golden test that `render_markdown(committed_json) == committed_markdown`. Do not invent the missing count.

scripts/data/build_navy_transition_readout.py
```diff
@@ -0,0 +1,1187 @@
...
+    ].reset_index(drop=True)
+    assignments = (
+        pairs.sort_values(
+            ["phase_iii_action_date", "phase_iii_contract_id"],
+            kind="stable",
+        )
+        .drop_duplicates("phase_ii_award_id", keep="first")
+        .reset_index(drop=True)
+    )
+    survival = _build_survival(phase_ii, pairs, data_cut)
```
hollomancer, line 544: **Blocking.** The pre-award filter removes only actions dated before the Phase II *award date*, then `assignments` keeps the earliest surviving action per Phase II award. Whenever a firm has both an action during Phase II performance and a later post-completion action, the in-performance one is assigned and the post-completion transition is discarded. `tests/unit/scripts/test_build_navy_transition_readout.py:389-444` encodes exactly that case (`P3-DURING` assigned, `P3-AFTER` dropped) and asserts it as correct without noting the consequence. 525 of 799 assignments (65.7%) are pre-completion and the median signed latency is −322 days.

The decile table and the figure subtitle (“Signed years: Phase II period end → assigned coded action”) present a latency distribution whose central tendency is partly produced by the assignment rule rather than by observed Navy timing. House rule: what a number does not cover must be stated, and this is the readout’s headline figure.

**Fix:** One sentence in Panel B and the figure subtitle: the earliest-qualifying-action rule assigns an in-performance action in preference to a later post-completion action for the same firm, so the negative mass is partly a selection property of the rule.

scripts/data/build_navy_transition_readout.py
```diff
@@ -0,0 +1,1187 @@
...
+
+
+def build_latency_panel(
+    navy_sbir: pd.DataFrame, transactions: pd.DataFrame, data_cut: date
+) -> tuple[dict[str, Any], pd.DataFrame]:
+    phase_ii = _prepare_sbir_gov_rows(navy_sbir.loc[navy_sbir["analysis_phase"].eq("II")].copy())
+    end_dates = pd.to_datetime(phase_ii["period_of_performance_end"], errors="coerce")
+    phase_ii = phase_ii.loc[end_dates.notna() & end_dates.dt.date.le(data_cut)].reset_index(
+        drop=True
+    )
```
hollomancer, line 494: **Blocking.** Panel A reports 2,780 Phase II awards; Panel B reports 2,012 (“Of 2,012 completed Phase II awards”). The 768-row gap from `end_dates.notna() & le(data_cut)` here, plus any `_prepare_sbir_gov_rows` drops, is nowhere in the JSON or markdown. Phase II awards with no UEI can never match (`pairs.py` `_join_on` filters the join key on `notna`) and sit in the denominator as permanent “no signal,” also uncounted.

Same pattern at builder `:1114`: `coded_transaction_n_all_dates_through_cut` 20,586 against 37,774 raw rows across the four pulls, with no breakdown of the 17,188 removed by union dedupe, DoN post-filter, and cut-date filter.

The owner asked for exactly this on 2026-09-06 (`phase_ii_award_without_uei_n` and `phase_ii_award_missing_end_date_n`); neither key exists at head. Recorded exclusions are an explicit analysis-hygiene requirement.

**Fix:** Emit `phase_ii_award_missing_end_date_n`, `phase_ii_award_end_after_cut_n`, `phase_ii_award_without_uei_n`, and a transaction-reduction breakdown. The values are free from frames already in memory.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/pairs.py
```diff
@@ -368,12 +372,26 @@ def transformed_phase_transition_survival(
...
     horizon_days = 5 * 365
     five_year_rate: float | None = None
     if total:
-        within = survival.loc[survival["event_observed"] & (survival["time_days"] <= horizon_days)]
+        within = survival.loc[
+            survival["event_observed"]
+            & (survival["time_days"] >= 0)
+            & (survival["time_days"] <= horizon_days)
+        ]
         five_year_rate = float(len(within) / total)
```
hollomancer, line 395: **Blocking — not raised by prior reviews.** The numerator floor (`time_days >= 0`) from `a09b6f21` is correct. The denominator is still every Phase II award with a non-null end date, with no restriction to `end_date + 5y <= data_cut`. Awards with months of follow-up are counted as non-transitions in a five-year rate.

Separately, nothing restricts `phase_ii_end_date <= data_cut`, so an award ending after the cut gets `time_days = data_cut - end_date < 0` with `event_observed` False: censored at a time before it entered observation and counted as a failure. `nonnegative_time_origin` (`:378-379`) is computed over *all* negative `time_days`, so an end-date-after-cut artifact flips it and `estimator_warning` then attributes it to prevalent events.

`docs/research-questions.md:639-645`, the entry this asset serves, states the opposite: awardees whose follow-up is not yet complete are right-censored, not counted as failures. This is the one number in the touched pipeline code shaped like a finding, written to `phase_transition_survival.checks.json` and Dagster asset metadata on every materialization.

The PR body claims it marks these frames not KM-ready without a nonnegative origin *and* a censoring assessment; the censoring assessment is the missing half.

**Also a breaking-value change under an unchanged key.** The `>= 0` floor changes `within_5_year_rate` in the survival checks JSON and asset metadata, and appears nowhere in the PR body’s “Breaking value change in the checks schema” section — the section that exists for this failure mode. On this PR’s own Navy cohort, 525 of 799 assignments (65.7%) are pre-completion, so the published value can fall by a large multiple for identical inputs.

**Fix:** Restrict the denominator to awards with a complete horizon (`end_date + horizon <= data_cut`) and emit the excluded count; or keep the cohort-wide denominator and rename the key (e.g. `observed_post_completion_event_rate_by_cut`) while recording follow-up composition. Split negative `time_days` into pre-completion events vs end-date-after-cut rows before setting `nonnegative_time_origin`. Add the value change to the PR body’s breaking-value section, or emit both definitions under distinct keys for one release.

hollomancer (commented):
## Open PR review — request changes

Reviewed against `origin/main` at `493340ea`. Head `c01142d0`. Epistemic core is sound: nothing is promoted, blocked panels stay blocked, historical DoD comparator stays unreproduced, arithmetic in the committed summary JSON checks out. Three things still block merge. Complements the 12:04 artifact-rebuild pass; this pass adds the five-year-rate denominator, assignment-rule, and exclusion-count defects that no prior reviewer had raised.

Cannot formally request changes on an own PR; treat this as **request changes**. CONTRIBUTING says to use a draft while validation is incomplete — with the artifact rebuild outstanding, draft is the honest state.

### Blocking

1. **Committed artifacts are not producible by the committed generator.** `render_markdown` against the checked-in summary raises `ValueError` on missing `post_cap_over_cap_n`. Thread on `navy-transition-v1.md:38` still open. Closes with a pinned-input rebuild of summary + markdown + both figures, plus a golden test that `render_markdown(json.loads(committed_summary)) == committed_markdown`. Drift is bounded to two content changes: the “N of M” denominator at md:38 and the Panel D coverage caveat at md:84.
2. **Panel D still publishes a complete-window claim** over Form D / EFTS snapshots that stop 10 and 11 days short (`form_d_snapshot_reaches_window_end: false`, latest filing 2026-08-20; EFTS 2026-08-21). Renderer already builds the caveat; committed markdown omits it. Closes with (1).
3. **`within_5_year_rate` denominator counts awards with no possibility of five years of follow-up**, and censors some at negative time. The `>= 0` numerator floor is a breaking value change under an unchanged key, undeclared in the PR body’s breaking-value section. See line comment on `pairs.py`.
4. **`docs/phase-transition-latency.md:85` still reads “Any Phase III within 5 years.”** After the numerator floor that is no longer what the check answers. (Line is outside the current diff hunk; fix in the same commit as (3).)
5. **Earliest-qualifying-action rule prefers in-performance actions** over later post-completion ones. 525/799 assignments (65.7%) are pre-completion; median signed latency −322 days. See line comment.
6. **2,780 → 2,012 and 37,774 → 20,586 are unreported exclusions.** Owner ask from 2026-09-06 still open. See line comment.

### Number audit

10 of 14 checked figures hold as stated. “8 later representatives above 250 trace to pre-cap contracts” is still misstated (no denominator). “Repeated pinned-input builds reproduced the summary and figures” is false at this head.

### Do these first

1. Rerun the pinned build on the host holding the snapshots; commit summary, markdown, and both figures together; add the golden round-trip test. Acceptance: only md:38 and md:84 should move.
2. Fix `within_5_year_rate`’s denominator (or rename the key); split negative `time_days`; update `docs/phase-transition-latency.md:85`; add the value change to the breaking-value section.
3. Emit `phase_ii_award_missing_end_date_n`, `phase_ii_award_end_after_cut_n`, `phase_ii_award_without_uei_n`, and an FPDS transaction-reduction breakdown.
4. One sentence in Panel B / figure subtitle on the assignment rule. Replace the dead `beta.fpds.gov` URL (`fpds_constraints.py:19`, `eval-validity.md:53`) as a follow-up, and open a tracked issue for `docs/guides/government-policy-demo-plan.md:225` rather than leaving the retracted lower-bound claim as PR-body prose.

B1/B2 require the snapshot host. B3–B6 do not and can land now.

hollomancer:
Merged `origin/main` (v0.16.0) in at `524fc716`. Conflicts resolved; the PR is mergeable again.

**Merge, not rebase** — this branch is checked out in the main working tree under uncommitted work, so rewriting its history would strand that checkout. Same reasoning as the earlier merge at `4d9b1278`.

**Two conflicts, both wording, both in files v0.16.0 changed under this branch.** `models/phase_transition.py` and `phase_iii.py` docstrings: main called the Phase III coding channel a "known undercount"; this branch had deliberately replaced that one-sided wording with "incomplete and inconsistent" everywhere, which is the retraction the branch exists to make. Kept this branch's wording and carried main's parent-IDV inheritance sentence onto it, so both the retraction and the new `phase_iii_evidence` column are described. No "known undercount" remains anywhere in the tree.

**CI is now red, for the reason this PR is already blocked on.** The merge brings in the study-artifact round-trip guard from #743, which independently detects the standing blocker:

```
docs/readouts/navy-transition-v1.md: build_navy_transition_readout.py:render_markdown
raised ValueError on the committed sidecar: panel_a.description.source_constraint.post_cap_over_cap_n
is required; refuse to interpolate a missing headline count
```

I registered the readout as a round-trip pair rather than waiving it. The 2026-09-10 review asked for "a schema or golden test tying artifacts to the generator" — this is that test, and it holds the branch red until the readout is rebuilt against the local snapshots, which only the host holding them can do. Waiving would have made CI green while the artifacts still did not reproduce.

Both failing jobs (`Lint, Types, and Guards`, `Fast Tests (4/4)`) are that single cause. The rest of the shard is 1,780 passed. Locally: 7,293 passed, ruff and mypy clean.

Unchanged and still open: the five-year denominator counting Phase II awards with only months of follow-up, and the pre-award filter assigning the in-performance action over the later post-completion one.

docs/readouts/navy-transition-v1.md
hollomancer: Confirmed, and partially addressed since: 51796760 makes the builder emit post_cap_over_cap_n and renders 'N of M', aec0d3d7 makes render_markdown fail closed on the missing count, and cfbc773d registers the readout as a round-trip pair so CI now fails on exactly this mismatch (it currently does). What remains is regenerating the sidecar, which needs the four pinned FPDS pulls (navy-transition-{sr3,st3}-{awarding,funding}.parquet + manifests). They are not on this machine — searched the repo tree, data/, and home; S3 is unreachable (aws session expired). Blocked on locating the snapshots or an authorized re-pull, which would be a new vintage and move every number. Leaving unresolved.

scripts/data/build_navy_transition_readout.py
hollomancer: The finding stands: assignments takes the earliest surviving action per Phase II award, so an in-performance action shadows a later post-completion transition, and 65.7% of assignments are pre-completion with median signed latency -322 days. Changing the assignment rule (earliest post-completion vs earliest overall, or a stratified presentation) is a methodology decision that changes the readout's central quantity, and verifying any change requires regenerating from the pinned FPDS pulls, which are absent from this machine. Needs a maintainer call plus the data; leaving unresolved.

scripts/data/build_navy_transition_readout.py
hollomancer: Still open at head: phase_ii_award_without_uei_n and phase_ii_award_missing_end_date_n do not exist in the committed sidecar, and the 768-row Panel A/Panel B gap and the 37,774→20,586 reduction remain unbroken-down. The builder change is writable, but the committed sidecar cannot gain the keys without a rerun against the pinned FPDS pulls, which are not on this machine. Related: 30dbda80 adds the analogous accounting to the transformed_phase_transition_survival checks (phase_ii_missing_end_date_n, phase_ii_end_after_cut_n), so the asset side of this pattern is covered. Leaving unresolved pending the data.

packages/sbir-analytics/sbir_analytics/assets/phase_transition/pairs.py
hollomancer: Fixed in 30dbda80, both parts. The five-year denominator is now only awards mature for the horizon (end_date + 5y <= data_cut), reported as five_year_denominator with phase_ii_immature_for_horizon_n beside it. _build_survival drops awards whose end date falls after the cut — they were censored at negative time and counted as failures before entering observation, and a single such row flipped nonnegative_time_origin artifactually — counted in phase_ii_end_after_cut_n. Three tests pin the behavior: numerator-only exclusion at a cut where maturity cannot confound it, the immature-denominator exclusion (rate 1/1, not the deflated 2/4), and the end-after-cut drop leaving the origin diagnostics clean. 72 tests pass.

PR #686, opened 2026-08-31
Stub multi-FY USAspending contract archive union

hollomancer:
## Description

Adds a pipelines-tier maintenance spec for replacing scalar USAspending archive selection with an explicit, provenance-bound fiscal-year set. This is a planning stub only; implementation remains unchecked in `tasks.md`.

## Type of Change

- [x] Documentation update
- [ ] Bug fix

## Versioning Impact

- [x] No release impact

## Testing

- [x] `make docs-check`
- [x] `git diff --check`

## Scope

The future implementation will select one revision per requested fiscal year, union at transaction grain, fail on conflicting duplicate transactions, and bind every archive hash to cache validity. Transition classification and analysis are out of scope.

### DIFF
--- a/specs/status.md
+++ b/specs/status.md
@@ -170,6 +170,10 @@ bypassing lifecycle review; the status and rationale still require human judgmen
   stays gated on the decoy/estimand work in `phase3-transition-groundtruth` and a
   `studies/transition-scoring/` promotion. No GitHub-runner execution, no
   committed corpus bytes, no server schedule.
+- **`usaspending-multi-fy-archive-union` — Maintenance.** Pipelines-tier remediation for
+  the transition source layer's scalar archive selection. Resolve one declared revision per
+  requested fiscal year, union at transaction grain, and bind the complete archive set to the
+  output manifest; no transition classification or analysis is in scope.
 - **`weekly-awards-report-refactor` — Maintenance.** Monolith is already split
   into weekly reporting modules. Remaining work is injection, coverage, and
   alias cleanup.
--- a/specs/usaspending-multi-fy-archive-union/design.md
+++ b/specs/usaspending-multi-fy-archive-union/design.md
@@ -1,0 +1,42 @@
+# USAspending Multi-FY Contract Archive Union — Design
+
+## Current flow
+
+The transition asset calls `find_latest_local_contract_archive`, which sorts all matching archive
+filenames by update date and returns one path. The extractor and manifest consequently model the
+source as a scalar even when the local directory contains multiple fiscal partitions.
+
+## Proposed flow
+
+1. Resolve the configured fiscal-year set before extraction.
+2. Group valid `Contracts_Full` filenames by fiscal year and choose the latest revision inside
+   each group.
+3. Stream each selected archive through `AwardArchiveContractExtractor` in deterministic order.
+4. Reconcile repeated transaction IDs across archives, retaining identical repeats and rejecting
+   conflicts.
+5. Publish one Parquet output and one manifest bound to the complete ordered archive set.
+
+The resolver should return a typed sequence carrying fiscal year, revision date, and path rather
+than infer those values repeatedly from strings. The asset should accept an explicit fiscal-year
+configuration; implicit "all files currently present" behavior would make the cut unstable.
+
+## Failure behavior
+
+- Missing requested fiscal year: fail before scanning.
+- Multiple same-date revisions for one year: require byte identity or fail as ambiguous.
+- Repeated transaction ID with different canonical content: fail and emit the conflicting archive
+  names and transaction identifier.
+- Existing scalar-cache manifest: invalidate and rebuild; do not reinterpret it as a set.
+
+## Provenance and migration
+
+The manifest schema gains an ordered `award_archives` collection and an archive-set SHA-256.
+During one compatibility release, the existing singular `award_archive_file` may be emitted only
+when the set contains one item. Consumers must use the set-level fingerprint for cache validity.
+
+## Testing strategy
+
+- Resolver fixtures with two fiscal years and multiple revisions within one year.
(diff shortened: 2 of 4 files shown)


hollomancer:
## Group review of #686–#691 (posting the cross-cutting analysis here; per-PR verdicts on each)

### Verdict for this PR: **grounded mechanically, low necessity — recommend deferring**

The defect is real and described exactly right. `sbir_etl/extractors/usaspending_award_archive.py:304-317`:

```python
for path in directory.glob("FY*_Contracts_Full_*.zip"):
    ...
return max(candidates, key=lambda item: (item[0], item[1].name))[1]
```

One global `max` over `updated` across **all** fiscal years — with FY2025 and FY2026 both present, one is silently dropped.

But the necessity case collapses on inspection: the sole consumer is `assets/transition/contracts.py:250`, gated behind `use_award_archive` which defaults **False** (line 248); it's `false` in `.env.server.example:89` and `docker-compose.server.yml:53`; and the canonical Phase III source is the PostgreSQL dump per `specs/phase-iii-source-materialization/design.md:25`. A real defect on a disabled fallback path with no named consumer and no operator using it. Suggest a one-line note under `phase-iii-source-materialization` that the resolver is FY-scalar, revisit if `USE_AWARD_ARCHIVE` is ever enabled — rather than a spec directory and a registry entry.

---

### Cross-cutting: `specs/status.md` does **not** conflict

I replayed all six merges with `git merge-tree --write-tree` chained through synthetic commits, in listed order and reverse. Both produced clean trees with all six entries present and no markers. Insertion points are 9+ lines apart at the closest (#691 ~75, #688 ~84) — outside git's conflict window. The registry guard (`check_removed_src_references.py:335-377`) is bidirectional and each PR carries both halves, so **any subset can land in any order.**

The real mechanical cost isn't conflicts, it's six review cycles and six copies of the same labeling error.

### Systematic error in all six — the one thing I'd block on

Every spec declares `**Lifecycle status:** Maintenance` alongside `**Spec-file progress:** Not yet started` and zero checked tasks. `specs/status.md:18-19` defines Maintenance as *"core behavior is mostly implemented; remaining work is test, documentation, or small cleanup."* Nothing is implemented in any of them. The correct label is **Gated backlog** (`:20-21`), or **Active** for whichever is actually being started. Six new "Maintenance" entries make the registry read as if six things are nearly done.

### Overlap

These are three surfaces, not six units of work:

1. **#686 + #687 + #688 — transition source/lineage.** That layer already has an owning spec: `specs/phase-iii-source-materialization/` (declares `pipelines`; `design.md:33` already claims *"Bind archive relation/member/TOC fingerprints, vendor-filter hash, extraction audit"*). `specs/status.md:127-129` says of it: *"Keep it aligned with the census and other transition consumers rather than extending it."* It has **no `tasks.md`** — tasks can be appended there with zero new registry entries.
2. **#690 + #691 — SEC source fidelity.** Both touch `sbir_etl/enrichers/sec_edgar/` and `sbir_etl/capital_events/`, same downstream. One spec.
3. **#689** — not a spec at all; three SQL queries in a 254-line exploratory script with one importer.

**One live collision:** #691 rewrites `form_d_inputs.py`, which is the direct input to the **Active** `agency-private-capital-comparison` spec, and there's a live branch `origin/agent/form-d-control-universe` (#582) in the same area. Coordinate before either lands.

### Highest value-to-effort item in the whole set, and it needs no spec

The one real defect inside #688 is three lines — `phase_iii.py:155`, `pairs.py:297`, `pairs.py:364` skip writing the parquet when the frame is empty while `checks.json` is overwritten unconditionally with `total_rows: 0`, leaving a stale parquet beside fresh zero-row checks. Land that as a standalone fix.

Tier declarations all pass CI (correct format, research-question anchors present), but #689 declares `pipelines` for a module that declares `EPISTEMIC_TIER = "exploratory"` — a silent promotion — and #690's `pipelines` is unwarranted since both real consumers are `scripts/data/*`.

---

hollomancer:
Scope-guard follow-up, refining the fold recommendation in the previous comment.

**Confirmed:** all six stubs mislabel `Lifecycle status: Maintenance` while `Spec-file progress: Not yet started`. Correct label is **Gated backlog**. I would not land any of these as written.

**Refined fold (do not dump everything into `phase-iii-source-materialization`):**

| PR | Action |
|---|---|
| #686 | Close. Optional one-line note that the archive resolver is FY-scalar. No task list. |
| #687 | Close. The narrow projection is a documented memory tradeoff, not an accidental fidelity bug. Notebook the real archive headers *before* any field work, and only if `USE_AWARD_ARCHIVE` is ever enabled. |
| #688 | Close the spec. **Do not** fold a fingerprint framework into `phase-iii-source-materialization` — that would extend a source-layer spec into transformation assets that do not cache. Land the empty-parquet skip (`phase_iii.py:155-156`, `pairs.py:297-298/:364-365`) as a standalone bugfix. |
| #689 | Close as a spec. ~20-line exploratory bugfix. |
| #690 / #691 | Do **not** merge into one SEC-source-fidelity spec. See comments on those PRs. |

Smallest set that captures the real defects: two or three small bug PRs and one Form D amendment notebook, not six registry rows.

hollomancer:
Closing as agreed — superseded by #692, which carries the consolidation of this batch.

The FY-scalar resolver defect is real and I've recorded it rather than dropped it, but as a note under `phase-iii-source-materialization` rather than a spec directory and registry entry. The reason stands: the sole consumer is `assets/transition/contracts.py:250`, gated behind `use_award_archive` which defaults False and is `false` in both `.env.server.example:89` and `docker-compose.server.yml:53`, and the canonical Phase III source is the PostgreSQL dump. Worth revisiting the moment `USE_AWARD_ARCHIVE` is enabled anywhere.

---

PR #687, opened 2026-08-31
Stub canonical contract source-field preservation

hollomancer:
## Description

Adds a pipelines-tier maintenance spec for repairing lossy USAspending contract projection. This is a planning stub only; implementation remains unchecked in `tasks.md`.

## Type of Change

- [x] Documentation update
- [ ] Bug fix

## Versioning Impact

- [x] No release impact

## Testing

- [x] `make docs-check`
- [x] `git diff --check`

## Scope

The future implementation will preserve research code versus label, awarding versus funding agency tiers, and transaction versus base-award descriptions as distinct canonical fields. Phase III classification, text scoring, and agency-specific filters are out of scope.

### DIFF
--- a/specs/contract-source-field-preservation/design.md
+++ b/specs/contract-source-field-preservation/design.md
@@ -1,0 +1,42 @@
+# Contract Source-Field Preservation — Design
+
+## Field contract
+
+The source layer should carry these concepts independently:
+
+| Concept | USAspending archive field | Canonical destination |
+|---|---|---|
+| Research code | `research_code` | `research_code` |
+| Research label | `research` | existing `research` during migration |
+| Awarding top tier | `awarding_agency_name` | existing `agency` |
+| Awarding sub-tier | `awarding_sub_agency_name` | existing `sub_agency` |
+| Funding top tier | `funding_agency_name` | `funding_agency` |
+| Funding sub-tier | `funding_sub_agency_name` | `funding_sub_agency` |
+| Transaction narrative | `transaction_description` | existing `description` |
+| Base-award narrative | `prime_award_base_transaction_description` | `base_award_description` |
+
+Exact source headers must be verified against the schema-extraction fixture before implementation;
+aliases must be explicit rather than fuzzy header matching.
+
+## Component changes
+
+1. Expand `CONTRACT_ARCHIVE_COLUMNS` and `_map_row` with the required raw fields.
+2. Extend `FederalContract` with stable optional fields where they are broadly consumed. Critical
+   scope fields should not be buried in an opaque metadata blob.
+3. Update Parquet serialization and source-provenance checks to carry the expanded schema.
+4. Add named compatibility accessors only where an existing consumer cannot migrate atomically.
+
+## Failure and migration behavior
+
+An archive governed by the new contract fails schema verification when a required header is
+absent. An older cached Parquet without the new schema must be invalidated, not padded with copied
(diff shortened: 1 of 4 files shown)


hollomancer:
## Verdict: grounded omission, but the framing is wrong — recommend folding into `phase-iii-source-materialization`

See #686 for the cross-cutting analysis of this batch (no `specs/status.md` conflicts; systematic `Maintenance` mislabel across all six).

**The omission is confirmed.** `sbir_etl/extractors/usaspending_award_archive.py:38-67` — `CONTRACT_ARCHIVE_COLUMNS` includes `awarding_agency_name`, `funding_agency_name`, `transaction_description`, `research`, and omits `research_code`, `funding_sub_agency_name`, `prime_award_base_transaction_description`. `_map_row` (line 405) maps `research` → `research`.

Two corrections to how the spec frames it:

**1. It is not an accident.** Lines 36-37 document the choice: *"Only these columns are decoded from each 297-column Contracts_Full CSV. Keeping the projection narrow makes Arrow's record batches small even for multi-GB members."* Calling it a "source-fidelity defect" mischaracterizes an explicit, commented memory tradeoff. Requirement 1 needs to argue *against* that tradeoff, not ignore it.

**2. It does not cause a silent zero-row Phase II/III result.** I expected it to and it doesn't. The archive's `research` column carries the Element 10Q *label*, while `phase_ii.py:62` matches the *codes* `{"SR2","ST2"}` — but `phase_ii.py:158` falls through to `_normalize_phase_label`, whose regex at line 143 (`\bPHASE\s+(III|II|I)\b`) recovers "II" from a label like `SBIR PHASE II ACTION`. `phase_iii.py:16` imports the same classifier. No live break.

That lenient regex is actually the **best argument for this spec** — it's a compensating hack that exists because the code was dropped. Argue from deleting the hack, not from a break that the fallback prevents.

**Testing gap worth recording regardless:** `tests/unit/extractors/test_usaspending_award_archive.py:64` synthesizes the CSV header *from `CONTRACT_ARCHIVE_COLUMNS` itself*, so no fixture derives from a real archive header, and `usaspending_award_archive.py:347` verifies column *presence* only, never value semantics. The repo cannot currently detect a code/label swap. The spec's own task 1 ("Verify and fixture the exact USAspending raw headers") concedes the headers are unverified — so the spec is written ahead of its own verification. Per CLAUDE.md's notebook-first rule, that audit belongs in a notebook before it becomes a frozen pipelines task list.

**Cut task 5** ("Migrate downstream readers without redefining legacy fields") — no downstream reader of the three new fields has been named. Requirement 2.1 asks for "unambiguous names for every added field" while the spec's own out-of-scope forbids "expanding the projection... without a named consumer." Name the consumer or cut the field.

---

hollomancer:
Closing as agreed — folded into `specs/phase-iii-source-materialization/tasks.md` in #692 rather than kept as a separate spec directory and registry entry.

The folded version keeps the substance and corrects the framing: the narrow projection is a documented memory tradeoff (`usaspending_award_archive.py:36-37`), not an accident, and the `phase_ii.py:143` regex prevents the failure mode — so the tasks argue from *deleting that compensating fallback*, which is now task 5 and the thing that justifies the work. Task 1 is restated as a notebook audit, since the projection has never been checked against a real archive header. Task 5 of the original (downstream reader migration) is cut, per the unnamed-consumer point.

---

PR #688, opened 2026-08-31
Stub semantic fingerprints for transition materializations

hollomancer:
## Description

Adds a pipelines-tier maintenance spec for invalidating transition artifacts when transformation semantics change. This is a planning stub only; implementation remains unchecked in `tasks.md`.

## Type of Change

- [x] Documentation update
- [ ] Bug fix

## Versioning Impact

- [x] No release impact

## Testing

- [x] `make docs-check`
- [x] `git diff --check`

## Scope

The future implementation will bind Phase II, Phase III, pair, and survival outputs to named transformation versions, normalized configuration, and verified upstream hashes. It does not authorize a live materialization or evidence-tier promotion.

### DIFF
--- a/specs/materialization-semantic-fingerprints/design.md
+++ b/specs/materialization-semantic-fingerprints/design.md
@@ -1,0 +1,52 @@
+# Materialization Semantic Fingerprints — Design
+
+## Fingerprint payload
+
+Each stage emits a canonical JSON payload with:
+
+```text
+stage_name
+transformation_contract
+transformation_version
+normalized_config
+upstream[{name, output_sha256, semantic_fingerprint}]
+direct_sources[{name, sha256}]
+```
+
+The semantic fingerprint is SHA-256 over canonical UTF-8 JSON with sorted keys and compact
+separators. Stage code owns a named version constant; a Git SHA and package version can be emitted
+as diagnostics but do not replace that contract.
+
+## Validation flow
+
+1. Resolve and validate every upstream manifest.
+2. Build the expected canonical payload from current configuration and contract versions.
+3. Compare its SHA-256 with the stored fingerprint before accepting a cached output.
+4. Verify the output bytes against the stored output SHA-256.
+5. If any check fails, report the mismatched payload component and require rematerialization.
+
+The validator should be a small shared pipeline utility so stages do not reimplement canonical
+JSON or mismatch reporting. The first consumers are `validated_phase_ii_awards`,
+`validated_phase_iii_contracts`, pair construction, and the survival frame.
+
+## Version-change discipline
+
+Focused tests should pin representative transformation behavior to its declared version. A test
+must fail when a fixture output changes without the expected version bump. This is narrower and
+more reviewable than hashing implementation source files, which would invalidate on comments and
+miss behavior imported from elsewhere.
+
+## Migration and failure behavior
+
+Legacy manifests without a semantic fingerprint are stale by definition. Migration is a local
+rematerialization, not an in-place manifest rewrite. Empty artifacts still publish an output (or
+an explicitly hashed empty representation) and a complete manifest so zero-row results cannot
+escape lineage checks.
+
+## Testing strategy
+
+- Canonical-payload stability across key order and non-semantic path changes.
+- Invalidation on contract version, normalized config, source hash, or upstream fingerprint change.
+- Legacy/missing manifest rejection.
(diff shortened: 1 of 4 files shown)


hollomancer:
## Verdict: **mis-diagnosed** — but there is a real 3-line fix inside worth landing immediately

**The stated background is a cache-validity failure. These assets have no cache.** `validated_phase_iii_contracts` (`assets/phase_transition/phase_iii.py:139-210`) recomputes unconditionally from `contracts_path` on every materialization; so do `phase_ii.py` and `pairs.py`. Requirements 2.x and 3.1/3.2 solve a problem that does not exist here.

**The one real defect is Requirement 3.3's subject, and it is three lines:**

```python
# phase_iii.py:155-156
if not phase_iii.empty:
    phase_iii.to_parquet(output_path, index=False)
```

Mirrored at `pairs.py:297-298` and `pairs.py:364-365`. When the frame is empty the **previous parquet is left on disk** while `checks.json` is overwritten unconditionally (`phase_iii.py:190-191`) with `total_rows: 0`. That is precisely the "valid-looking zero-row checks artifact" the spec's own background describes — caused by the empty-write skip, not by a missing semantic fingerprint.

**Recommend landing that as a standalone fix now** (always write the parquet, including empty). Highest value-to-effort ratio in the entire six-PR set, and it needs no spec.

**On the framework half:** `assets/phase_transition/sbir_gov_source.py:290-449` already implements a manifest with source SHA, output SHA, ordered-column SHA, and a `fingerprint` contract block plus a validator that rejects mismatches. `design.md:28` proposes "a small shared pipeline utility" for canonical JSON and mismatch reporting — that is a second, unnamed implementation of an existing concept, which is the thing CLAUDE.md's reuse rule exists to prevent.

**Cut tasks 1, 2, and 4** — a semantic-fingerprint framework, transformation-contract versioning, and cross-stage lineage propagation for assets that do not cache. Keep only the empty-output half of task 3.

Also note Requirement 3.3 here and Requirement 3.2 in #689 both govern empty/incomplete rows in the same `pairs`/`survival` outputs.

---

hollomancer:
Scope-guard confirmation: **do not build the fingerprint framework.** These assets recompute every run; there is no cache. `sbir_gov_source.py` already has a fingerprint contract. Folding a second helper into `phase-iii-source-materialization` would violate the reuse rule.

The one real defect remains the empty-write skip at `phase_iii.py:155-156` (mirrored in `pairs.py`). Land that as a standalone bugfix: always write the parquet, including empty, matching `_write_parquet_atomic` in `phase_ii.py`. Close this spec PR.

hollomancer:
The 3-line kernel from my review above is implemented on branch [`claude/review-open-prs-3y08y9`](https://github.com/hollomancer/sbir-analytics/compare/main...claude/review-open-prs-3y08y9) (commit `b6f52cd6`), since I couldn't push to this branch.

`phase_iii.py:155`, `pairs.py:297`, and `pairs.py:364` now write unconditionally. I confirmed the empty frames always carry their declared schema (`PHASE_III_COLUMNS` / `PAIR_COLUMNS` / `SURVIVAL_COLUMNS`), so an empty write preserves columns rather than producing a schema-less file.

New regression file `tests/unit/phase_transition/test_empty_output_write.py` covers both directions — a stale parquet being overwritten by a zero-row run, and a zero-row run creating the file where none existed. I verified both tests **fail against the previous behavior and pass after the change** rather than assuming it. 557 tests pass across the affected suites; `make lint` and `make lint-boundaries` are clean.

That leaves the framework half of this spec (tasks 1, 2, 4) as the open question — my recommendation above is still to cut it, given these assets have no cache and `sbir_gov_source.py:290-449` already implements the manifest/fingerprint concept.

---

hollomancer:
Closing as agreed — superseded by #692.

The real defect from this spec is **fixed and shipped** there: all three sites write their parquet unconditionally, with `tests/unit/phase_transition/test_empty_output_write.py` covering both the stale-overwrite and first-write cases, verified failing-then-passing. It's recorded as the one completed task in `specs/phase-iii-source-materialization/tasks.md`.

The framework half is cut, for the two reasons in my review: these assets have no cache (they recompute unconditionally every materialization), and `sbir_gov_source.py:290-449` already implements source SHA, output SHA, ordered-column SHA, and a rejecting validator. The one surviving follow-up is to extend *that* manifest to the Phase II/III assets rather than build a second one.

---

PR #689, opened 2026-08-31
Stub shared phase-transition report grain repair

hollomancer:
## Description

Adds a pipelines-tier maintenance spec for removing mixed pair- and award-grain statistics from the shared transition report. This is a planning stub only; implementation remains unchecked in `tasks.md`.

## Type of Change

- [x] Documentation update
- [ ] Bug fix

## Versioning Impact

- [x] No release impact

## Testing

- [x] `make docs-check`
- [x] `git diff --check`

## Scope

The future implementation will use a one-row-per-Phase-II follow-up frame for headline incidence and event-conditional latency, while retaining candidate-pair multiplicity only as a diagnostic. Matching policy and Kaplan–Meier estimation are out of scope.

### DIFF
--- a/specs/phase-transition-report-grain/design.md
+++ b/specs/phase-transition-report-grain/design.md
@@ -1,0 +1,51 @@
+# Phase-Transition Report Grain — Design
+
+## Current flow
+
+`phase_transition_analysis` reads both `pairs` and `survival`. Percentiles, histograms, and agency
+median latency query all pair rows; transition rates query survival rows. This makes the report a
+mixture of pair-weighted and award-weighted summaries.
+
+## Proposed reporting contract
+
+The survival/follow-up artifact becomes the only headline analysis table:
+
+- grain: one row per canonical Phase II award;
+- event fields: observed flag, selected first-event contract/action, signed time from the declared
+  origin;
+- censor fields: cut date and follow-up time when no event is observed;
+- strata: Phase II agency and cohort attributes copied from the same award row.
+
+Headline incidence, event-conditional latency, agency summaries, and cohort summaries query this
+table. The pairs table remains an input to construction and an output for link diagnostics only.
+
+## Multiplicity audit
+
+Before report generation, compute and emit:
+
+1. candidate Phase III contracts per Phase II award;
+2. Phase II assignments per selected Phase III contract;
+3. counts with multiplicity greater than one and each maximum;
+4. row totals at pair, Phase II award, and distinct selected-contract grain.
+
+These diagnostics explain linkage ambiguity without changing the chosen headline unit.
+
+## Failure behavior
+
+Duplicate Phase II identifiers, incomplete event/censor fields, or inconsistent observed flags are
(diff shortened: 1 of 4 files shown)


hollomancer:
## Verdict: real defect, over-built by an order of magnitude, and mis-tiered — recommend a plain bug PR instead of a spec

**The defect is real, exactly as described.** `sbir_etl/reporting/phase_transition_analysis.py`:
- `:71` and `:79` — latency percentiles and month-bin histogram both `FROM pairs`
- `:96` — transition rate `FROM survival s`
- `:104-107` — agency table computes `phase_ii_total`/`matched` from `survival`, then `LEFT JOIN pairs pa` for `median_latency_days`

Mixed denominators under one transition vocabulary.

**But look at what's being fixed.** The file is 254 lines, has **one** importer (`assets/jobs/phase_transition_archive.py:18`) and one test file. The fix is repointing three queries at `survival` — roughly 20 lines. The spec proposes a versioned report schema, blocking schema errors, a multiplicity audit, and downstream-reader migration. That is production hardening on a script with a single caller.

**Tier is a silent promotion.** The target module declares `EPISTEMIC_TIER = "exploratory"` at line 33 with a non-citable banner at lines 3-4. This spec declares `**Target epistemic tier:** pipelines` while its own out-of-scope section says *"Promoting the report beyond its current non-citable tier."* Exploratory → pipelines **is** a tier change, and CLAUDE.md is explicit that promotion is explicit work. Either state it as promotion with justification, or build the fix at exploratory tier — which the ~20-line size of the real fix argues for.

**Recommendation:** close this as a spec and open a normal bug PR against `phase_transition_analysis.py:60-111`. **Cut tasks 3, 4, and 5** (multiplicity diagnostics, report-key versioning, downstream-reader migration, documentation reconciliation).

One cosmetic note the registry guard won't catch — it checks set equality and duplicates, not ordering: this PR inserts `phase-transition-report-grain` between `phase-iii-source-materialization` and `phase-iii-hand-label-validation` in `specs/status.md`. The list is alphabetical and `phase-transition-*` sorts after all `phase-iii-*` entries and before `phase3-*`. Move it down two entries if the spec survives.

---

hollomancer:
Closing as agreed — the defect is **fixed** in #692 as a plain bug fix, with no spec.

All three queries in `sbir_etl/reporting/phase_transition_analysis.py` now read `survival` at one row per Phase II award, using `time_days` over observed events. The agency table's `LEFT JOIN pairs` is gone too — it fanned each agency out to its pair rows, so the median was pair-weighted while the rate beside it was award-weighted.

Worth noting the pre-existing test was itself pinning the bug: it asserted `latency_distribution.n == 2` from two `pairs` rows that shared a single award. It now asserts `1`, and a new `test_latency_and_rate_share_the_award_grain` pins the invariant with a three-contract award. Both verified failing against the old queries.

No tier change: the module keeps `EPISTEMIC_TIER = "exploratory"`. The proposed pipelines-tier report schema would have been a silent promotion of a 254-line module with one importer, and the fix turned out to be ~20 lines of SQL.

---

PR #690, opened 2026-08-31
Stub EDGAR type-specific event-date preservation

hollomancer:
## Description

Adds a pipelines-tier maintenance spec for preserving EDGAR mention type and filing date at event grain. This is a planning stub only; implementation remains unchecked in `tasks.md`.

## Type of Change

- [x] Documentation update
- [ ] Bug fix

## Versioning Impact

- [x] No release impact

## Testing

- [x] `make docs-check`
- [x] `git diff --check`

## Scope

The future implementation will persist accession-level mention events and derive latest dates within each type. It will not validate completed acquisitions, change classifiers, add a search backend, or estimate M&A rates.

PR #691, opened 2026-08-31
Stub Form D amendment-chain deduplication

hollomancer:
## Description

Adds a pipelines-tier maintenance spec for preventing Form D amendment chains from being summed as independent raises. This is a planning stub only; implementation remains unchecked in `tasks.md`.

## Type of Change

- [x] Documentation update
- [ ] Bug fix

## Versioning Impact

- [x] No release impact

## Testing

- [x] `make docs-check`
- [x] `git diff --check`

## Scope

The future implementation will preserve accession-grain lineage, resolve only auditable offering chains, select one as-of amount per series, and exclude unresolved amendments from default dollar totals. Company matching, causal attribution, and undisclosed-capital estimation are out of scope.

### DIFF
--- a/specs/form-d-amendment-chain-deduplication/design.md
+++ b/specs/form-d-amendment-chain-deduplication/design.md
@@ -1,0 +1,53 @@
+# Form D Amendment-Chain Deduplication — Design
+
+## Data flow
+
+```text
+Form D filings → accession-grain normalized rows → versioned chain resolver
+                                               ↘ resolution audit
+                    resolved series → one as-of representative → monetary aggregate
+                    unresolved rows ───────────────────────────→ excluded/audited
+```
+
+The accession-grain table remains canonical. Chain and firm aggregates are derived outputs with
+their own grain and policy version.
+
+## Chain identity
+
+Implementation begins with an audit of SEC XML fields and known original/amendment sequences. Use
+an explicit predecessor or offering identifier if the source exposes one. Any fallback key must be
+documented, deterministic, and tested against collisions; plausible-but-ambiguous candidates stay
+unresolved. Issuer name alone is never a chain key.
+
+Candidate fallback components may include issuer CIK, date of first sale, exemption, security
+types, and stable offering attributes, but the implementation may only adopt them after the audit
+shows the collision and missingness behavior. The named chain-policy version makes later rule
+changes explicit.
+
+## Representative selection
+
+Within a resolved series and declared as-of cut:
+
+1. order filings by valid filing date and accession tie-break;
+2. select the latest valid filing representing cumulative progress;
+3. retain its reported `total_amount_sold` as the series amount;
+4. retain earlier accessions as lineage, not additive dollars.
+
+Conflicting same-day records, decreasing/restated values, and orphan amendments are surfaced in
+checks. A decrease is not silently rewritten; the latest disclosed amount and audit flag are both
+preserved when the chain itself is unambiguous.
+
+## Consumer migration
+
+`agency_private_capital` inputs, capital-event summaries, and bootstrap leverage code must consume
+resolved series amounts rather than summing offering rows. Participation outputs may continue to
+use accession-grain filings if their names state that grain. Legacy dollar artifacts are not
+relabelled as chain-resolved.
+
+## Testing strategy
+
+- Original plus two cumulative amendments (5, 8, 10) produces 10, not 23.
+- Two independent resolved offerings produce the sum of two representatives.
+- Orphan and ambiguous amendments are excluded with reasons.
+- Duplicate-accession conflict fails closed.
+- As-of cut selects the latest filing available by that date.
--- a/specs/form-d-amendment-chain-deduplication/requirements.md
+++ b/specs/form-d-amendment-chain-deduplication/requirements.md
@@ -1,0 +1,86 @@
+# Form D Amendment-Chain Deduplication — Requirements
+
+> **Lifecycle status:** Maintenance
+> **Spec-file progress:** Not yet started
+> Anchors inventory questions **F1 and F3** in
+> [docs/research-questions.md](../../docs/research-questions.md).
+
+**Target epistemic tier:** `pipelines`
+
+**Research question anchor:** F1 Form D fundraising profile and F3 disclosed Form D leverage
+**Answers for:** pipeline engineers maintaining SEC private-capital inputs
+**RQ complexity tier:** Descriptive / Inferential downstream; this spec normalizes filings only
+
+---
+
+## Done when
+
+Form D monetary aggregates use at most one as-of amount from each auditable offering/amendment
+chain, retain every source accession for lineage, and exclude unresolved amendment chains from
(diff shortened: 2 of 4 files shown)


hollomancer:
## Verdict: **grounded — highest live risk of the six.** Coordinate before landing

**The defect is confirmed and it silently inflates a dollar figure.** `packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/form_d_inputs.py:129`:

```python
total_sold = _sum_float(o.get("total_amount_sold") for o in kept_offerings)
```

`kept_offerings` (lines 115-119) filters only on `industry_group`. Nothing filters `is_amendment`. This becomes `total_form_d_raised` at line 163. Since Form D amendments restate *cumulative* offering amounts, an original plus two amendments of 5/8/10 sums to **23 instead of 10**.

The fields needed to fix it are already present and simply ignored: `is_amendment` at `sbir_etl/models/sec_edgar.py:195`, carried through `sbir_etl/capital_events/sources/form_d.py:85`; `accession_number` at `form_d.py:79`. The same raw-sum shape recurs at `scripts/data/bootstrap_form_d_leverage_ci.py:129` and `scripts/data/nano_form_d_temporal.py:128` — worth fixing together.

### Live collision — coordinate before this lands

`form_d_inputs.py` is the direct input to the **Active** `agency-private-capital-comparison` spec (`specs/status.md:33-36`), whose Phase 2 is gated on *"a reproducible Form D control-universe producer."* There is a live branch `origin/agent/form-d-control-universe` (**#582**) working the same file. Two efforts will collide here. Worth settling ownership before either merges — and note that #582 has its own open finding about `load_form_d_control_universe` in this same module.

### Two adjustments

**Task 1 belongs in a notebook first.** `design.md:17-25` openly states the SEC amendment-link field is unknown. Per CLAUDE.md's notebook-first rule, an unresolved key/schema question isn't ready to be a frozen pipelines task list — resolve the audit, then write the tasks.

**Cut task 6** ("Document legacy-artifact boundaries and run repository guards") — running `make docs-check` is not a task.

**Recommend merging with #690** into a single SEC-source-fidelity spec; both touch `sbir_etl/enrichers/sec_edgar/` and `sbir_etl/capital_events/` and feed the same downstream.

---

hollomancer:
Correction after an independent scope-guard pass on this batch.

**The defect (raw-sum of amendment restatements) still holds.** `is_amendment` is already on the model and is ignored in `form_d_inputs.py` dollar totals. Same shape in `bootstrap_form_d_leverage_ci.py`, which is the Computable F3 implementation.

**Two claims in the previous comment need tightening:**

1. **This stub does not currently rewrite `form_d_inputs.py`.** There is no file-level merge collision with #582 on that module. The collision is *ownership*: #582's DERA producer already requires `ISAMENDMENT` / `PREVIOUSACCESSIONNUMBER` on `OFFERING.tsv`, while this spec's task 1 treats the amendment-link field as unknown. A later implementation that rewrites `form_d_inputs.py` would fight Phase 2; this PR as a stub does not.

2. **Do not merge this with #690 into an "SEC-source-fidelity" umbrella.** Mentions ≠ Form D offerings; consumers and owning specs differ. An umbrella spec is more process, not less.

**Recommended path:** close this spec PR. Notebook first on `form_d_details.jsonl` — how much does summing D/A restatements move the F3 1.82× ratios? If material, patch `bootstrap_form_d_leverage_ci.py` (the Computable number) without waiting for #582. If a chain resolver is still wanted, add it as a task on the Active `agency-private-capital-comparison` spec / #582 using DERA `PREVIOUSACCESSIONNUMBER`, with unresolved D/A excluded from default dollar totals.

hollomancer:
A **partial** fix for the amendment double-count is on branch [`claude/review-open-prs-3y08y9`](https://github.com/hollomancer/sbir-analytics/compare/main...claude/review-open-prs-3y08y9) (commit `b6f52cd6`). Please read the caveat before treating this as done.

**What I could not do, and why.** The correct fix is to collapse each amendment chain and take the final restatement — the 5/8/10 example should yield **10**. That needs the SEC file number (the `021-…` offering identifier), which is the actual chain key. I searched: `file_number` appears **nowhere** in the codebase, and `FormDFiling` (`sbir_etl/models/sec_edgar.py:195`) carries only the `is_amendment` boolean and a per-filing `accession_number`, neither of which groups filings into chains. So `design.md:17-25` is right that the chain key is unknown, and the exact fix stays blocked on your task 1.

**What I implemented instead** — conservative, and labeled as such in the code comment: sum non-amendment offerings only, falling back to the *largest* restatement when a chain reaches us as amendments alone (so a truncated window doesn't zero out real money). `offering_count` gets the same treatment, since it carried the identical double-count. `_offering_from_flat_record` now propagates `is_amendment` so flat records aren't silently treated as originals.

On the 5/8/10 example this reports **5**, not 23 and not the correct 10. It trades a silent overcount for a documented undercount. For a control-comparison denominator I judged the conservative direction the safer one to sit on while the audit is outstanding, but **it is a lower bound, not a total**, and the code comment says so. If you'd rather it stayed visibly broken until the chain key lands, that's a reasonable call — say the word and I'll revert it.

Two regression tests cover both branches. `make lint` and `make lint-boundaries` clean.

Note this also touches `form_d_inputs.py`, the same file as **#582**'s staging guard (also on that branch) — which is the collision I flagged above. The two changes are independent and coexist there, but whichever PR lands second will need to account for the other.

---

hollomancer:
Closing as agreed — merged with #690 into `specs/sec-source-fidelity/` in #692, which also carries the interim code fix.

To be clear about what is and isn't done: the double-count no longer inflates totals, but the fix is **partial and labeled as such in the code**. It reports **5** on a 5/8/10 chain — not 23, and not the correct 10. Exact collapse needs the SEC file number, and your `design.md:17-25` was right that it's unknown: I searched, and `file_number` appears nowhere in the codebase, with `FormDFiling` carrying only `is_amendment` and a per-filing `accession_number`. That gate is now recorded in the merged spec, with the chain-key audit as its task 1 and the exact collapse as task 5, explicitly blocked on it.

`offering_count` got the same treatment — it carried the identical double-count — and `_offering_from_flat_record` now propagates `is_amendment` so flat records aren't silently treated as originals.

The collision I flagged still needs your call: `sec-source-fidelity` and the Active `agency-private-capital-comparison` spec both own `form_d_inputs.py`, and PR #582 is working the same file. The merged spec says to settle ownership before starting.

---

hollomancer:
Correction to my closing comment above, for the record.

I said this was merged into `specs/sec-source-fidelity/` alongside #690. That merge was a mistake and is undone in #693. The two proposals share a topic — SEC source fidelity — but not a package or an owner: your Form D work changes `form_d_inputs.py`, which lives inside the Active `agency-private-capital-comparison` spec's own package and is imported only by that spec's assets and tests.

This work now lives there as tasks **F.1–F.3**: the shipped staging refusal, the blocking chain-key audit, and the exact collapse that depends on it. Same content, correct home — and it resolves the ownership collision I flagged rather than relocating it.

---

PR #699, opened 2026-09-06
Ship fail-closed M&A discovery and the confirmatory recall floor

hollomancer:
## Description

Replaces #662 and #663. Those PRs were the search-tool and extractor-eval slices; this branch rebases them onto current `main` (including #698) and adds the rest of the selected `ma-discovery-integration` slice.

**Shipped (spec tasks 1–5)**

- Fail-closed search default (`none`); `mock` is explicit; live backends need a key.
- One confirmed row per `(company, acquirer)`; YAML timeout/rate-limit honored.
- LLM pair-name match via `RECIPIENT_V1`; citation is the search result URL.
- C3 collision join and the design confidence table.
- Frozen-snippet backend and sample-run / review-queue CLI (`--skip-pairs`, `--stop-when dated_confirm`, `--strict-recall`, `--fail-on-gate`).

**Measured, not promoted (tasks 6–8)**

The `ma-discovery-recall` study stays `evidence_status: exploratory`. Materialization stays closed. `docs/research-questions.md` is not edited.

Pilot (pairs 1–500): 200-pair recall 9<10; 500-pair enlarge was post-hoc (evidence-auditor 2026-09-02 BLOCK). Labels: 14 true / 1 false / 1 ambiguous.

Held-out confirmatory (pairs 501–1500, hashed `confirmatory-design.md` before capture). All three gates met:

- Strict medium/high C3 insert-or-promote: **13** (floor ≥10).
- Precision: 20 medium rows labeled **19 true / 0 false / 1 ambiguous**. FP = 0/19 (cap 0.25). Ambiguous is Synlogic/Mirna.
- Cost ~$0.0186/pair (Brave $5 + LLM ~$13.64, under $0.10).
- 140 discovered rows (84 low, 52 medium, 4 high). C3: 138 inserts, 2 promotes.

Promotion still needs evidence-auditor. Do not cite these numbers.

Brave snippets and raw LLM JSONL stay gitignored. Tracked hash manifests: `studies/ma-discovery-recall/run-manifest.json`, `confirmatory-run-manifest.json`, `confirmatory-labels.jsonl`.

Keyword remains the orchestrator default. LLM confirmation is opt-in on the sample-run CLI.

## Type of Change

- [x] Documentation update

## Versioning Impact

## Testing

```
uv run pytest tests/unit/enrichers/ma_discovery/ tests/unit/scripts/test_run_ma_discovery_sample.py tests/unit/config/test_schemas.py -q
uv run python scripts/ci/validate_study_manifests.py
uv run python scripts/ci/check_tier_boundaries.py
```

Focused tests pass. Study hashes match. No live Brave or LLM calls in CI.

## Checklist


tests/unit/enrichers/test_openai_client.py
```diff
@@ -26,6 +26,49 @@ def test_chat_success(self):
...
+        client = OpenAIClient(
+            api_key="test-key",
+            chat_url="https://api.x.ai/v1/chat/completions",
+        )
+        client._client = mock_http
+        assert client.chat("system", "user") == "ok"
+        assert mock_http.request.call_args.args[1] == "https://api.x.ai/v1/chat/completions"
+        headers = mock_http.request.call_args.kwargs["headers"]
+        assert headers["Authorization"] == "Bearer test-key"
+
```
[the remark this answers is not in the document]
hollomancer: False positive from the same redaction. The assertion is `"Bearer test-key"` against `OpenAIClient(api_key="test-key")`.

sbir_etl/enrichers/openai_client.py
```diff
@@ -82,10 +86,12 @@ def __exit__(self, *exc: object) -> None:
         self.close()
 
     def _headers(self) -> dict[str, str]:
-        return {
+        headers = {
             "Authorization": f"Bearer {self._api_key}",
             "Content-Type": "application/json",
         }
+        headers.update(self._extra_headers)
+        return headers
```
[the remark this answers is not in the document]
hollomancer: False positive from GitHub redacting the diff. The source is `f"Bearer {self._api_key}"`; masking belongs in logs, not this header.

hollomancer:
## Update 2026-09-08: held-out capture ran, missed, and was BLOCKED

8 commits pushed. The held-out 1501-2500 capture executed and did not pass; this adds the gates that should have caught how it failed, and corrects the study record.

**The reported number was wrong.** The capture process scored **528 of the 1,000 cut pairs** and reported `strict_medium_high_n: 2` under a field name reading as cut-level recall. The resume path drops already-recorded queries, so the 472 frozen pairs never reached the extractor — and those were the entire hit-bearing half of the cut. The replay over all 1,000 pairs, from a freeze hashed and committed beforehand (`held-out-run-manifest.json`), gives the real figure: **strict recall 4 against a floor of 10**.

evidence-auditor returned BLOCK on four independent grounds, all recorded in `study.yaml`. None is cleared by this branch.

### New in these commits

- **Coverage gate** — fails when scored pairs differ from the protocol cut. Its absence is what let a subset count present as recall.
- **Network faults counted, not scored as evidence.** A failed search recorded an empty hit; an LLM timeout recorded an unconfirmed verdict. Both were indistinguishable from a true negative and counted nowhere.
- **Asymmetric fault rule.** Faults only understate recall, so a pass carrying them stands and a miss carrying them is not measured. An earlier commit here voided any run with a fault, which was wrong for the pass case.
- **Gate 3 fails closed.** `cost_per_pair_usd` was hardcoded `None`, so a run could report `accepted` without gate 3 being evaluated — `intended_rank: validated` was unreachable while looking reachable.
- **Retries actually retry.** Faulted rows froze their own query, so the prescribed "rerun the faulted pairs" would have skipped exactly those pairs.
- **`code_version`** (HEAD + dirty flag) now recorded in summaries. The overnight capture ran from a dirty tree and its code state is unrecoverable.

### Corrections to earlier commits on this branch

`1fc7d936` wrote a false statement into the frozen protocol note — that no Brave or LLM call had touched pairs 1501+ when the capture path changed. 948 of 1,000 pairs had been captured. Corrected in `bb7c2ac3` with the superseded hash recorded in the note rather than silently replaced.

The same commit asserted that faults can never manufacture a confirmation. That is false for the gated quantity: `strict_medium_high_n` is not monotone in the discovered set, because `apply_c3` matches on company key alone and credits promotions to the existing row pair key. A counterexample executed against the real `apply_c3` shows strict recall going 0 -> 1 under a fault. Exposure here is 3 company keys over 6 pairs. The proof is replaced with the bounded statement.

### Verification

616 tests pass; all 11 `frozen_artifacts` pins match; `make lint-boundaries` clean. New gates mutation-checked — removing either fails its test. One test was initially vacuous (asserted `main() == 1`, which passes with the fault gate disabled because a recall miss also exits 1); fixed to assert the fault message and re-verified.

### Power finding

The floor of 10 is not reachable on the remaining population. 362 strict-eligible pairs remain, the eligible fraction decays down file order (68 -> 50 -> 31 -> 29 percent), and at observed rates the entire remainder yields ~5 to 9 expected. A further cut under this design is preregistered to fail. Recorded as a blocker.

Related: #704, #702, #703.

scripts/data/run_ma_discovery_sample.py
```diff
@@ -0,0 +1,1038 @@
...
+    name = "counting_llm"
+
+    def __init__(self, inner: SnippetExtractor) -> None:
+        self.inner = inner
+        self.timeout_n = 0
+
+    def extract(self, item: Any) -> ExtractionVerdict:
+        try:
+            return self.inner.extract(item)
+        except (TimeoutError, httpx.HTTPError, OSError) as exc:
```
hollomancer, line 257: **[bug]** `_CountingExtractor` only increments `timeout_n` when `extract()` raises `TimeoutError`/`httpx.HTTPError`/`OSError`. The live path never does that. `OpenAIClient._request` swallows `httpx.TimeoutException` / `RemoteProtocolError` after retries and returns `None` (`sbir_etl/enrichers/openai_client.py:128`). `RecordingLlmExtractor.extract` then treats a missing payload as an ordinary unconfirmed verdict and does not raise (`sbir_etl/enrichers/ma_discovery/extractor.py:311`). The pair is scored, `llm_timeout_n` stays 0, `fully_measured` stays true, and a recall miss is reported as a clean miss. That is the exact fail-closed hole this counter was added to close; the unit test only covers a stub that raises `ReadTimeout`.

**Suggestion:** Treat a `None`/empty chat payload as a transport fault in `RecordingLlmExtractor` (raise or return a sentinel that `_CountingExtractor` counts). Alternatively, make `OpenAIClient.chat` raise on timeout/protocol failure instead of returning `None`, and add a test that drives `OpenAIClient` → `RecordingLlmExtractor` → `_CountingExtractor` and asserts `timeout_n == 1` with no freeze row.

sbir_etl/enrichers/ma_discovery/collision.py
```diff
@@ -0,0 +1,175 @@
...
+        company_key = name_key(str(hit.get("company_name") or ""))
+        if not company_key:
+            continue
+        hit_date = hit.get("event_date") or hit.get("date")
+        match_idx: int | None = None
+        for idx, row in enumerate(rows):
+            if name_key(str(row.get("company_name") or "")) != company_key:
+                continue
+            existing_date = row.get("event_date")
+            if parse_event_date(existing_date) is None:
```
hollomancer, line 108: **[bug]** C3’s primary key is `(company_name, event_date ± 30 days)`, with company-only matching only when the existing date is missing. The loop checks the dateless fallback first and `break`s, so a dated discovery binds to the first dateless row for that company even when a later existing row is inside the ±30-day window. Design calls empty `event_date` common for EFTS-only rows, so this is the likely join, not an edge case. The hit can promote the wrong event and never collide with the dated one.

**Suggestion:** Prefer a ±30-day dated match when one exists; use company-only matching only if no dated collision is found. Add a fixture with `[dateless EFTS row, dated row]` on the same company and a discovery dated to the second row.

scripts/data/run_ma_discovery_sample.py
```diff
@@ -0,0 +1,1038 @@
...
+    summary_path = args.output_dir / "sample_run_summary.json"
+    with discovered_path.open("w", encoding="utf-8") as handle:
+        for row in discovered:
+            handle.write(json.dumps(row, sort_keys=True) + "\n")
+    with queue_path.open("w", encoding="utf-8") as handle:
+        for row in queue:
+            handle.write(json.dumps(row, sort_keys=True) + "\n")
+
+    medium_high = [row for row in discovered if row.get("confidence") in {"medium", "high"}]
+    strict_n = strict_medium_high_n(events, collision.inserted + collision.promoted)
```
hollomancer: **[bug]** The confirmatory estimand counts pairs the extractor confirms at medium or high **and** that C3 inserts or promotes. `strict_medium_high_n` instead inspects post-C3 confidence on `inserted + promoted`. C3 always bumps a non-Form-D low existing row to medium on any confirm (`collision.py:145`), including an undated/low discovery. `dates_collide` matches when both dates are missing, so a low confirm against a dateless low EFTS row becomes a strict-recall credit for the **existing** pair key (which may not even be the discovered acquirer). Inserts of low rows are filtered; this promotion path is not. There is no test for low existing + low discovery.

**Suggestion:** Count a promotion toward strict recall only when the **discovery** row was medium/high (and, if you keep company-only C3, report the discovered pair, not the existing one). Keep the C3 join behavior for the event set; do not let it define the gated estimand.

sbir_etl/enrichers/openai_client.py
```diff
@@ -101,16 +107,32 @@ def _request(
...
+                        time.sleep(wait)
+                        continue
+                    logger.warning(f"OpenAI API {type(exc).__name__} after {MAX_RETRIES} retries")
+                    return None
             finally:
                 self._semaphore.release()
 
+            if resp.status_code in {401, 402, 403}:
+                logger.warning(f"OpenAI API error: {resp.status_code} (no retry)")
+                raise RuntimeError(f"OpenAI API error: {resp.status_code}")
```
hollomancer: **[bug]** 401/402/403 now raise `RuntimeError` with no retry. That is reasonable for M&A capture (do not write an empty freeze row on OpenRouter 402), but `OpenAIClient.chat` previously returned `None` after retries and other callers still assume that. `sbir_etl/reporting/weekly/llm.py` `generate_weekly_synopsis` calls `_openai_chat` with no try/except; a 401 now crashes the weekly report instead of omitting the synopsis. `_CountingExtractor` also does not catch `RuntimeError`, so a 402 aborts the whole sample run rather than counting a fault and checkpointing.

**Suggestion:** Raise a dedicated exception (or keep returning `None` for 401/403 and fail-fast only for 402). Catch that type in the M&A capture path as a run-aborting billing/auth fault. Restore None-on-auth-failure for weekly/procurement callers, or wrap `_openai_chat` so synopsis generation degrades as before.

sbir_etl/enrichers/ma_discovery/orchestrator.py
```diff
@@ -22,39 +23,96 @@
...
+                    company=company,
+                    acquirer=acquirer,
+                    snippet=str(snippet),
+                    source_url=source,
                 )
-                break  # one hit per (company, acquirer) is enough
+            )
+            if not verdict.confirmed:
+                continue
+            confidence = assign_confidence(verdict, source_count=len(source_urls))
```
hollomancer: **[suggestion]** `assign_confidence(..., source_count=len(source_urls))` counts every URL seen so far in the result list, including unconfirmed hits, and does not dedupe. Design high is date + value + **≥2 distinct confirming sources**. One confirming snippet with date+value after an unrelated earlier URL becomes `high`. High rows are dropped from the medium-only review queue, so this also hides candidates from the precision sample.

**Suggestion:** Count unique URLs on **confirmed** hits only. A single confirming source with date+value should stay `medium`.

sbir_etl/enrichers/ma_discovery/orchestrator.py
```diff
@@ -22,39 +23,96 @@
...
+            }
+            if stop_when == "first_confirm" or confidence in {"medium", "high"}:
+                seen.add(key)
+                verified.append(event)
+                undated = None
+                break
+            if undated is None:
+                undated = event
+        else:
+            if undated is not None:
```
hollomancer: **[suggestion]** Under `dated_confirm`, an undated confirm on the first query is committed and the pair is added to `seen`, so later query templates for that pair never run. The stop rule is “until a dated confirm or hits are exhausted.” With `queries_per_pair=1` this is fine; with the four `generate_queries` templates it permanently keeps a low row that a later template might have dated.

**Suggestion:** For `dated_confirm`, only add the pair to `seen` on a medium/high hit. Carry the first undated confirm across remaining query rows and emit it only after every template is exhausted.

specs/ma-discovery-integration/design.md
```diff
@@ -99,7 +99,7 @@ These are choices the design intentionally does *not* pin, because they're indep
 
 | Decision | Default I'd lead with | Why deferred |
 |---|---|---|
-| Search backend (Tavily / Brave / Bing / Serper) | Tavily — purpose-built for snippet-focused agentic search; pluggable interface stays put | Pricing changes; better resolved with a real cost estimate |
+| Search backend (Tavily / Brave / Bing / Serper) | **Tavily** is the default production client; Brave is a second same-shape backend. Runtime default stays mock until a key is set. Comparison: [`search-backends.md`](search-backends.md) (exploratory, non-citable — not a recall benchmark). | Comparison note landed; revisit only if the sample run shows snippet quality is the bottleneck |
```
hollomancer: **[suggestion]** The open-decision table still says “Runtime default stays mock until a key is set,” and step 3 claims “Tavily default, optional Brave, mock without a key.” The implemented contract — and `search-backends.md` — is `none` (fail-closed), with `mock` as explicit opt-in. That contradiction will cause someone to set a live backend in YAML and assume missing keys fall back to the Physical Optics fixture.

**Suggestion:** Change both sentences to match `MADiscoveryConfig.search_backend` default `"none"` and `build_search_tool`’s fail-closed error.

hollomancer:

### Status: ✅ Ready for Review

- **Draft:** No
- **Age:** 3 days
- **Size:** +5,801/-55 across 47 files — **largest open PR**

Ships fail-closed M&A discovery with a confirmatory recall floor. Replaces two prior PRs (#662, #663) and adds the rest of the ma-discovery-integration slice. Shipped spec tasks 1–5 (fail-closed search, confirmed pairs, LLM pair matching, C3 join, frozen-snippet CLI). Tasks 6–8 deliberately held back (study stays exploratory, materialization stays closed).

All CI checks pass. No reviews yet.

**Things to note for reviewers:**
- Largest PR by file count and lines (47 files, +5.8k)
- Body notes: "Promotion still needs evidence-auditor. Do not cite these numbers."
- The confirmatory recall study met all three gates (≥10 strict C3 inserts, 19/0/1 precision, cost under $0.10/pair)
- CLI supports `--skip-pairs`, `--stop-when dated_confirm`, `--strict-recall`, `--fail-on-gate`

**Recommended action:** Needs review. This is a substantial change — review the spec task mapping and confirm that shipping tasks 1–5 without tasks 6–8 leaves no dangling dependencies.

hollomancer:
## Consolidated review — request changes

Reviewed at `0fde3d95`.

- **Blocker — the held-out partition is not actually disjoint.** `scripts/data/run_ma_discovery_sample.py:262-288,860-869,922` partitions and counts raw `(company_name, acquirer)` values, while the protocol defines distinct `RECIPIENT_V1`-normalized pairs. Five normalized identities in positions 501-1500 already occur in the first 500. The claimed held-out result does not satisfy its protocol; partition on normalized `pair_key`, refreeze, and rerun before retaining the reported 13.
- **Blocker — high confidence does not require two confirming sources.** `orchestrator.py:73-105` counts every seen URL, including rejected/duplicate hits, and `dated_confirm` exits on the first dated confirmation. Collect distinct URLs only from confirmed verdicts and continue until a genuine second confirmation or exhaustion.
- **Blocker — fault and cost gates are not implemented end-to-end.** `openai_client.py:118` converts terminal transport failure to `None`; `extractor.py:310` turns that into an ordinary negative, so `_CountingExtractor` sees zero timeouts. `run_ma_discovery_sample.py:950-962` also always passes `cost_per_pair_usd=None` and `cost_measured=False`, despite the tracked manifest claiming a passed cost gate. Propagate explicit faults and measured usage/cost through the same process that writes the result.
- **Blocker — C3 can credit the wrong event.** `collision.py:101-147` selects a same-company dateless row before a later dated match and can promote an existing low row regardless of the discovery's own confidence. Match dated candidates first and count only discoveries that themselves qualify as medium/high.
- **Cross-PR blocker — evidence schema conflicts with #701.** Inserts/collisions use `signals.discovery_confirmed`, `source`, and `evidence`; #701 recognizes different fields and discards the provenance. Standardize the contract before merging either branch.

The 99 focused tests pass but do not cover these measured-cut failures.

scripts/data/run_ma_discovery_sample.py
hollomancer: Landed in 56ea0087. `RecordingLlmExtractor` raises `LlmTransportError` on None-after-retry; `_CountingExtractor` counts it. OpenAIClient timeout → recording → counting asserts `timeout_n == 1` and writes no freeze row.

sbir_etl/enrichers/ma_discovery/collision.py
hollomancer: Landed in 56ea0087. C3 checks the ±30-day dated window first; company-only dateless fallback only if no dated collision. Fixture `[dateless EFTS, dated row]` binds a dated discovery to the second row.

specs/ma-discovery-integration/design.md
hollomancer: Landed in 56ea0087. Open-decision table and step 3 now match the implemented contract: runtime default is `none` (fail-closed); `mock` is explicit opt-in.

sbir_etl/enrichers/openai_client.py
hollomancer: Landed in 56ea0087. `OpenAIClient.chat` returns `None` on 401/402/403 by default (weekly synopsis still omits). M&A capture sets `raise_on_auth_error=True`; a 402 raises `OpenAIAuthError` and aborts without a freeze row.

scripts/data/run_ma_discovery_sample.py
hollomancer: Landed in 56ea0087. `strict_medium_high_n` counts `applied_hits` at extractor (pre-C3) confidence on the discovered pair. A low discovery that C3 would bump to medium is not a strict-recall credit.

sbir_etl/enrichers/ma_discovery/orchestrator.py
hollomancer: Landed in 56ea0087. Under `dated_confirm`, an undated confirm is not added to `seen`. Later query templates still run; the first undated is emitted only after templates are exhausted.

sbir_etl/enrichers/ma_discovery/orchestrator.py
hollomancer: Landed in 56ea0087. `source_count` is distinct confirming URLs only. One dated+value confirm after an unconfirmed URL stays `medium`; two confirming sources with date+value is `high`.

hollomancer:
Review replies at `56ea0087`.

**Held-out partition left open.** `bound_queries` still skips/keeps on raw `(company_name, acquirer)`. Switching that cut to `RECIPIENT_V1` `pair_key` would re-select pairs 501–1500 and 1501–2500 against `confirmatory-run-manifest.json` and `held-out-run-manifest.json`, so the recorded 13 and 4 would no longer be the protocol cut. Not rewriting those frozen counts on this branch.

Issue #682, opened 2026-08-29
Public release tracker: SBIR M&A preprint v1


hollomancer:
## Release status

**Not ready for public deposit.** PR #672 is a useful packaging draft, but evidence, privacy, interpretation, review, and personal-capacity gates remain open.

This issue is the single coordination point for preparing the SBIR M&A paper for public release.

## Release artifact

- Packaging PR: #672
- Working paper version: 1.0.0
- Planned outlet: Zenodo preprint
- Publication action: human only
- Current evidence status: exploratory

### 1. Public-data boundary

- [ ] #676 — audit proposed files for privacy, redistribution, minimization, and licensing
- [ ] Decide which artifacts may enter public Git LFS
- [ ] Decide whether a separate public derivative is required

Do not place raw study payloads in the public repository before #676 closes.

### 2. Reproducibility

- [ ] #674 — recover and freeze the April 2026 / PR #286 study snapshot with Git LFS
- [ ] Pin SHA-256, byte sizes, row counts, schemas, and provenance
- [ ] Add blocking verification
- [ ] Establish whether the snapshot is `recovered-original` or only `recovered-consistent-with-committed-results`

### 3. Measurement validation

- [ ] #675 — complete blinded adjudication of accepted M&A detections
- [ ] Report tier-specific and weighted precision
- [ ] Report false-discovery rate and confidence bounds
- [ ] Separate classification error from SEC coverage limits

### 4. Observation-window decision

- [ ] #673 — decide whether vintage-adjusted cumulative incidence is required for v1

Choose one path explicitly:

- **Include in v1:** complete and integrate the vintage analysis; or
- **Defer to v2:** retain 8.1% only as pooled historical accepted-signal prevalence, not a lifetime acquisition probability.

### 5. External interpretation

- [ ] #678 — reconcile external acquisition comparators
- [ ] Remove any comparison whose denominator, horizon, method, or independence cannot be established

This issue is a hard gate only if comparative claims remain in the manuscript.

### 6. Manuscript and metadata

- [ ] #677 — revise the preprint and metadata
- [ ] Use final denominator language
- [ ] Distinguish raw and precision-adjusted results
- [ ] Add AI-assistance disclosure
- [ ] Add `exploratory preprint — not peer reviewed`
- [ ] Preserve the exact personal-capacity statement
- [ ] Remove all public-file placeholders and unresolved `[UNVERIFIED]` markers

### 7. Personal-capacity review

[...]

PR #672, opened 2026-08-29
Package SBIR M&A working paper as Zenodo v1 preprint

hollomancer:
## Summary

- packages the committed SBIR M&A working paper as an exploratory public preprint
- adds Markdown source and a built PDF under `studies/sbir-ma-exits/`
- adds claim-level provenance and one consolidated list of unverified items
- adds Zenodo v1 metadata, repository/paper cross-citation, and exact human publication steps

## Evidence boundary

This is packaging only. No pipeline was run and no value was recomputed.

The repository still classifies these M&A findings as exploratory and not approved for citation as evidence. This package preserves that status. A Zenodo deposit would make the document discoverable, but it would not validate the underlying detections.

## Checks

- abstract: 168 words
- paper: 7 rendered pages
- exact personal-capacity statement present
- firms without detections use `not-yet-observed`
- paper license: CC BY 4.0
- code license: MIT
- analysis pin: `v0.11.0` at `241fe9d763309aabc28eb6d0cf85bb536c9289e8`
- DOI remains `[UNVERIFIED — human review]`
- no Zenodo record was created or published

PR #705, opened 2026-09-09
Drop Form D matches the scorer already rejected (do not land yet)

hollomancer:
Draft — **not ready to land.** Two open questions below must close first, and the diff changes numbers a preprint depends on.

## The defect

`form_d_details.jsonl` carries the multi-signal verdict from `compute_form_d_confidence` in `match_confidence.tier`. `detect_sbir_ma_events.py` never read it, so a Form D filed by an unrelated company gets attributed to an SBIR firm and then graded `high` by `assign_confidence`, which asks only whether a Form D exists.

Of the 552 business-combination records, **374** have an SEC filer name that does not match the SBIR company name under `RECIPIENT_V1`, and **323 of those 374 were already tier `low`**. The scorer caught them; the detector overrode it.

The case that surfaced it: SBIR firm `3D Control Systems, Inc.` joined to filer `3D SYSTEMS CORP` — CIK 0000910638, Rock Hill SC, "Over $100,000,000" revenue. A large public 3D-printing company, graded `high`, null acquirer, one signal.

Three sibling consumers of the same file already filter on tier (`capital_events/sources/form_d.py:61-62`, `bootstrap_form_d_leverage_ci.py`, and the FY2026 commercialization benchmark). This path was the outlier.

## Effect

| | before | after |
|---|---:|---:|
| Form D business-combination events | 552 | 141 |
| high-confidence events | 1,197 | 811 |
| high with null acquirer | 407 | 105 |

## Why this cannot land yet

**1. Mismatch rate is not error rate.** The 374 figure counts names failing strict normalization, which is not the same as wrong. `Bruker Analytical Systems -> BRUKER CORP` and `Sepratech -> Sepratech TopCo, LLC` look like legitimate parent/subsidiary and rename relationships that strict matching wrongly rejects, while `SiliconCore Technology -> SILICON STORAGE TECHNOLOGY` looks like a genuine false positive. **51 high/medium-tier mismatches need adjudication** — sample at `data/processed/form_d_join_adjudication.jsonl`, seed 0, `label: UNREVIEWED`. Until that is labeled, the true error rate is unknown and so is the right threshold.

**2. The producer is inferred, not confirmed.** The edited file is `scripts/archive/data/detect_sbir_ma_events.py`. Two indirect signals point to it as the producer of `sbir_ma_events.jsonl` — its own `--output` default (`:240`) and `run_agency_private_capital_phase1.py:13-14` documenting it as such. But the data file is gitignored (`.gitignore:147`) so there is no history, and its mtime (Apr 24) predates this copy of the script (Jul 12). **If that inference is wrong, this PR fixes nothing.** Re-running the chain against the pre-fix code and checking it reproduces 4,306 events / 1,197 high would settle it.

**3. The direction question is unresolved.** `is_business_combination` maps to `businessCombinationTransaction/isBusinessCombinationTransaction` — Form D Item 3, checked by the **issuer raising capital**. The usual reading is a firm raising money to *fund* a combination, i.e. the acquirer rather than the target. Nothing in the pipeline distinguishes direction. Not verified against filings. If direction is inverted, the surviving 141 need rethinking too, not just re-tiering.

## Blast radius

`sbir_ma_events.jsonl` -> `press.py` -> `enriched_sbir_ma_events.jsonl`, read by `capital_events/sources/ma_events.py`, `build_nano_cohort.py`, `nano_ma_signal.py`, `nano_prime_acquisitions.py`. Note `ma_events.py` filters on *event* confidence — the inflated one — not match tier, so all 407 null-acquirer events currently enter `capital_events` with a null counterparty.

The preprint (#682, PR #672) cites **1,197 high-confidence detections** feeding 2,790 high-plus-medium and the headline **8.1 percent** exit rate. This is a recompute, not a footnote. See #682 for detail.

## Also not fixed here

The 105 surviving null-acquirer events are well-matched filings that still grade `high` on a checkbox with no counterparty, and still outrank EFTS full-text evidence of an actual acquisition. That is the grading defect, separate from the join, and it needs the direction question answered first.


hollomancer:
## Caveat 2 update: producer confirmed for the Form D path, but the dataset is not reproducible overall

Ran the pre-fix code against the same inputs. The Form D path reproduces **exactly**:

```
PRE-FIX code   events=4306  high=1197  high-null-acquirer=407
shipped file   events=4306  high=1197  high-null-acquirer=407
```

So `detect_sbir_ma_events.py` is the producer for the path this PR touches, and the fix lands where intended. Caveat 2 is closed for the Form D path.

**But the file as a whole does not reproduce.** Shipped medium/low is 1593/1516; the pre-fix code gives 1178/1931.

Cause: 420 shipped events are `medium` on `efts_ma_definitive` alone, plus 67 on `ma_definitive + ma_proxy`. Current code maps `ma_definitive -> "low"` (`:90`). `MA_TYPES` has only ever existed in one committed form — introduced by `aa057c12` (PR #286) with `ma_definitive: "low"`, never edited since, and the archive move changed only two docstring lines.

Timeline:

```
Apr 21 18:35  sec_edgar_scan.jsonl          input
Apr 23 13:05  form_d_details.jsonl          input
Apr 23 18:23  sbir_ma_medium_refined.jsonl  1178 rows = the pre-fix medium count
Apr 24 06:38  sbir_ma_events.jsonl          output
Apr 30 12:31  PR #286 merged
```

Same input, different output, so the difference is code — and the data predates the merge by six days. `ma_definitive` graded `medium` in the development branch, was changed to `low` before merge, and the data was never regenerated.

**No committed version of this repo reproduces `data/sbir_ma_events.jsonl`.**

This does not change the defect this PR fixes or the measured effect (552 -> 141 Form D events, 1197 -> 811 high). It does mean a full chain re-run would rebaseline the medium tier for a reason unrelated to this fix, so regeneration should be a separate, deliberate operation.

Also relevant: `refine_ma_medium_tier.py` requires live SEC EDGAR traffic (~1,200-1,600 document fetches), so the chain cannot be re-run offline.

hollomancer:
## Blocker 3 resolved: direction is issuer-side. This PR is necessary but not sufficient.

Full evidence on #682. Summary: Form D **Item 10** flags a **Rule 145** transaction, which is a deemed offer and sale of securities *by the issuer* to the other company's holders. The filer issues the stock, so the filer is the **acquirer**. Of 23 `clarificationOfResponse` texts read from EDGAR, 18 say the issuer acquired, 5 are corporate reorganizations, 0 say the issuer was acquired. Corroborated by a monotone 25x base-rate gradient with issuer size (2.1% at no-revenue, 54.0% over $100M), a fully-subscribed-at-filing rate of 66.4% vs 30.8%, a 5.8-year median gap to any EFTS mention, and 3 of 3 testable confirmed targets carrying no flag.

### What that means for this PR

The tier filter is still worth landing — the join defect is real and independent, and the reviewer confirmed `tier == "high"` is the right threshold because `medium` carries no identity evidence at all (`form_d_scoring.py:339-344`: person >= 0.7 OR ZIP -> high; state overlap alone -> medium).

But **this PR alone leaves the pipeline recording acquirers as exits**, just fewer of them. 1,197 -> 811 is a smaller wrong number. The 141 surviving combo records are, on a 16-row sample, 5 explicitly acquirer-side, 2 reorganizations, 0 targets, 9 empty clarification.

### Consequence for sequencing

The earlier reviewer note said the grading fix should land with this or before it. The direction finding strengthens that: `assign_confidence` returning `high` on `has_form_d` alone is not merely over-confident, it is **directionally wrong**, and it outranks EFTS full-text evidence that is directionally correct.

I would now go further — the Form D combination flag should probably not produce an M&A **exit** event at all in its current form. Options, in increasing order of work:

1. Drop `form_d_business_combination` as an exit signal; keep it as a separate `acquirer_side` marker.
2. Parse `clarificationOfResponse` and classify per filing (acquirer / reorganization / target / unknown), keeping only genuine target-side cases.
3. Invert the signal deliberately — see below.

### The interesting possibility

`clarificationOfResponse` names the counterparty and the direction. That makes it a candidate signal in its own right, and possibly a better one than the boolean ever was: searching **other** firms' clarifications for an SBIR firm named as the acquired party would detect acquisitions of SBIR firms from the acquirer's own filing — the direction A1/A3/A4 actually care about.

Coverage is the open question: the field is optional and roughly half the sampled offerings leave it empty. The backfill proposed on #682 would measure that.

### Still open on this PR

- The 81-row adjudication sample (`data/processed/form_d_join_adjudication.jsonl`, all `UNREVIEWED`, stratified 30 zip_only / 30 person_and_zip / 21 person_only). The `zip_only` stratum is the risk: 58 of 84 survivors share a ZIP with 5+ SBIR firms, and 16 more have `address_score: 1.0` with no reproducible ZIP match.
- Regeneration of `sbir_ma_events.jsonl` and the enriched file, which needs `refine_ma_medium_tier.py` (~1,200-1,600 live SEC EDGAR fetches) and will independently move the medium tier by ~490 events, since the current file predates PR #286's merge and reproduces from no commit.

Both commits are pushed (`293d55b7`, `683a299c`). Still draft.

hollomancer:
## Census backfill confirms the grading fix. No change needed to this PR.

Fetched `primary_doc.xml` for all 645 business-combination accessions (645/645 HTTP 200, 0 parse errors). Full detail on #682.

**Classification of the 353 non-empty clarifications: 302 acquirer-side (85.6%), 32 reorganization, 15 ambiguous, 4 target-side (1.1%).** Of the 4 target-side, only one survives adjudication of the SBIR join.

Against the 407 Form-D-only events this PR's grading change targets: 215 have text — **180 acquirer-side, 21 reorganization, 10 other, 4 target-side.** Roughly 84% of the readable half is directionally inverted and another 10% is not M&A in either direction.

So `9df63cf1` (Form D combination no longer grades an exit) is confirmed at census scale rather than sample scale. The remaining 192 no-text events stay unverifiable, which is an argument for the removal rather than against it.

One incidental result worth recording: all 645 filings carry the boolean correctly in the XML. The extractor reads it faithfully. The defect was entirely in interpretation, not parsing.

### Follow-up, not blocking this PR

The clarification names the target often enough to invert into a target-side signal — a bounded test found **3 SBIR firms this pipeline misses entirely** plus 3 date corrections. That work is now #707, and it needs the full DERA corpus from #582 rather than `form_d_details.jsonl`. It does not gate this PR.

### Still outstanding here

- The 81-row adjudication sample, untouched.
- Regeneration of `sbir_ma_events.jsonl` and the enriched file.

On regeneration I would still do it **once**, after adjudication, rather than now: it will move the medium tier by a further ~490 events for an unrelated reason (the shipped file predates PR #286's merge and reproduces from no commit), and regenerating erases the evidence of that discrepancy before anyone has explained it.

scripts/data/apply_ma_medium_refinement.py
```diff
@@ -0,0 +1,142 @@
...
+                row["refined_direction"] = direction
+                stats[f"demoted_{direction}"] += 1
+            else:
+                row["refined_direction"] = direction
+                stats[f"kept_{direction}"] += 1
+        elif row.get("confidence") == "low" and signals.get("efts_ma_definitive"):
+            if direction is None:
+                stats["ma_definitive_no_verdict"] += 1
+            elif direction == PROMOTE_DIRECTION:
+                row["confidence"] = "medium"
```
hollomancer: **[P1] The legacy `target` verdict is not safe enough to cross the low-to-medium inclusion threshold.** `classify_direction()` currently returns `target` for the explicitly active sentence `ACME acquired Target LLC.` because `_TARGET_AFTER` makes the passive auxiliary optional; its generic `merger agreement` fallback is also role-blind. This promotion can therefore recreate the same acquirer/target inversion this PR is correcting. Please require role-sensitive target grammar and add active-versus-passive regression tests before promoting these rows.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -145,17 +182,31 @@ def merge_events(
...
+
+    The flag previously returned ``high`` on its own, which put a
+    self-reported acquirer-side boolean above EFTS full text that names a
+    filer and has passed directional review. 407 events carried it with no
+    other signal.
+
+    The flag is still recorded in ``signals.form_d_business_combination`` --
+    it is real evidence of a combination, in the other direction, and which
+    SBIR firms are doing the acquiring is a question worth keeping. It just
+    does not grade an exit.
```
hollomancer, line 240: **[P2] Once Form D is retained as opposite-direction evidence, its date cannot be allowed to become the exit date.** `merge_events()` still chooses the earlier Form D/EFTS date. A synthetic 2010 Form D record plus target-side 2020 EFTS acquisition evidence becomes a medium-confidence exit dated 2010; on the available inputs the Form D date wins in 41 of 48 overlaps. Please source the top-level exit date from the target-side evidence and retain the Form D date only in its nested detail.

docs/research/sbir-ma-exit-analysis.md
```diff
@@ -5,11 +5,99 @@
...
+population, produced end to end from committed code:
+
+| | published | regenerated |
+|---|---:|---:|
+| high | 1,197 | **676** |
+| medium | 1,593 | **1,098** |
+| low | 1,516 | 2,230 |
+| events | 4,306 | 4,004 |
+| high + medium | 2,790 | **1,774** |
+| **exit rate (of 34,460 firms)** | **8.1%** | **5.1%** |
```
hollomancer, line 75: **[P1] This regenerated headline no longer matches the final code in this PR.** The last commit now promotes low `ma_definitive` rows with a `target` verdict—420 rows by the PR's own account—while this table says those promotions are absent and reports 1,098 medium / 1,774 H+M / 5.1%. Lines 88-94 also say regeneration has not occurred, contradicting lines 58-59. Please rerun the current HEAD, reconcile the memo, and add an aggregate chain regression check before publishing a corrected rate.

scripts/data/refine_ma_medium_tier.py
```diff
@@ -283,15 +283,23 @@ async def main():
...
+    # events sat below the review threshold and needed an out-of-band promotion
+    # before refinement could reach them. A definitive merger agreement naming
+    # the firm is exactly the case worth reading the filing for.
     events = []
     with open(args.input) as f:
         for line in f:
             r = json.loads(line)
-            if r["confidence"] == "medium":
+            signals = r.get("signals") or {}
+            if r["confidence"] == "medium" or signals.get("efts_ma_definitive"):
```
hollomancer: **[P2] This is now an actively maintained producer but remains under `scripts/archive/`.** The same PR adds an archive rule saying active producers must leave the archive because editing them changes what the archive reproduces, yet the new apply step depends on this materially widened output. The canonical research note also says rerunning the detector alone refreshes the signal, which is no longer true. Please freeze the archived implementation and add/move an active successor, then document one canonical detector → refiner → apply/finalize command.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -145,17 +182,31 @@ def merge_events(
...
-        efts is not None and "subsidiary" in efts.get("mention_types", [])
-    )
-    has_acq_text = efts is not None and (
-        "acquisition" in efts.get("mention_types", [])
-    )
+    has_efts_high = efts is not None and "subsidiary" in efts.get("mention_types", [])
+    has_acq_text = efts is not None and ("acquisition" in efts.get("mention_types", []))
 
-    if has_form_d or has_efts_high:
+    if has_efts_high:
```
hollomancer, line 246: **[P1] Lowering Form-D-only rows to `low` does not stop existing consumers from counting them as exits.** `main()` still writes every merged record, while `agency_private_capital/asset.py::_load_ma_event_companies`, `phase2_outcomes.py::keys_from_ma_events`, `run_agency_private_capital_phase1.py`, and the archived leverage decomposition all treat row presence as an exit without filtering `confidence`. The surviving Form-D-only, acquirer-side records therefore remain false positives in those metrics. Please either omit/separate acquirer-only rows from the exit artifact or update every exit consumer to require target-side high/medium evidence.

docs/research/sbir-ma-exit-analysis.md
```diff
@@ -5,11 +5,99 @@
...
+| tier | published | corrected |
+|---|---:|---:|
+| High | 1,197 | **676** |
+| Medium | 1,593 | **1,211** |
+| High + Medium | 2,790 | **1,887** |
+| events total | 4,306 | 4,004 |
+
+High-plus-medium falls 32%.
+
+**Regenerated 2026-09-09.** `data/sbir_ma_events.jsonl` now holds the corrected
```
hollomancer: **[bug]** The correction section cannot be used as written. Line 58 says `data/sbir_ma_events.jsonl` “now holds the corrected population” with a 5.1% exit rate (1,774 H+M) and input hashes; line 88 still says the same file “still holds the published numbers” and that regeneration has not been done. Those two paragraphs were stacked across commits and never reconciled. Worse, the 5.1% / medium=1,098 figures come from `7205798b`, which explicitly dropped 420 published medium events as “unexplained promotions.” HEAD (`5fae303c`) then taught `apply_ma_medium_refinement.py` to put those 420 back (low + `efts_ma_definitive` + `direction=="target"` → medium) while leaving lines 82–86 claiming the mechanism is “still unknown” and that those rows are absent from the regenerated file. Re-running the committed pipeline would not reproduce the table on line 58, and none of `sbir_ma_events.jsonl`, the hashes, or `ma_events_unexplained_promotions.jsonl` are in this PR. This is also citable-claim drift: a new headline rate from exploratory scripts, shaped like an evidence result, in a note a preprint already cites.

**Suggestion:** Pick one state of the world and delete the other. Either (a) the artifact is not regenerated — keep “not yet done,” remove the 5.1% table and hashes — or (b) it is regenerated from *current* HEAD, including the 420 promotions, with a new hash and a single table. Strike the “mechanism is still unknown” sentence; it is false as of `5fae303c`. Do not land a new exit rate in this note until the file that rate describes is the file the committed scripts emit.

docs/research/sbir-ma-exit-analysis.md
```diff
@@ -92,7 +180,7 @@ original scan, so they are inherently weaker candidates.
 
 | Tier | Rule | Count |
 |------|------|-------|
-| High | Form D business combination OR EFTS `subsidiary` | 1,197 |
+| High | ~~Form D business combination OR~~ EFTS `subsidiary` | ~~1,197~~ 676 |
```
hollomancer: **[bug]** The “Confidence Tiers (Final)” table was patched in one cell and left inconsistent with itself and with the correction section. High is now “~~Form D business combination OR~~ EFTS `subsidiary` | ~~1,197~~ 676”, but Medium is still 1,593, Low 1,516, Total 4,306. 676+1,593+1,516=3,785, not 4,306. The correction tables above give Medium as 1,211 (detect-only) or 1,098 (regenerated) and total events 4,004. Below this table, Exit Rate / agency / high-tier composition / caveats still treat Form D as a high-confidence exit (552 companies, 407 Form-D-only events that “identify the target”). The supersede banner does not save a methodology section that still describes the defect as current method.

**Suggestion:** Either freeze the rest of the note as a dated historical snapshot (explicitly: every table below the banner is the 2026-04-23 published file) or update every count and the Layer 1 “high confidence / 552 companies” paragraph to the same corrected population. A one-cell strikethrough on High is worse than no edit.

scripts/data/refine_ma_medium_tier.py
hollomancer: **[suggestion]** This PR’s archive README now says an actively maintained producer does not belong in `scripts/archive/` because editing it in place changes what the archive reproduces — and then this commit changes the archive producer’s eligibility from `confidence == "medium"` to `medium or efts_ma_definitive`. That is a live behavior change (more EDGAR fetches; the input to `apply_ma_medium_refinement.py`) left in the archive next to the copy that produced the 2026-04 published refinement file. `apply_ma_medium_refinement.py` was correctly added under `scripts/data/`; the refiner it depends on was not moved.

**Suggestion:** Move `refine_ma_medium_tier.py` to `scripts/data/` the same way the detector moved, or leave the archive copy untouched and put the eligibility change on a `scripts/data/` fork. Do not both announce the archive-immutability rule and violate it in the same PR.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -18,18 +18,46 @@
...
 def extract_form_d_signals(records: list[dict]) -> list[dict]:
     """Extract M&A events from Form D business combination flags.
 
     For each company with at least one is_business_combination offering,
     produces one event using the earliest combo filing date.
+
+    A record is used only when its SBIR-to-SEC match tier is ``high``. The
+    join is fuzzy, and ``form_d_details.jsonl`` already carries the
+    multi-signal verdict from ``compute_form_d_confidence`` in
+    ``match_confidence.tier``. Without this filter a filing by an unrelated
```
hollomancer, line 40: **[suggestion]** `extract_form_d_signals`’s docstring still describes current `assign_confidence` behavior as “a filing by an unrelated company is attributed to an SBIR firm and then graded `high`.” That was true when the filter landed; `9df63cf1` then removed Form D from the ladder, so an unmatched Form D is now a *low* event (or, with this filter, not an event). The same block (lines 37–53) plus `assign_confidence` (184–203), the `match_person_score` comment (83–85), and the `apply_ma_medium_refinement.py` module docstring are changelog/design-history, not why the next line exists. They will rot the next time the ladder moves — this docstring already has.

**Suggestion:** Collapse to the invariant: keep `match_confidence.tier == "high"` because the join is fuzzy and the scorer already decided; Form D Item 10 does not grade an exit because the filer is the issuer in a Rule 145 transaction. Move the 2026-09-08 counts, the 23-text census, and the 407-event autopsy to the research note.

docs/research/sbir-ma-exit-analysis.md
```diff
@@ -5,11 +5,99 @@
...
+
+**Not yet done.** `data/sbir_ma_events.jsonl` still holds the published
+numbers. Regeneration needs `refine_ma_medium_tier.py`, roughly 1,200-1,600
+live SEC EDGAR fetches, and will move the medium tier by a further ~490 events
+for an unrelated reason: the shipped file was generated six days before PR #286
+merged, its generating code changed before merge, and it reproduces from no
+commit. The corrected rate should be computed once, after regeneration, rather
+than restated twice.
+
+Also outstanding: 81 surviving Form D matches await human adjudication
```
hollomancer: **[suggestion]** “81 surviving Form D matches await human adjudication” overstates what `sample_form_d_join_adjudication.py` writes. The detector keeps the high-tier combination population (the note itself says 141, 84 of them ZIP-only). The script samples `--per-stratum 30`, so 81 is the worksheet size (30+30+21 or similar), not the surviving join. A reader of the open-questions list will think only 81 high-tier Form D combinations remain.

**Suggestion:** Say 141 high-tier combination records remain, of which 81 were drawn into `form_d_join_adjudication.jsonl` as `UNREVIEWED`. Keep the ZIP-only crowding numbers attached to the population (84/141), not the sample.

scripts/data/sample_form_d_join_adjudication.py
```diff
@@ -0,0 +1,205 @@
...
+    if address:
+        return "zip_only"
+    return None
+
+
+def load_sbir_firms_by_zip(awards_csv: Path) -> tuple[dict[str, set[str]], dict[str, set[str]]]:
+    """Return distinct SBIR company names per 5-digit ZIP, and each firm's ZIPs.
+
+    A firm holds awards across years and can move, so it has a set of ZIPs, not
+    one. The scorer compares the Form D ZIPs against all of them; taking only
```
hollomancer: **[suggestion]** The comment claims “the scorer compares the Form D ZIPs against all of [a firm’s ZIPs].” `scripts/archive/data/fetch_form_d_details.py` `load_award_data` stores the *first* CSV ZIP per exact `Company` key (`if zip_code and name not in zips`) and `compute_form_d_confidence` takes that single `sbir_zip`. Using every award ZIP is a reasonable worksheet enrichment, but it is not a reconstruction of the scorer, and it can show a shared ZIP the scorer never used (or, with the `.upper()` key vs the scorer’s raw `Company` key, fail to show the ZIP it did use). That is a plausible source of the 16 null `matched_zip` rows the docstring treats as “unverified ZIP claims.”

**Suggestion:** Reconstruct with the same rule as `load_award_data` (first ZIP, same name key the scorer used), and label any extra ZIPs as “other award ZIPs, not used for `address_score`.” Point at `fetch_form_d_details.py`, not an imagined all-ZIP scorer.

docs/research-questions.md
```diff
@@ -449,7 +449,7 @@ NASEM calls this quantity the *leverage ratio*.
 
 > **Implementation note — M&A detection is script-driven, not orchestrated.**
 > M&A event detection runs as a CLI script
-> (`scripts/archive/data/detect_sbir_ma_events.py`), not as a Dagster asset. The
+> (`scripts/data/detect_sbir_ma_events.py`), not as a Dagster asset. The
 > script merges two signals: Form D filings (entity_type-based
```
hollomancer, line 453: **[suggestion]** The implementation note was updated to the new script path but still describes the detector as merging “Form D filings (entity_type-based business-combination heuristics)” into the M&A *exit* signal that feeds A1/A3/A4. After this PR that flag does not grade an exit; treating it as one is the defect being fixed. Inventory readers who never open the research note will keep wiring Form D combinations into exit questions.

**Suggestion:** One clause: Form D Item 10 is retained as an acquirer-side signal and is not an exit grade; exits come from the EFTS mention scan plus directional refinement.

tests/unit/scripts/test_apply_ma_medium_refinement.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_missing_verdict_keeps_the_tier() -> None:
+    """Absence of a verdict is not evidence the company is a comparator."""
+    out, stats = apply_refinement([_event("Acme", "medium")], {})
+    assert out[0]["confidence"] == "medium"
+    assert stats["medium_no_verdict"] == 1
+
+
+def test_high_and_low_are_untouched() -> None:
```
hollomancer: **[suggestion]** `test_high_and_low_are_untouched` still says “Only the medium tier is in the refiner’s scope.” That was true when the merge step only demoted. HEAD promotes `low` + `efts_ma_definitive` + `target` (covered by later tests), so this test’s name and docstring now describe a rule the function does not implement. The assertion still passes only because the fixture’s low row has no `efts_ma_definitive` flag.

**Suggestion:** Rename/re-scope to “high is never moved; low without `efts_ma_definitive` is never moved,” and keep the promotion tests as the specification for the low+definitive path.

tests/.test_durations
```diff
@@ -2945,20 +2945,20 @@
...
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_extract_efts_signals_skips_passive": 0.000395727999148221,
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_extract_efts_signals_subsidiary": 0.00042061099975398974,
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_extract_form_d_signals_finds_business_combination": 0.000422087000515603,
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_extract_form_d_signals_skips_non_combo": 0.000372796999727143,
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_extract_form_d_signals_uses_earliest_combo_date": 0.00041444399994361447,
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_merge_events_both_sources": 0.0004504089997681149,
-    "tests/unit/scripts/archive/test_detect_ma_events.py::test_merge_events_separate_companies": 0.0003380479997758812,
+    "tests/unit/scripts/test_detect_ma_events.py::test_assign_confidence_acquisition_text_is_medium": 0.00043330300013622036,
+    "tests/unit/scripts/test_detect_ma_events.py::test_assign_confidence_form_d_is_high": 0.00041682500022943714,
```
hollomancer: **[nit]** Remapping durations from `tests/unit/scripts/archive/test_detect_ma_events.py` to `tests/unit/scripts/test_detect_ma_events.py` is justified by the move, but the file still lists `test_assign_confidence_form_d_is_high`, which was renamed to `test_assign_confidence_form_d_alone_is_low`. The new filter/ladder tests have no duration entries. Harmless xdist noise; it does not belong in a behavior PR.

**Suggestion:** Drop the durations hunk, or regenerate after the rename so keys match collected node ids.

scripts/data/refine_ma_medium_tier.py
```diff
@@ -41,19 +41,29 @@
...
 # ...in context with Merger Sub, Acquiror, etc.
+# Role-bearing terms only. "merger agreement", "plan of merger" and "agreement
+# and plan of" name a document, and a merger agreement names both parties, so
+# their presence says nothing about which side the company is on. They were
+# dropped 2026-09-09; what remains identifies a counterparty role, which implies
+# the company is the other side.
 _MERGER_AGREEMENT = re.compile(
-    r"(?:merger sub|acquiror|acquisition corp|merger agreement|"
-    r"plan of merger|agreement and plan of)",
+    r"(?:merger sub|acquiror|acquisition corp)",
```
hollomancer: **[bug]** `_MERGER_AGREEMENT` still matches `acquiror` anywhere in the 500-character window around the company name. The new comment claims that term "identifies a counterparty role, which implies the company is the other side," but a filing that names the SBIR firm as the Acquiror (`ACME Inc. (the "Acquiror")`) matches too. That used to be mostly harmless because `target` and `ambiguous` both kept medium. It is not harmless now: `refine_ma_medium_tier.py` reviews every `efts_ma_definitive` row regardless of tier, and `apply_refinement` promotes a low `ma_definitive` row to medium on `direction == "target"`. The new tests only cover simple active-voice sentences (`ACME Inc. acquired Target LLC`) and `"entered into a merger agreement"`; they never assert that a company labeled Acquiror is not a target. The result is an acquirer promoted into the high+medium exit set that `capital_events/sources/ma_events.py` consumes.

**Suggestion:** Drop `acquiror` from this bag-of-words check, or require it to refer to a *different* entity than the matched company (for example only treat `merger sub` / `acquisition corp` as target-side, and add a `not_target` path for `the "Acquiror"` / `as Acquiror` attached to the company name). Add a classifier test for `ACME Inc. (the "Acquiror") entered into an agreement and plan of merger with Target LLC` expecting not `target`, plus an `apply_refinement` case that a low `efts_ma_definitive` row with that verdict stays low.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -128,9 +168,15 @@ def merge_events(
...
-            if efts_date and (not existing_date or efts_date < existing_date):
+            # The EFTS date wins whenever there is one, even if it is later.
+            # The competing date comes from a Form D business-combination
+            # filing, which is acquirer-side: it dates the issuer raising
+            # capital to buy something, not the SBIR firm being acquired.
+            # Taking the earlier of the two put an acquirer-side date on the
+            # exit for 30 of 36 overlapping companies - nLight Photonics was
+            # dated 2013 from a Form D while its target-side evidence is 2026.
+            # The Form D date is still available in form_d_detail.
+            if efts_date:
```
hollomancer, line 179: **[suggestion]** Overlap dating now prefers any non-empty EFTS date, even when it is later than the Form D date. The function docstring immediately above still says the merge "Uses the earliest valid ISO date across sources." `specs/sbir-ma-match-rate-by-fy/requirements.md` was retargeted at this producer (path-only edit) and still describes the old earlier-of-two-dates rule as what the producer does. Anyone regenerating `sbir_ma_events.jsonl` and then running the FY signal-count reporter will get EFTS-dated rows while the spec still documents Form D-earlier dating.

**Suggestion:** Rewrite the `merge_events` docstring to match the new rule (EFTS date wins when present; Form D date is only the fallback and remains in `form_d_detail`). Update the spec's evidence-boundary paragraph to the same rule, or explicitly mark the earlier-date sentence as describing the pre-correction artifact only.

docs/research-questions.md
```diff
@@ -449,11 +449,18 @@ NASEM calls this quantity the *leverage ratio*.
...
 > business-combination heuristics) and an SEC EDGAR full-text mention scan across
 > multiple filing types — operationally 8-K, 10-K, DEFM14A, PREM14A, SC TO-T, and
-> SC 14D9 (see `scripts/archive/data/refine_ma_medium_tier.py`).
+> SC 14D9 (see `scripts/data/refine_ma_medium_tier.py`).
+>
+> **The Form D business-combination flag does not grade an exit.** Form D Item
+> 10 marks a Rule 145 transaction, a deemed offer and sale of securities *by
+> the issuer*, so the filer is the acquirer. It is retained as an acquirer-side
+> signal, and rows carrying only that flag are written to
```
hollomancer: **[suggestion]** The new implementation note says rows whose only evidence is the Form D flag are written to `data/sbir_ma_acquirer_side.jsonl`. That is true only after the join filter. `extract_form_d_signals` drops `tier != "high"` before any event is built, so a rejected-tier business-combination record is not an acquirer-side row either — it disappears. The note also never mentions the join filter, which is the change named in the PR title. A reader of the north-star doc would not know that low/medium Form D combinations are omitted from both outputs.

**Suggestion:** State both drop paths: (1) `match_confidence.tier != "high"` is discarded entirely; (2) a surviving high-tier combination with no EFTS signal is written only to `sbir_ma_acquirer_side.jsonl`.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -18,18 +18,49 @@
 sys.path.insert(0, str(Path(__file__).resolve().parents[2]))
 
 
+# Match tier accepted from form_d_details.jsonl. capital_events/sources/form_d.py
+# hardcodes the same value, so the two paths keep the same records.
```
hollomancer, line 22: **[suggestion]** The new comment says `capital_events/sources/form_d.py` "hardcodes the same value, so the two paths keep the same records." They share the `"high"` tier gate, not the record set. `form_d.py` keeps every high-tier offering (fundraising equity/debt included) and emits `FORM_D_FILING` rows; this script keeps only high-tier `is_business_combination` offerings and then further splits Form D-only rows out of the exit file.

**Suggestion:** Say the two paths share the `"high"` match-tier gate, not that they keep the same records. Prefer one named constant over a comment claiming they cannot drift.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -18,18 +18,49 @@
...
     """Extract M&A events from Form D business combination flags.
 
     For each company with at least one is_business_combination offering,
     produces one event using the earliest combo filing date.
+
+    A record is used only when its SBIR-to-SEC match tier is ``high``. The
+    join is fuzzy, and ``form_d_details.jsonl`` already carries the
+    multi-signal verdict from ``compute_form_d_confidence`` in
+    ``match_confidence.tier``. Without this filter a filing by an unrelated
+    company is attributed to an SBIR firm. When this filter landed such a row
```
hollomancer, line 41: **[suggestion]** Several new comments and docstrings narrate the defect history rather than the current rule: this `extract_form_d_signals` block, `assign_confidence` (the 23 EDGAR texts / 407 events writeup), the `merge_events` nLight anecdote, `apply_ma_medium_refinement.py`'s April 2026 420/1,199 reconstruction, and the `_TARGET_AFTER` / `_MERGER_AGREEMENT` "dropped 2026-09-09" notes. That belongs in the research note, not in the producer.

**Suggestion:** Keep a short why (Form D Item 10 is acquirer-side; the join already has a scorer tier; EFTS dates the target-side evidence). Move counts, dates, and prior-bug narrative to `docs/research/sbir-ma-exit-analysis.md`.

hollomancer:

### Status: 🔴 Blocked — Do Not Land Yet

- **Draft:** Yes
- **Merge status:** Merge conflicts with `main` (DIRTY)
- **Age:** 1 day
- **Size:** +1,419/-359 across 17 files

The title and body both make clear this should not land yet. Two blocking open questions identified:

1. **Mismatch rate ≠ error rate** — 51 high/medium-tier mismatches need adjudication (seed at `data/processed/form_d_join_adjudication.jsonl`). True error rate is unknown until labeled.
2. **Preprint dependency** — the diff changes numbers the preprint depends on.

The merge conflict will need resolution when ready.

**Recommended action:** Complete the mismatch adjudication, resolve preprint dependency, rebase, then bring out of draft.

hollomancer:

Reviewed at `bc77227e`. The new anchored target rules address the earlier loose-regex promotion concern. One separate cross-stage blocker remains:

- **Blocker — demoted false exits remain exits for presence-only consumers.** `scripts/data/apply_ma_medium_refinement.py:89-106` changes comparator/not-target rows to `confidence: low` but always appends them. These consumers add every JSONL name/CIK without inspecting confidence: `packages/sbir-analytics/sbir_analytics/assets/agency_private_capital/asset.py:165-193`, `phase2_outcomes.py:13-40`, and `scripts/archive/data/run_agency_private_capital_phase1.py:99-120`. A demoted ACME row therefore still produces an M&A outcome key. Filter all outcome readers to accepted high/medium target rows, or emit a separate exit-only artifact, and add a refiner-to-consumer integration regression.
- **Important — the memo still mixes superseded vintages.** `docs/research/sbir-ma-exit-analysis.md:58-102` retains provisional 5.1%/1,774 results and obsolete press/promotion explanations. Regenerate one current-head result with inputs/hashes after adjudication, or remove the interim quantitative claims.

Draft status, merge conflicts, unresolved Form D adjudication, and the absence of ordinary CI remain appropriate blockers.

hollomancer:
## Rebased onto current main; five commits dropped as duplicated

`main` independently landed equivalent work via #669/#698. Comparing:

| this branch had | main now has |
|---|---|
| `$`-anchored `_TARGET_BEFORE` | same, with matching reasoning in the comment |
| `_MERGER_AGREEMENT` dropped from the direction decision | same |
| `^`-anchored `_OPERATING_ENTITY` | same regex |
| refine scope extended to low-tier definitive-merger events | same |
| `apply_ma_medium_refinement.py` | `apply_ma_direction_refinement.py` |

Two people reached the same conclusions from the same evidence, which is corroboration that the analysis was right. It also means five of this branchs fourteen commits added nothing.

Dropped: `58351d5e`, `5e834fe6`, `70ff196c`, `89e42e30`, `1882366a`.

## What remains — the Form D direction work

Nine commits, none duplicated on main:

- Form D matches below `tier == "high"` are dropped before an event is built
- The Form D business-combination flag no longer grades an exit. Item 10 is acquirer-side: a Rule 145 transaction is a deemed offer and sale of securities *by the issuer*
- Rows whose only evidence is that flag go to the non-exit artifact, not the exit artifact
- Refuted rows (`not_target`, `comparator`) leave the exit artifact rather than being demoted within it — three consumers key off row presence and never read `confidence`
- The exit date comes from target-side evidence; a Form D filing date is acquirer-side and was winning for 30 of 36 overlapping companies
- The detector moves out of `scripts/archive/` (main still has it archived)
- The memo records the figures as superseded

## Resolution notes

Three conflicts, all refiner files main now owns — resolved by dropping them from this branch rather than merging two versions of the same fix. One doc reference retargeted to `apply_ma_direction_refinement.py`.

699 tests pass across `tests/unit/scripts/` and `tests/unit/capital_events/`. `make lint-boundaries` and `make docs-check` clean.

Still a draft. The 81-row adjudication worksheet is unlabeled, and the corrected figures need regenerating once the remaining fixes settle.

hollomancer:
## Thread triage against HEAD `eefa0f8c`

Draft. Do not land. 19 unresolved inline threads classified; nothing in this comment changes code.

**Addressed in HEAD — reply and resolve:**
- Form D date as exit date (EFTS wins; tests cover it)
- Form-D-only rows still counted as exits (now written to `sbir_ma_non_exit.jsonl`)
- 81 vs 141 wording (memo now says 141 remain, 81 sampled)
- ZIP reconstruction comment (worksheet labeled wider than the scorer)
- `test_durations` stale node id
- Both drop paths in `research-questions.md`

**Stale after rebase / #669+#698 — resolve, do not re-implement here:**
- `apply_ma_medium_refinement` / archive refiner / `_MERGER_AGREEMENT` / `_TARGET_AFTER` threads. Equivalent direction work is on main as `apply_ma_direction_refinement.py`.

**Still true of HEAD:**
- Memo 5.1% / 1,774 table and methodology that still treats Form D as a high-confidence exit
- Docstring / north-star wording (`merge_events` still says earliest-of-two dates; “keep the same records”; defect-history comments)

### Separate from the threads — real land-blockers

1. **Restore `scripts/archive/data/refine_ma_medium_tier.py`.** Fast Tests fail with `ModuleNotFoundError` because the rebase deleted it while archive tests still import it.
2. **Rebase onto current main (v0.13.0).** Form D `high` is now `corroborated-person-v2`; person-name alone is no longer high. The 141/81 figures need a re-count after that. GitHub mergeable state is CONFLICTING (10 ahead / 12 behind).
3. 81-row worksheet is still `UNREVIEWED`.
4. Corrected events file is not regenerated at final HEAD.
5. Main’s `apply_ma_direction_refinement.py` still appends demoted rows to the exit artifact. Presence-only consumers never read `confidence`. That is now a **main** defect unless this PR re-takes it.

Preprint #682 still cites 1,197 high / 8.1%. This PR must not land until adjudication + regeneration + consumer filter are a deliberate operation.

scripts/data/refine_ma_medium_tier.py
hollomancer: Stale. `_MERGER_AGREEMENT` / `acquiror` bag-of-words is gone on main; this PR no longer owns the file. Resolving.

scripts/data/refine_ma_medium_tier.py
hollomancer: Stale. Archive refiner is no longer this PR's producer. Resolving.

tests/unit/scripts/test_apply_ma_medium_refinement.py
hollomancer: Stale. `test_apply_ma_medium_refinement.py` was dropped with the apply script. Resolving.

scripts/data/refine_ma_medium_tier.py
hollomancer: Stale. The rebase dropped this producer; equivalent work is on main. Resolving rather than re-implementing here. The remaining archive issue is the opposite one: HEAD deleted `scripts/archive/data/refine_ma_medium_tier.py` while archive tests still import it, which is why Fast Tests are red.

scripts/data/detect_sbir_ma_events.py
hollomancer: Addressed for Form-D-only rows: they now go to `data/sbir_ma_non_exit.jsonl` and are not written to the exit file. Resolving this thread.

Presence-only consumers on main's apply step (`apply_ma_direction_refinement.py` still appends demoted rows) is a separate follow-up. The 2026-09-10 consumer-filter commit was dropped with the five duplicated commits.

docs/research/sbir-ma-exit-analysis.md
hollomancer: Addressed. The note now says 141 high-tier combination records remain, of which 81 were sampled. ZIP crowding is attached to the population (84/141). Resolving.

scripts/data/apply_ma_medium_refinement.py
hollomancer: Stale after the rebase. Equivalent direction work landed on main via #669/#698 (`apply_ma_direction_refinement.py`); this PR no longer owns `apply_ma_medium_refinement.py`. Resolving.

docs/research-questions.md
hollomancer: Addressed. Both drop paths are now stated; artifact name is `sbir_ma_non_exit.jsonl`. Resolving.

scripts/data/detect_sbir_ma_events.py
hollomancer: Addressed in HEAD. `merge_events` prefers a non-empty EFTS date and keeps the Form D date in `form_d_detail`. Resolving.

scripts/data/sample_form_d_join_adjudication.py
hollomancer: Addressed. The worksheet comment now says it is wider than `load_award_data`'s first-ZIP scorer. Resolving.

tests/.test_durations
hollomancer: Addressed. The renamed node id is present; the old `test_assign_confidence_form_d_is_high` key is gone. Resolving.

scripts/data/detect_sbir_ma_events.py
hollomancer: Still open. `extract_form_d_signals` still narrates the 2026-09-08 autopsy. Collapse to the invariant (keep `tier == "high"` because the join is fuzzy; Item 10 does not grade an exit) and move the counts to the research note.

scripts/data/detect_sbir_ma_events.py
hollomancer: Still open. Code prefers EFTS; `merge_events` docstring still says “earliest valid ISO date,” and the FY spec still describes earlier-of-two as producer behavior. Rewrite the docstring (and the spec sentence) to match HEAD.

scripts/data/detect_sbir_ma_events.py
hollomancer: Still open. Defect-history comments remain in `extract_form_d_signals`, `assign_confidence`, and `merge_events`. Keep a short why; move counts and prior-bug narrative to the research note.

docs/research/sbir-ma-exit-analysis.md
hollomancer: Still open. The 5.1% / 1,774 table is still in the note; regeneration at final HEAD has not happened. Leave this thread open until one current-head result (or no quantitative claim) is in the memo.

docs/research-questions.md
hollomancer: Still open. The correction paragraph exists, but the first sentence still says the script merges Form D filings into the detection signal. One clause: Form D Item 10 is an acquirer-side signal and is not an exit grade.

docs/research/sbir-ma-exit-analysis.md
hollomancer: Still open. Banner says superseded, but the 5.1% table, hashes, and “mechanism is still unknown” remain. Same close-out as the headline-table thread.

scripts/data/detect_sbir_ma_events.py
hollomancer: Still open. The two paths share the `"high"` match-tier gate, not the record set. `form_d.py` keeps every high-tier offering; this script keeps only high-tier combinations and then splits Form-D-only rows out of the exit file.

docs/research/sbir-ma-exit-analysis.md
hollomancer: Still open. April counts now add up, but Layer 1 / caveats still describe Form D as a current high-confidence exit. Freeze the rest of the note as a dated 2026-04-23 snapshot, or update every count to the same corrected population.

docs/research-questions.md
```diff
@@ -449,11 +449,21 @@ NASEM calls this quantity the *leverage ratio*.
...
+>
+> **The Form D business-combination flag does not grade an exit.** Form D Item
+> 10 marks a Rule 145 transaction, a deemed offer and sale of securities *by
+> the issuer*, so the filer is the acquirer. It is retained as an acquirer-side
+> signal: rows that survive the SBIR-to-SEC match filter and carry only that
+> flag are written to `data/sbir_ma_non_exit.jsonl` rather than the exit
+> artifact, alongside rows the directional refiner finds are not acquisitions
+> of the firm, while rows whose match falls below the `high` tier are dropped
+> before any event is built. Do not wire Form D combinations into A1/A3/A4
+> exit questions.
```
hollomancer, line 466: **Blocking.** This implementation note says rows are written to `data/sbir_ma_non_exit.jsonl` “alongside rows the directional refiner finds are not acquisitions of the firm.” Grepping the head tree for writers of that artifact returns exactly one: `scripts/data/detect_sbir_ma_events.py`, which writes a single reason, `non_exit_reason='acquirer_side'`. Nothing writes `not_target` or comparator rows there. The commit that did (`491eb77d`) operated on `apply_ma_medium_refinement.py`, which the 2026-09-12 rebase dropped.

`docs/research-questions.md` is the north star and the scope gate. A false statement here is worse than no statement: a reader wiring an exit question will assume refuted rows are already excluded when they are not.

**Fix:** Cut the “alongside rows the directional refiner finds…” clause, or land the refiner-side routing in this PR. Keep the two Form D drop paths, which are accurate. If main’s `apply_ma_direction_refinement.py` is supposed to do the refuted-row half, say so with a file:line — the 12:04 triage already notes that main still *appends* demoted rows, so presence-only consumers still count them as exits.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -144,18 +190,60 @@ def merge_events(
...
+    Such rows do not belong in the exit artifact. Demoting them to low was not
+    enough: agency_private_capital/asset.py, phase2_outcomes.py, and
+    run_agency_private_capital_phase1.py all treat row presence as an exit and
+    never read confidence, so a demoted row still counted. They are written to
+    a sibling file instead, which keeps the evidence without letting an exit
+    consumer mistake it.
+    """
+    if not signals.get("form_d_business_combination"):
+        return False
+    return not any(signals.get(key) for key in EFTS_SIGNAL_KEYS)
```
hollomancer, line 219: **Blocking.** `EFTS_SIGNAL_KEYS` includes `efts_ma_proxy` and `efts_ownership_active`, so a Form D business-combination row is kept in the *exit* artifact if it carries a single SC 13D or merger-proxy mention. The docstring justifies the split as “With no EFTS mention alongside it there is no target-side evidence at all,” and `test_form_d_with_efts_evidence_is_not_excluded` encodes that keep-any-EFTS rule — but the memo’s own signal table grades `ma_proxy` (“may be comp table entry”) and `ownership_active` (“>5% stake with intent — pre-acquisition”) as Low, and neither is target-side evidence.

Such a row gets confidence `low` and is still written to `data/sbir_ma_events.jsonl`, where the three presence-only consumers count it as an exit. That reproduces, in a narrower population, the defect the split was built to close. Upper bound is the 36 of 141 surviving Form D records that overlap an EFTS event.

**Fix:** Restrict the keep-set to genuinely target-side evidence — `efts_subsidiary` and `efts_acquisition_text` — and route Form D rows whose only companion is `ma_definitive` / `ma_proxy` / `ownership_active` to the non-exit file with a distinct `non_exit_reason`. Rename the test to the rule it asserts and print per-reason counts in `main()`.

scripts/data/detect_sbir_ma_events.py
```diff
@@ -128,9 +168,15 @@ def merge_events(
...
+            # The EFTS date wins whenever there is one, even if it is later.
+            # The competing date comes from a Form D business-combination
+            # filing, which is acquirer-side: it dates the issuer raising
+            # capital to buy something, not the SBIR firm being acquired.
+            # Taking the earlier of the two put an acquirer-side date on the
+            # exit for 30 of 36 overlapping companies - nLight Photonics was
+            # dated 2013 from a Form D while its target-side evidence is 2026.
+            # The Form D date is still available in form_d_detail.
+            if efts_date:
                 existing["event_date"] = efts_date
```
hollomancer, line 180: **Blocking (docstring drift).** This code is the new date rule and is correct: EFTS wins when non-empty. The `merge_events` docstring above it still says “Uses the earliest valid ISO date across sources.” Raised 2026-09-09 (comment 3974101642); the behavior thread was resolved, the docstring was not.

`specs/sbir-ma-match-rate-by-fy/requirements.md:40-44` still describes the earlier-of-two-dates rule, and this PR retargeted that spec’s input line at this producer. The top-level `event_date` is the sole input to a fiscal-year histogram.

**Fix:** Rewrite the docstring to this rule, and either update the spec’s evidence boundary or mark that sentence as pre-correction only. Also reconcile “30 of 36” here with the earlier review measurement of “41 of 48 overlaps.”

hollomancer (commented):
## Open PR review — keep draft

Reviewed against `origin/main` at `493340ea`. Head `eefa0f8c`. The Form D join-filter and Item-10 grading fixes are real and well-evidenced (645/645 accessions; 302 of 353 clarifications acquirer-side). The branch around the fix is not ready. Complements the 12:04 thread triage; this pass adds the keep-set and north-star defects that triage did not raise as line comments.

Cannot formally request changes on an own PR; treat this as **needs work — keep draft**.

### Blocking

1. **Restore `scripts/archive/data/refine_ma_medium_tier.py`.** The rebase deleted it while archive tests and `apply_ma_direction_refinement.py` still import it. Best-supported cause of the uniform 4/4 Fast Tests shard failure. Three doc references now point at `scripts/data/refine_ma_medium_tier.py`, a path this PR never creates. The owner’s explicit ask was: move it, or leave the archive copy untouched — not delete it.
2. **`docs/research-questions.md` documents routing that does not exist.** The “alongside rows the directional refiner finds…” clause is false at this head. See line comment.
3. **Refuted-row half of P1 is missing.** Acquirer-side Form-D-only rows are split out. `not_target` / comparator rows are not. Main’s `apply_ma_direction_refinement.py` still appends demoted rows; presence-only consumers never read `confidence`. By the memo’s own counts that is 56+51 (original medium) / 95+93 (low-tier expansion) still counted as exits.
4. **`merge_events` docstring vs code.** Docstring says earliest-of-two; code says EFTS wins. FY spec still describes the old rule. See line comment.
5. **`EFTS_SIGNAL_KEYS` keeps Form D rows on `ma_proxy` / `ownership_active`.** Those are Low and not target-side. See line comment.
6. **Two mutually inconsistent “corrected” tables** in the memo (medium 1,211 vs 1,098; H+M 1,887 vs 1,774). Threads on lines 58 and 75 already open; still unresolved. Delete Table B, the 5.1%/2.0% rates, and the 16-hex “hashes.”
7. **PR body is first-commit numbers** (high 1,197 → 811). At head the Form D flag no longer grades an exit and the memo’s corrected high is 676.

### Number audit

3 of 13 checked figures hold as stated. 811 is the join-filter-only state, not head. The two correction tables disagree by 113 medium rows. “High-plus-medium falls 32%” is Table A; Table B is 36.4%.

### Do these first

1. Restore `scripts/archive/data/refine_ma_medium_tier.py` (or land it at `scripts/data/` with importers updated) — this is almost certainly what turns CI green.
2. Close the refuted-row half of P1: file:line that main routes those rows out, or land the routing here; cut the false clause in `docs/research-questions.md`.
3. Collapse the memo to one state of the world; rewrite the PR body against head (676, not 811).
4. Fix the `merge_events` docstring / FY spec, and decide whether `ma_proxy` / `ownership_active` should keep a Form D row in the exit artifact.

Do not leave draft until adjudication + regeneration + consumer filter are a deliberate operation. Preprint #682 still cites 1,197 high / 8.1%.

hollomancer:
## Review findings at `eefa0f8c`

I do not think this branch should be merged or rebased wholesale onto current `main`. It is currently conflicting, and all four Fast Tests shards fail. The qualitative correction—Form-D-only is not target/exit evidence—is valuable, but the implementation and numerical memo need rebuilding around current-main contracts.

1. **P1 — A required module is deleted without replacement.** `scripts/archive/data/apply_ma_direction_refinement.py:24-27` still imports `scripts.archive.data.refine_ma_medium_tier`, while this PR deletes `scripts/archive/data/refine_ma_medium_tier.py`. Importing the apply step raises `ModuleNotFoundError`, which causes the test-collection failures. Restore or deliberately relocate the refiner, update all imports/tests/docs, and update its obsolete Form D behavior rather than restoring it unchanged.

2. **P1 — Form-D-only should be excluded fail-closed, but not universally labeled acquirer-side.** `scripts/data/detect_sbir_ma_events.py:202-230,394-396` asserts every Form D filer is the acquirer and persists `non_exit_reason="acquirer_side"`. [Form D Item 10](https://www.sec.gov/info/edgar/forms/formd.pdf) does not encode transaction direction, and this PR’s own memo reports corporate reorganizations in its clarification sample. Use a neutral reason such as `form_d_only_no_target_evidence`; promote a directional label only from filing-level evidence.

3. **P1 — Refuted and context-only rows still enter the exit artifact.** `EFTS_SIGNAL_KEYS` at `detect_sbir_ma_events.py:193-219` treats any EFTS flag—including proxy and ownership-only context—as sufficient to retain a row. Separately, `apply_ma_direction_refinement.py:115-130` lowers confidence for `not_target`, comparator, or incomplete rows but still appends all of them to the same artifact. Presence-only consumers then count those rows as exits. Route rows after directional refinement, or require every consumer to enforce one explicit accepted-target-evidence contract. `docs/research-questions.md` should not claim this routing is already complete.

4. **P1 — The selected exit date is not anchored to the M&A evidence.** `extract_efts_signals` at `detect_sbir_ma_events.py:113-133` attaches the aggregate profile-wide `latest_mention_date`, and `merge_events` at `:166-180` makes it win over Form D whenever present. Upstream aggregates mention types and maximum dates independently, so that date can belong to another filing or mention type. When it is absent, the direction-unresolved Form D filing date remains as the exit date. Use an accession/type-anchored date or leave the event date unavailable, and add a mixed-filing adversarial test before any FY reporter consumes this artifact.

5. **P1 — The detector regresses current-main scoring safeguards.** `detect_sbir_ma_events.py:50-63` and `sample_form_d_join_adjudication.py:98-110` accept `tier == "high"` without requiring a rule version and document the retired person-or-ZIP-v1 behavior. Current `main` requires `corroborated-person-v2` and rejects missing/stale versions. Conflict resolution must retain `require_form_d_tier_rule`, force a complete rescore, persist the version in derived rows, and retain the fail-closed tests.

6. **P1 — Record-level pooled confidence cannot validate a particular combination filing.** The producer pools persons, locations, dates, and potentially multiple CIKs into one company-record confidence, while the detector applies that tier to every combination offering. The adjudication worksheet can consequently display one record-level CIK/confidence beside another CIK’s filing. Score/select at offering-and-CIK grain, emit the actual accession/CIK, and quarantine cross-filing or cross-CIK evidence. #683’s accession-level identity review cannot retroactively validate these pooled rows.

7. **P1 — The legacy input parser also disagrees with the new Form D protocol.** Shared `form_d_scoring._bool` maps XML Schema `1` to `False`, whereas #683 correctly treats it as true. Fix this as a shared tri-state source primitive and regenerate/re-audit the legacy `form_d_details` input before comparing or publishing counts.

8. **P2 — The memo republishes incompatible provisional results without a final supporting artifact.** `docs/research/sbir-ma-exit-analysis.md:46-56` reports high+medium as 1,887, while `:58-80` reports 1,774 and 5.1%; `:88-109` says final regeneration and adjudication remain pending. Current `main` intentionally suppresses all Form-D-dependent totals until the v2, filing-grain, amendment-collapse, and human-review gate closes. Preserve that suppression and add only the qualitative correction until one fingerprinted final-head materialization exists.

Verification: 34 isolated detector tests pass, but full collection reproduces two missing-module errors and GitHub reports all four Fast Tests shards failing. Recommendation: rebase onto current `main`, preserve its rule-version and retired-result safeguards, then reimplement the direction/routing/date changes with filing-grain tests.

hollomancer:
## What is still needed — do not land, do not rebase wholesale

Draft. Conflicts with `main`. All four Fast Tests shards fail. Head `eefa0f8c`. Title already says do not land.

The join defect is real (detector ignored `match_confidence.tier`; Form D Item 10 is not an exit). The branch around that fix is not. Repair on current `main`; do not try to salvage this head.

**Repair on current `main`:**

1. **Unbreak CI.** The rebase deleted `scripts/archive/data/refine_ma_medium_tier.py` while `apply_ma_direction_refinement.py` still imports it. Restore or relocate it; don’t restore the obsolete Form D grading.

2. **Keep `main`’s scoring contract.** Current `main` requires `corroborated-person-v2` and `require_form_d_tier_rule`. This branch still accepts bare `tier == "high"` (person-or-ZIP-v1). A naive rebase would drop that safeguard.

3. **Form-D-only is “no target evidence,” not “acquirer.”** Census: of 353 non-empty clarifications, 302 acquirer-side, 32 reorganizations, 4 target-side. Label `form_d_only_no_target_evidence`; directional labels only from filing-level text.

4. **Presence-only consumers still count demoted rows as exits.** `apply_ma_direction_refinement.py` on `main` appends `not_target` / comparator rows. `EFTS_SIGNAL_KEYS` also keeps proxy/ownership context. Route those out of the exit artifact, or make every consumer require accepted target-side evidence. `docs/research-questions.md` currently claims routing that does not exist.

5. **Date and grain.** Exit date is a profile-wide `latest_mention_date`, not accession/type-anchored. Confidence is pooled at company-record grain, then applied to every combination filing. Score at offering-and-CIK grain.

6. **Memo.** Two incompatible “corrected” tables (1,887 vs 1,774 / 5.1%). `main` already suppresses Form-D-dependent totals until the v2 gate. Keep that suppression until one fingerprinted regeneration exists.

**Do not regenerate `sbir_ma_events.jsonl` yet.** The 81-row worksheet is still `UNREVIEWED`. A full re-run also moves medium by ~490 events for an unrelated pre-#286 discrepancy, and needs live EDGAR (~1,200–1,600 fetches). Adjudicate first, regenerate once.

Preprint #682 still cites 1,197 high / 8.1%. This PR must not land until adjudication + regeneration + consumer filter are one deliberate operation.

The XML boolean `1`/`true` split is the only hard coupling with #683. Fix it once as a shared primitive. #683 can then land as a pre-run protocol; this PR should be re-implemented on current `main` around that primitive, not merged from this head.

hollomancer:
## Closing: the defensible changes are extracted into #735

This branch is `CONFLICTING`, failing four tests, and carrying nine live threads against a base that has moved twelve commits. The deciding factor is not the count — it is that a **Blocking** thread established `docs/research-questions.md`, the repository's north star and scope gate, now describes behaviour this branch no longer contains: the commit that wrote `not_target` and comparator rows (`491eb77d`) operated on `apply_ma_medium_refinement.py`, which the 2026-09-12 rebase dropped.

Rebasing would mean resolving four conflicts and nine threads to preserve history whose main content is superseded reasoning.

### What was recovered, in #735

- **Tier filter** — `extract_form_d_signals` drops records below `KEEP_MATCH_TIER`, with a printed drop count so the filter is visible.
- **Form D alone is no longer `high`** — Item 10 is filed by the acquirer, so grading it high inflated the exit population with rows recording purchases by the SBIR firm.
- **Sibling artifact** — acquirer-side-only rows go to `data/sbir_ma_non_exit.jsonl` with `non_exit_reason='acquirer_side'`, because demoting them was not enough when several consumers read row presence and never read confidence.

### One thread settled rather than inherited

This branch's `EFTS_SIGNAL_KEYS` included `efts_ma_proxy` and `efts_ownership_active`, so a single SC 13D or merger-proxy mention kept an otherwise acquirer-side row in the exit artifact — and `test_form_d_with_efts_evidence_is_not_excluded` encoded that rule. The signal memo grades both Low and neither is target-side evidence. #735 restricts the set to `subsidiary`, `ma_definitive`, and `acquisition_text`.

### Not carried over

The move out of `scripts/archive/`, which is a separate question the archive README has its own view on. And the docstring-drift and research-questions threads, which describe code that is no longer here — #735 starts from current `main`, so they do not arise.

Closing in favour of #735.

PR #711, opened 2026-09-10
Require a validation design above reproducible, and adversarial cases for matchers

hollomancer:
PR 1 of 3 from `docs/superpowers/specs/2026-09-10-verification-practices-design.md`.

Between 2026-09-08 and 2026-09-10, ten defects were found in the M&A detection pipeline. Seven would have been caught by tests that did not exist. Two were caught by tests that existed and asserted the defect. This PR addresses the design-side half.

## What lands

**`validation_design` block in the study-manifest schema.** Four required fields — addressable population, expected yield, decision threshold, and the derivation of that threshold. `scripts/ci/validate_study_manifests.py` requires it when `evidence_status` is `validated` or `citable`, and does not require it below that: a census has no pass/fail threshold, and forcing one produces a paragraph reading "n/a".

The defect it targets is concrete. `studies/ma-discovery-recall` failed three times against a recall floor of 10 stated as a bare count and never normalised to a shrinking eligible pool (342, then 503, then 307). At the observed detection rate that floor passes about 30% of the time even when the method performs exactly as measured.

**Adversarial-case convention in `CLAUDE.md`.** Two kinds of function need tests that try to break them, because a false positive from either reads as evidence rather than as a crash: role assignment from free text, and cross-population entity matching. For every rule that can return a positive, write an input where it must not fire.

Deliberately narrow. A field normaliser like `_normalize_state` cannot produce a false positive that reads as evidence and is excluded, with the reasoning stated.

## Verification

- 89 tests pass across `tests/unit/quality/` and `tests/unit/scripts/ci/`
- `make lint-boundaries` and `make docs-check` clean
- All three existing study manifests still validate — each is `exploratory` or `reproducible`, so none newly fails
- **Mutation-checked:** setting `_REQUIRES_VALIDATION_DESIGN` to `frozenset()` fails exactly 2 tests. Reproduced independently by the reviewer rather than taken from the implementer report.

That last check matters here more than usual. This repository has a documented history of tests passing for the wrong reason — one asserted `main() == 1` and passed with the gate under test disabled, because an unrelated path also exits 1.

## Two errors the review caught, both mine

The spec cited `tests/unit/scripts/test_refine_ma_direction.py` as an existing worked example; it arrives with #705 and is now labelled as such. And it claimed `test_press_wire.py` holds 28 tests — it holds 25, and I never counted. Both corrected, and the `0/18` precision figure now names the artifact backing it (`data/enriched_sbir_ma_events.jsonl`, 18 rows carrying `press_wire_signals`) so a reader can check it.

Worth stating plainly: a convention document about unverified assumptions shipped with an unverified number in it, for one review round.

## Not in this PR

PR 2 adds artifact-boundary invariant tests. PR 3 audits the six functions in the two classes above, starting with `identity/company_names.py` — `primitives` tier, 62 tests, one negative assertion.

Neither of the three reaches the largest defect of the ten: that Form D Item 10 marks the acquirer rather than the target. The test category that would is ground-truth calibration, and every label set in this repository labels the detector own output and inherits its selection. Building an unbiased acquisition list is a prerequisite, not a test to write. Recorded in the spec, out of scope here.


hollomancer:
## All three findings fixed — `0f96a349`

**The whitespace hole was real and is closed.** `Field(min_length=1)` accepted `"   "`, so a manifest could declare `evidence_status: validated` and satisfy the entire gate with four whitespace strings. Verified by construction before fixing, and verified closed after:

```
spaces   rejected      tab      rejected
newline  rejected      mixed    rejected
legitimate values still accepted
```

Mutation-checked: removing the validator fails all four new tests with `DID NOT RAISE ValidationError`.

That finding matters more than its size. This PR adds a gate whose stated purpose is to stop a study satisfying a check while stating nothing — and shipped it satisfiable by whitespace. Thank you for catching it.

**Both citation findings were the same error, and it was mine twice.** `studies/ma-discovery-recall/design.md` is not in this repository; it lives on `fix/form-d-match-tier-join` (PR #705, unmerged). The figures it backs — eligible pools of 342, 503, 307 and the ~30% pass rate — are real, but unverifiable from this branch.

Both are now attributed to PR #705 rather than to a path that does not resolve, matching the treatment already applied to the `test_refine_ma_direction.py` reference in `CLAUDE.md`.

Worth recording that this is the second round of the same mistake: the earlier review caught the identical cross-branch citation in `CLAUDE.md`, and I reintroduced it in two more places. The rule I should have applied — every path a document cites must resolve on the branch that ships it — is now applied uniformly.

## State

93 passed, 3 skipped across `tests/unit/quality/` and `tests/unit/scripts/ci/`. `make lint-boundaries` and `make docs-check` clean.

Issue #714, opened 2026-09-11
form_d_scoring: a 0.70 fuzzy person-name match alone promotes a filing to high tier


hollomancer:
An adversarial audit of `sbir_etl/enrichers/sec_edgar/form_d_scoring.py` found one defect: the tier rule promotes a Form D match to `high` on a fuzzy person-name score of 0.70, which ordinary name variation clears.

## The rule

`form_d_scoring.py:339`:

```python
ps = person_score if person_score is not None else 0.5
ads = address_score if address_score is not None else 0.5
if ps >= 0.7 or ads >= 1.0:
    tier = "high"
```

`person_score` is the best fuzzy match between an SBIR award PI name and a name on the Form D filing (`:282`, `fuzz.token_set_ratio`). A single such match at 0.70 is sufficient for `high` on its own.

## Why 0.70 is too low for person names

Distinct people clear it routinely:

| PI name | officer name | score | tier |
|---|---|---:|---|
| Robert Chen | Roberta Chen | 0.96 | **high** |
| John Smith | Jonathan Smith | 0.83 | **high** |
| M. Patel | Mark Patel | 0.78 | **high** |
| David Kim | Daniel Kim | 0.74 | **high** |

These are not adversarial constructions. They are a nickname, an initial, a gendered variant, and two common given names sharing a surname — all normal in SEC filings and SBIR award records.

The audit also found `ADELPHI TECHNOLOGY` vs `ADEPT TECHNOLOGY` reaching `high` through this path at 0.8824.

## Why it matters

`match_confidence.tier` is the gate several consumers use to decide whether a Form D filing belongs to an SBIR firm at all. PR #705 filters the M&A detector to `tier == "high"` precisely because that field is supposed to be the trustworthy one. If `high` can be reached by one fuzzy person-name hit, that filter admits what it was added to exclude.

A related defect in `sec_edgar/enricher.py` is already filed as #713: distinct firms resolving to the same CIK at threshold 90. Same shape — a threshold set where ordinary variation clears it.

## What is NOT wrong

**The rest of the scorer.** 8 of 9 adversarial pairs land correctly on `low`. The multi-signal design is sound; one threshold in it is not.

**`company_fuzzy_matcher.py`**, audited alongside this: no defect. 0 of 9 pairs reach `fuzzy-auto` at threshold 90, closest miss 89.

**`identity/company_names.py`**, audited earlier: no defect. 0 of 9 pairs reach 0.95, max 0.882.

[...]

Issue #713, opened 2026-09-11
sec_edgar enricher resolves distinct firms to the same CIK at threshold 90


hollomancer:
An adversarial audit of `company_name_similarity` found the primitive sound and its three consumers in `sbir_etl/enrichers/sec_edgar/enricher.py` defective. `_resolve_cik` scores **exactly 1.0** for pairs that are different companies.

## Reproduction

```python
from sbir_etl.identity import CompanyNameProfile
from sbir_etl.identity.company_names import company_name_similarity, CompanyNameMetric

for a, b in [("3D Control Systems, Inc.", "3D SYSTEMS CORP"),
             ("ADT Pharmaceuticals", "ADT Inc."),
             ("COMPASS SYSTEMS", "Compass, Inc.")]:
    print(company_name_similarity(a, b, metric=CompanyNameMetric.TOKEN_SET,
                                  profile=CompanyNameProfile.SEC_EDGAR_V1))
# 1.0
# 1.0
# 1.0
```

`_resolve_cik` accepts at 90. All three resolve to the wrong SEC entity.

**This is not hypothetical.** `3D Control Systems, Inc.` joined to filer `3D SYSTEMS CORP` (CIK 0000910638, a large public company) is a real attribution error found in production Form D data on 2026-09-08, and it graded `high` confidence. This is its mechanism.

## The three consumers

| function | threshold | pairs crossing |
|---|---:|---:|
| `_resolve_cik` (~:731) | 90 | 3 at **1.0** — wrong-CIK resolution |
| `_search_form_d_filings` (~:615) | 85 | 4 |
| `_search_filing_mentions_filtered` (~:460) | 80 | 4 |

The third fails in the opposite direction: it silently drops legitimate M&A mentions as "self-filings" when an unrelated firm scores over 80.

## Root cause

Two compounding choices, both at `enricher.py:731-745`:

```python
score = max(
    company_name_similarity(query_clean, entity_clean, metric=CompanyNameMetric.TOKEN_SET),
    company_name_similarity(query_clean, entity_clean, metric=CompanyNameMetric.RATIO),
) * 100.0
```

[...]

Issue #708, opened 2026-09-09
press_wire _match_company does unanchored substring matching — all matches are false positives


hollomancer:
Every press-wire match in the shipped artifact is a false positive. The matcher tests whether a normalized company name appears **anywhere** in a release title or summary, with no word boundary.

## Evidence

`data/enriched_sbir_ma_events.jsonl` carries 18 matched releases. All 18 are spurious. The actual matched spans:

| watchlist firm | normalized | matched inside |
|---|---|---|
| `BAL` | `bal` | "(nyse: rcl), a glo**bal** vacation leader" |
| `APP, Inc` | `app` | "irr of **app**roximately 54%" |
| `ATI, INC.` | `ati` | "sliders to kiosks n**ati**onwide" |
| `DRI` | `dri` | "alexan**dri**a, va., april 27, 2026" |
| `Ert` | `ert` | "hardwareintegri**ert**e ai-box-lösung" |
| `NFERENCE, INC.` | `nference` | "natural resources co**nference** at uh law" |

The headlines are Kenyan sports betting, MADD Canada awards, Slovak pharma litigation, and securities class-action alerts. **Precision 0/18.**

This is structural. **391 of 3,980 normalized watchlist names are five characters or fewer** — `act`, `apt`, `app`, `abi`, `aes`, `ares`, `ames`. Short names match inside ordinary English words constantly.

## Where it bites

`sbir_etl/enrichers/press_wire.py:277-286` (`_match_company`), reached from:

- `sbir_etl/reporting/weekly/enrichment.py:379` (`poll_press_wire`) -> weekly digests. **This is live.**
- `sbir_etl/enrichers/ma_discovery/press.py` -> being deprecated separately for an unrelated tier reason.

So the weekly digest has been surfacing unrelated press releases as awardee news.

## Why fixing the feeds first would make it worse

Two of three feeds currently fail (no descriptive User-Agent; PRNewswire 301s, GlobeNewsWire times out). A fix exists on `fix/press-wire-feeds` and is **deliberately held**: restoring the feeds means more items scanned per poll at a precision of zero. More working feeds, more false hits in the digest.

Fix this first, then land the feed repair.

## Second defect in the same function

[...]

PR #715, opened 2026-09-11
Adversarial audit of the company-matching functions

hollomancer:
PR 3 of 3 from `docs/superpowers/specs/2026-09-10-verification-practices-design.md`.

Tests only. No source file changed. The deliverable is what the tests found.

## Method

For every function that can return a positive match, an input where it must **not** fire. The table is nine company pairs — **eight are real attribution errors from production Form D data**, found 2026-09-08. `3D Control Systems, Inc.` joined to `3D SYSTEMS CORP`. `Linked, Inc.` to `LINKEDIN CORP`. `BAL` to `BALL CORP`.

## Result: the primitives are sound, the thresholds are not

| audited | verdict |
|---|---|
| `identity/company_names.py` | **no defect** — 0/9 reach 0.95, max 0.882 |
| `company_fuzzy_matcher.py` | **no defect** — 0/9 reach `fuzzy-auto` at 90, closest miss 89 |
| `form_d_scoring.py` | **defect** — 1 pair reaches `high` on a name collision → #714 |

And two consumers of the sound primitive:

| consumer | threshold | result |
|---|---:|---|
| `enricher._resolve_cik` | 90 | **3 pairs score exactly 1.0** → #713 |
| `enricher._search_form_d_filings` | 85 | 4 pairs cross |
| `enricher._search_filing_mentions_filtered` | 80 | 4 cross, dropping real M&A mentions as self-filings |

`identity/company_names.py` is `primitives` tier — contract "comprehensive tests" — and had 62 tests with one negative assertion. The audit confirms it earns that tier. Its callers do not.

## Two issues filed

**#713** — `_resolve_cik` scores `3D Control Systems, Inc.` against `3D SYSTEMS CORP` at **1.0** and accepts at 90. `max(TOKEN_SET, RATIO)` takes the more permissive metric, and `TOKEN_SET` scores subset-containment near 100 by design. This is the mechanism behind the production join.

**#714** — `form_d_scoring.py:339` promotes to `high` on a fuzzy person-name match of 0.70. `Robert Chen` vs `Roberta Chen` scores 0.96; `John Smith` vs `Jonathan Smith` 0.83. Ordinary name variation clears it, and PR #705 filters on `tier == "high"` precisely because that field is meant to be trustworthy.

## What the audit deliberately did not do

**It did not lower any matcher to make a test pass.** Each agent was instructed that a high score is not automatically a defect — these functions return scores, and a caller threshold turns a score into a match. Editing a matcher to reduce a score repairs the ruler, not the thing measured. That instruction is why the findings landed on thresholds rather than on the primitives.

**It did not fix the two consumers.** Rewriting thresholds in a live SEC enricher is its own change with its own blast radius, particularly while `enricher.py` feeds the Form D join still mid-correction on #705.

## Scope, narrowed by evidence

The plan named six functions. Two (`ucc/matcher.py`, `sec_edgar/enricher.py`) route through the primitive, so #713 covers their threshold risk without re-deriving it. One (`press_wire._match_company`) is already measured at 0/18 precision under #708. The remaining two are what this PR audits.

## Verification

1,108 tests pass across `tests/unit/identity/` and `tests/unit/enrichers/`. `make lint-boundaries` clean.

### DIFF
--- a/tests/unit/enrichers/test_company_fuzzy_matcher.py
+++ b/tests/unit/enrichers/test_company_fuzzy_matcher.py
@@ -8,6 +8,7 @@ import pytest
 from sbir_etl.enrichers.company_fuzzy_matcher import enrich_awards_with_companies
 from sbir_etl.utils.text_normalization import normalize_company_name
 from tests.factories import DataFrameBuilder
+from tests.unit.identity.test_company_names import MUST_NOT_MATCH
 
 
 pytestmark = pytest.mark.fast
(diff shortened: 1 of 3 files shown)


hollomancer:
## Positive counterpart added, and it produced the finding that matters

Two commits since the initial audit. The audit was negative-only — nine pairs that must not match — which meant every matcher could be "fixed" by rejecting everything and no test would catch it.

`MUST_MATCH` now sits alongside it: 14 real pairs from DUNS values carrying more than one company rendering in `award_data.csv`. One legal entity, written several ways.

**Four were excluded as renames**, not variants — `ALTRAZEAL LIFE SCIENCES` / `Uluru Inc.`, `Joule Therapeutics` / `TARN BIOSCIENCES`, `RGBSI AEROSPACE` / `Rapid Global Business Solutions`. A name matcher cannot connect those, and demanding it would force thresholds down, reintroducing #713 and #714 exactly.

## No threshold separates the two tables

Scored with `RECIPIENT_V1`, which every live consumer passes:

| metric / combinator | min(MUST_MATCH) | max(MUST_NOT_MATCH) | separates? |
|---|---:|---:|---|
| RATIO | 0.526 | 0.882 | no |
| TOKEN_SET | 0.816 | **1.000** | no |
| TOKEN_SORT | 0.526 | 0.882 | no |
| JARO_WINKLER | 0.867 | 0.950 | no |
| `min(TOKEN_SET, RATIO)` | 0.526 | 0.882 | no |
| `min(TOKEN_SET, JARO_WINKLER)` | 0.816 | 0.893 | no |

Computed twice independently. The worst offender is `3D Control Systems, Inc.` vs `3D SYSTEMS CORP` at **1.000** — a real production error — outranking `WEINBERG MEDICAL PHYSICS, LLC` / `Weinberg Medical Holdings` at 0.816, a DUNS-confirmed true match.

**Company-name similarity alone cannot decide firm identity on this data.**

`test_a_must_not_match_pair_outscores_a_must_match_pair` pins that directly. It is the opposite polarity of every other test here, and its docstring says what to check before anyone reads a future failure as progress.

## This changes the fix for both open issues

**#713** — I had suggested swapping `max(TOKEN_SET, RATIO)` for `min()`. The table says that narrows the overlap without removing it. CIK resolution needs corroborating evidence — state, address, officer names — not a better metric. Corrected on the issue.

**#714** — the conjunction fix is confirmed as the right shape. Person names are worse than company names: `Robert Chen` and `Roberta Chen` differ by one character and are different people.

## One correction made along the way

The first version of this table called `company_name_similarity` without a profile, so no normalization ran, a pure case difference scored 0.188, and the conclusion was blamed on case sensitivity. That was an artifact of the call — every consumer passes a profile. The headline conclusion survived; the explanation did not.

111 tests pass. `make lint-boundaries` clean.

PR #719, opened 2026-09-12
Repair the press-wire feeds, now that the matcher is fixed

hollomancer:
Stacked on #718. Review and merge that first.

This fix was written on 2026-09-10 and **deliberately held**. Restoring broken feeds while the matcher had 0/18 precision would have scanned more items at zero precision — making the weekly digest worse, not better. #718 fixes the matcher, so this can land.

## What was broken

`BaseAsyncAPIClient._build_headers` requests `application/json` and sends a bare `SBIR-Analytics/<version>` agent. Both are wrong for a news feed:

- **PRNewswire** answered 301 and the client did not follow it
- **GlobeNewsWire** timed out
- **BusinessWire** returned a valid RSS envelope whose description reads *"The channel you requested is unavailable due to an error during request processing"* — so it parsed as a well-formed feed with zero items, and the failure was invisible

Two of three feeds had been dead, and `poll` reported success either way.

## Three changes

1. **Override `_build_headers`** with a descriptive agent and an RSS/Atom `Accept`. Headers set on the `httpx.AsyncClient` constructor never applied, because `_request_raw` rebuilds them per request — the override has to live in `_build_headers`, which is the layer that takes effect.
2. **Drop BusinessWire.** Its `?rss=` token is revoked. A replacement URL is needed before it can return.
3. **Raise when every feed fails.** A total fetch failure produced the same empty list as "nothing in the news matched".

Verified against the live feeds: **0 items before, 40 after** — 20 from each surviving feed.

## Two superseded tests

- One asserted an empty list when the only feed 500s. That is the defect: nothing observed is not the same as observing nothing.
- One expected BusinessWire in the defaults.

## Verification

1,015 tests pass with #718s matcher fix underneath. `make lint-boundaries` clean.

### DIFF
--- a/sbir_etl/enrichers/press_wire.py
+++ b/sbir_etl/enrichers/press_wire.py
@@ -1,7 +1,8 @@
 """Press wire RSS/Atom feed client for SBIR awardee news monitoring.
 
-Polls RSS/Atom feeds from PR Newswire, BusinessWire, and GlobeNewsWire
-for press releases mentioning known SBIR awardee companies. Designed
+Polls RSS/Atom feeds from PR Newswire and GlobeNewsWire for press
+releases mentioning known SBIR awardee companies. BusinessWire was
+dropped on 2026-09-09; see the note on ``FEEDS`` below. Designed
 as a leading-indicator source for commercialization events (contract
 wins, acquisitions, product launches, partnerships) that appear in
 press releases weeks/months before they surface in USAspending or FPDS.
--- a/sbir_etl/enrichers/press_wire.py
+++ b/sbir_etl/enrichers/press_wire.py
@@ -50,10 +50,20 @@ from sbir_etl.enrichers.rate_limiting import RateLimiter
 from sbir_etl.exceptions import APIError
 from sbir_etl.identity import CompanyNameProfile, normalize_company_name
 
+DEFAULT_HEADERS: dict[str, str] = {
+    # Without a descriptive agent PRNewswire 301s and GlobeNewsWire times out,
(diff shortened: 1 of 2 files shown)


hollomancer:
## Rebuilt onto `main`; the stale copy of #718 is gone

Retargeted from the deleted `fix/press-wire-word-boundary` to `main`, then rebuilt rather than merged. The merge was not safe.

**#718 was rewritten before it merged**, so `24e411a9` never became an ancestor of `main`. This branch carried that stale copy of the same work, and git saw two independent additions of the same field rather than one. Merging as-is produced a **duplicated `matched_in` field** in the `PressRelease` dataclass and reverted both review fixes made on #718 before merge:

- `matched_in` moved back ahead of `content_hash`, re-breaking positional compatibility for the public API.
- `assert matches[0].matched_in == "summary"` reverted to the form that accepts every location.

Worth noting `git merge-tree` reported this as clean. The legacy three-argument form prints conflicts to stdout but always exits 0, so it is not a predicate — an actual merge in a scratch worktree is.

**What is here now:** `main` + only the feeds commit (`cb8129d8`), plus one docstring fix. Both #718 review fixes verified intact.

Verification:
- 6,698 unit tests pass; `ruff check` and `ruff format --check` clean.
- Both feeds live: 40 items, 20 PRNewswire / 20 GlobeNewsWire.

**Still open from review:** `poll_all_unfiltered()` does not apply the total-failure check, so it still returns `[]` when every feed is unavailable. That is a behavioral change needing its own test, so I left it rather than fold it into a retarget.

PR #722, opened 2026-09-13
Anchor ATC reviewer hours in Gallo 2019 and record the price signal

hollomancer:
Stacked on #720 (`study/allocation-transaction-costs`) at `33fcb0a0`. Merge #720 first; this PR is the hour-evidence follow-up only.

## What this is

The parent study's `reviewer_hours` note said hours per review are not published as a time-use survey. That is no longer true.

Gallo, Sullivan and Glisson (2019) is a published US measurement: 874 AIBS panel reviewers reported 4.5 ± 0.08 and 5.0 ± 0.08 hours of pre-meeting time per assigned application. The existing 2/4/8 grid already brackets that, so **no computed numbers change**. A `reviewer_hour_citations` block records the measurement, its pre-meeting-only scope, and that it describes panel review rather than NSF ad hoc review.

Also records published 2026 consultant list prices for outsourced SBIR proposal preparation as an indirect, non-citable market signal on applicant hours (roughly 33–120 outsourced hours implied). That signal is barred from narrowing the 20–200 hour applicant grid.

Adds two limitations: the bimodal applicant population SBA reports, and the scope of the reviewer survey.

## What does not change

- The 2/4/8 reviewer-hour grid.
- The 20–200 applicant-hour grid.
- Any computed series, break-even, or generator code.
- Evidence status: still `reproducible`, not approved for citation.

## Files

- `studies/allocation-transaction-costs/assumptions.yaml` — citation block, corrected reviewer-hours note, consultant-price caveat on the applicant-hour note.
- `studies/allocation-transaction-costs/sources.yaml` — `aibs_gallo_2019` and `sbir_consultant_list_prices_2026`.
- `studies/allocation-transaction-costs/study.yaml` — two new limitations.

## Type of Change

- [x] Documentation update

## Versioning Impact

- [x] No release impact (internal or unreleased work)

## Testing

YAML-only. No generator or test changes. Parent #720's unit tests and study-manifest checks still apply; this commit does not alter any computed output.

## Checklist

- [x] Self-review of the three study YAML files
- [x] Grid values left unchanged; citations recorded, not used to retune
- [x] Consultant list prices marked non-citable and barred from narrowing the grid

PR #723, opened 2026-09-13
Add the NSF 2021 Merit Review Survey as a second hour anchor

hollomancer:
Stacked on #722 (`study/atc-hour-evidence`) at `47dfa945`. Merge #720, then #722, then this. This PR is the second hour anchor only.

## Why a second anchor

The reviewer citation added in #722 measures NIH-style panel review, and it says so explicitly — it does not cover NSF review. That left the NSF side of the comparison unanchored.

NSF's own **2021 Merit Review Survey** covers it. Reviewers reported an average of **7 hours** reading, writing, and submitting their most recent review, excluding panel travel and participation. That is the same scope as the AIBS pre-meeting figure, so the two are comparable rather than double-counting.

Both sit inside the existing 2/4/8 grid, so **no computed numbers change**.

## The applicant-side anchor matters more

The same survey gives a university-side applicant measurement that does not depend on the FDP percentage-of-time transfer:

- **126 hours** to prepare the most recent proposal
- regression-adjusted directorate estimates of **78–104 hours**

This is a survey measurement, not a Paperwork Reduction Act estimate, so it is admissible under `forbidden_hour_sources`. It sits inside the 80–200 grid and **corroborates** it rather than redefining it.

## Scope

| | |
|---|---|
| Files | `assumptions.yaml`, `sources.yaml` |
| Computed series changed | none |
| Grid changed | none |

Body written from the commit — the `PR3_atc-nsf-merit-review-survey.md` file named in the plan was not present on this machine.


hollomancer:
The exclusivity threads are fixed in #736. `reviewer_hours_note` now names both published measurements with their differing scopes (AIBS pre-meeting time on NIH-style panel review; NSF 2021 across NSF review methods) instead of referring to a single one, and the `study.yaml` limitation does the same. Separate PR because this branch is mid-stack — see the note on #722.

PR #725, opened 2026-09-13
Record that the SBIR-versus-R01 contrast is not a performer comparison

hollomancer:
Stacked on #724 (`study/atc-external-benchmark`) at `e5221787`. Last in the stack: merge #720 → #722 → #723 → #724 → this.

## The point

The difference between NIH SBIR Phase I and NIH R01-equivalent grants sums **at least four effects this design cannot separate**:

1. performer
2. award architecture
3. award size
4. research type

The estimand and permitted claims already refuse a directional ranking. This records **why** such a ranking would be *unidentified* rather than merely *underpowered* — so the refusal is not misread as caution about sample size. Those are different failures and they have different remedies; more data fixes one and not the other.

## Why an amendment and not an edit

`design.md` is frozen and pinned by SHA-256. This goes in an amendment log following the `form-d-fundraising` convention rather than editing the frozen bytes.

Revision 1 also names the designs that *would* identify the performer effect:

- same funding opportunity and same instrument
- STTR as an internal matched pair

and carries forward the rule that **neither university F&A nor a company indirect rate is a transaction-cost rate**.

## Scope

| | |
|---|---|
| Files | `amendments.md` (new), `study.yaml` |
| Mechanism, table, assumption, or computed series changed | **none** |

Body written from the commit — the `PR5_atc-identification-limits.md` file named in the plan was not present on this machine.


hollomancer:
All three threads are fixed in #736.

**Identification.** The design list is retitled to designs that make the performer contrast *more comparable*, with an explicit statement that none identifies a causal performer effect on its own — holding opportunity and instrument fixed removes mechanism variation but leaves topic, team, quality and award size differing, and selection into performer type is not random. The list is now a ranking of comparability, not identification strength, and each design needs its own matching assumptions.

**The company-indirect claim overstated the code.** `forbidden_hour_sources` covers `fa_rate` and `pra_estimate` only, no such label exists to be refused, and a caller could still pass one to `applicant_hours_from_source`. It is now marked as a rule for the next revision to implement and test.

**`amendments.md` is pinned** in `frozen_artifacts` at its post-edit digest.

PR #726, opened 2026-09-13
Redefine validated as a recorded result, not a cleared threshold

hollomancer:
Independent of the C4 stack (#720–#725). Based on `origin/main` at `493340ea` (v0.13.0). One commit, `4908c40a`.

## Why this exists

Nothing on main is above `reproducible`, and the one study that tried to cross (`ma-discovery-recall`) was blocked four times. Two of those blocks were process defects the discipline was right to catch. The other two were the shape of the gate:

1. An absolute recall floor of 10 over a population whose strict-eligible share fell 68% → 50% → 31% as capture proceeded.
2. A pass/fail reading that made a held-out 13/19 authorize either a claim or nothing.

The schema recorded a validation *design* but had no field for a validation *result*. Pass/fail was the only possible reading of `validated`. That is a structural defect, not a documentation gap.

## What changed

The rules lived in four places. All three fixes are enforced in `StudyManifest` where CI can see them; the prose in `studies/README.md`, `docs/steering/epistemic-tiers.md`, and the evidence-auditor instructions is brought into line.

### 1. Proportional floors

`ValidationDesign` gains `threshold_basis` — `proportion` or `count_on_frozen_population` — required at `validated` and above. A count basis requires a `frozen_population_artifact` that must also be pinned in `frozen_artifacts`. The exact failure that sank `ma-discovery-recall` is now a schema error rather than a study failure.

### 2. The result travels

New `validation_result` block, required at `validated` and above:

- `design_sha256`, `evaluated_on`, `metric`
- `numerator`, `denominator`
- interval bounds and method
- `threshold_met`, `confirmatory`, `post_hoc_analyses`

Coherence is checked: numerator ≤ denominator, point estimate inside the interval.

**`validated` means the preregistered test ran as written and its outcome is on the record; it does not mean the threshold was met. `citable` does.** A held-out 13/19 now authorizes exactly "13/19, Wilson 95% [0.46, 0.85]" — a reader can discount that where they couldn't discount "passed". A fair test the method failed becomes a legitimate `validated` end state and a reportable finding about the method.

### 3. Confirmatory vs. post-hoc, distinguished rather than the latter forbidden

`confirmatory: true` is required to promote, and `design_sha256` must match a frozen-artifact hash — so the "was the design in git before the run" check is now a field the auditor verifies from history, not a judgment. Post-hoc analyses are listed, reportable anywhere the study is, and no longer a BLOCK by themselves. Presenting one as confirmation still is. The auditor's stop condition now says explicitly that recording a failed validation with its interval is the required form of the record, not concealment.

## What is unchanged

Hashed designs, SHA-pinned inputs, blocking asset checks, the estimand, and the four-item evidence contract. `reproducible` is unchanged. All four live manifests load without modification because the new fields are optional below `validated`.

## One decision left for review

This changes what the word `validated` means to an outside reader of `docs/research-questions.md`. Nothing currently holds that rank, so there is no retroactive relabelling — but the `studies/README.md` table mapping inventory Status to `evidence_status` still points `Validated` at a rank that no longer implies a met threshold. That table is left alone deliberately. If the inventory word `Validated` should require `threshold_met`, that is a one-line change to `check_research_question_status.py`.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Schema gains constraints at `validated`/`citable`. Existing manifests below `validated` load unchanged. No live study holds `validated` or `citable`.

## Versioning Impact

Recorded under `[Unreleased]` in `CHANGELOG.md`. Package versions are not bumped here; the next release PR applies the increment.

## Files

| File | Role |
|---|---|
| `sbir_etl/quality/study_manifest.py` | `ThresholdBasis`, `ValidationResult`, validators |
| `tests/unit/quality/test_study_manifest.py` | 13 new tests (24 → 37) |
| `studies/README.md` | Rank definitions and schema |
| `docs/steering/epistemic-tiers.md` | The three rules |
| `.claude/agents/evidence-auditor.md` | Auditor checks and stop condition |
| `CHANGELOG.md` | Unreleased notes |

## Testing

Built and verified in a clean `git archive` of `origin/main` (avoids the local `.env` collection problem).

| Check | Result |
|---|---|
| `tests/unit/quality/test_study_manifest.py` | 37 passed (24 existing + 13 new) |
| Neighbouring suites | 463 passed, 3 skipped |
| `validate_study_manifests.py` | 4 repository manifests load unchanged |
| `check_research_question_status.py` | passed |
| CI scripts | passed |

The one neighbouring failure (`test_config_boundaries`) is an artefact of the archive not being a git checkout (`git ls-files` exits 128). It is not a regression in this change.

New tests cover: required result and `threshold_basis` at `validated`, omission still allowed at `reproducible`, post-hoc cannot promote, design hash must be a frozen artifact, missed threshold is valid at `validated` and rejected at `citable`, count threshold requires a pinned frozen population, interval coherence.

## Checklist

- [x] Code follows the style guidelines of this project
- [x] Self-review of schema, tests, and the three prose documents
- [x] Corresponding documentation updated in lockstep with the schema
- [x] Tests added that prove the new constraints
- [x] Existing manifests below `validated` still load


hollomancer (commented):
## Review

The diagnosis is right and the reframing is the valuable part. A schema that recorded a validation *design* with no field for a validation *result* forces pass/fail as the only reading of `validated`; that is a structural defect, not a documentation gap. Making a fair test the method failed into a legitimate `validated` end state is what would have let `ma-discovery-recall` terminate honestly instead of being blocked four times. The proportional-floor fix traces directly to a real failure — an absolute floor of 10 over a population whose eligible share fell 68% → 50% → 31%.

I verified the coherence checks and the `citable`-requires-`threshold_met` split hold. mypy passes on 265 files; all four boundary guards pass.

Two real findings, both now fixed in `15267c1b`.

### 1. `design_sha256` matched any frozen artifact

The body claimed this made the freeze check mechanical:

> `design_sha256` must match a frozen-artifact hash — so the "was the design in git before the run" check is now a field the auditor verifies from history, not a judgment.

It did not. Pointing `design_sha256` at `amendments.md`'s hash on the live `form-d-fundraising` manifest was accepted — the one file that by definition records changes made *after* the design froze.

`ValidationResult` now carries a required `design_path`, and the manifest requires the hash pinned at that path to equal `design_sha256`. The path is named rather than inferred because a study may pin several designs: `ma-discovery-recall` pins both `design.md` and `confirmatory-design.md`, so no naming convention would work.

A study can still name `amendments.md` as its `design_path`, but that is now an explicit declaration in the manifest rather than a hash coincidence — visible to a reader, which is the right residual.

### 2. `confirmatory` is self-asserted and the docstring said otherwise

> `confirmatory` is true only when `design_sha256` was fixed in git before the evaluated run

Nothing enforced that. There is no `git log` or `rev-list` call in the model or in `validate_study_manifests.py`, and this PR leaves the CI validator unchanged. It is a boolean the author sets.

I did not add a check — see the comment below on why that is a real design question rather than an oversight. What I did change is the promise: the docstring, `studies/README.md`, and the `CHANGELOG` now say the schema checks the hash pairing only, and that establishing the freeze predates `evaluated_on` is an auditor step against history. Worth noting the evidence-auditor instructions already said this correctly — "Verify the ordering from history, not from the manifest's say-so." The schema docstring was the one place overclaiming.

### Also fixed

`Lint, Types, and Guards` was failing, and the body did not mention it. The whole failure was `ruff format` on `tests/unit/quality/test_study_manifest.py`.

### Probed and deliberately not reported

A `[0.0, 1.0]` interval is accepted — correct, that is a real Wilson interval at small *n*. `threshold_basis: proportion` alongside a count-worded `decision_threshold` is accepted — no schema can catch that in prose.

Minor and left alone: `evaluated_on: 2099-01-01` is accepted.

### Verification

6,916 unit tests pass (41 in `test_study_manifest.py`, up from 37); ruff check and format clean across 999 files; mypy clean; `make lint-boundaries` clean; 4 manifests validate unchanged.

PR #729, opened 2026-09-13
Plan the ma-discovery-recall promotion #726 unblocked

hollomancer:
**Study 1 of 3.** Plan only — no manifest change in this PR.

`ma-discovery-recall` was blocked four times. Two blocks were process defects the discipline was right to catch; two were the shape of the gate, and #726 changed that shape.

## Everything needed is already frozen

| Requirement | Source | State |
|---|---|---|
| `threshold_basis: count_on_frozen_population` | the floor is a count | derivable |
| `threshold_value: 10` | `held-out-1501.yaml` `recall_floor` | pinned |
| `frozen_population_artifact` | `held-out-1501.yaml` (pins `events_sha256`) | already in `frozen_artifacts` |
| `design_path` | `held-out-1501.md` | already in `frozen_artifacts` |
| the result | `held-out-run-manifest.json` | recorded |

The number is **4 of 307 strict-eligible pairs**, Wilson 95% `[0.0051, 0.0330]`, against a floor of **10**.

## It turns on one question

`StudyManifest` refuses to promote a study whose `validation_result.confirmatory` is false. So the promotion reduces to: **can `confirmatory: true` be asserted honestly?**

For: `held-out-1501.md` was hashed and committed before the capture ran.

Against — and these are recorded by the people who ran it, in the run manifest's own `provenance_defects`:

1. **948 of 1000 cut pairs were captured by code whose state is not recoverable from git.** The design was frozen; the instrument was not.
2. 476 snippet rows are `unverified_empty`, captured before fault instrumentation existed.
3. **The frozen `held-out-1501.md` contains a false statement** — it says no Brave or LLM call had touched pairs 1501+ before two named commits, and 948 had.
4. The earlier 528-pair capture is superseded.

Defect 1 is about the instrument, not preregistration. **Defect 3 is about the design itself**, which is why this is an evidence-auditor call and not a schema question.

## The plan specifies both end states

**A — `validated`, threshold not met.** Authorizes `Computable`, not `Validated`. A four-times-blocked study becomes a finished, reportable finding about the method.

**B — `exploratory`, result under `post_hoc_analyses`.** #726 made post-hoc results reportable rather than a blocking defect, so this is a real end state, not a refusal.

Both beat the status quo: five open blockers and no recorded outcome.

## One thing this surfaced

`0023e204` — "hash held-out-1501 protocol before capture" — **is not reachable from `main`.** I squash-merged #699, so the protocol and results now arrive in one commit. The ordering is real and lives in that PR's history.

Any future mechanical `confirmatory` check must read the originating PR's commits, not `main`'s. A check reading `main` would fail on honest work.

## Not in scope

No rerun, no changed floor, no reinterpretation. The result is 4 against a floor of 10 either way — `replay_rules` is explicit that a second replay against a changed freeze is a new version, not a correction.


hollomancer:
Review findings:

1. Route A is not a defensible end state. The frozen `held-out-1501.md` says the validity rule changed after 948/1,000 pairs existed and that this cut “does not promote on any reading.” The replay also did not evaluate the precision or cost gates, and the capture used unrecoverable dirty code. A failed confirmatory test may now reach `validated`, but this run was neither fully executed as written nor confirmatory. Remove Route A and use the exploratory/post-hoc outcome.

2. The implementation steps omit a required `validation_design`. The current manifest has no such block, so adding only the count-threshold fields and `validation_result` would fail schema validation.

3. Route B uses the schema incorrectly. `post_hoc_analyses` is a `list[str]` inside `validation_result`; it cannot contain the structured 4/307 result. Add a complete `validation_result` with `confirmatory: false`, and reserve `post_hoc_analyses` for textual qualifications.

4. Reconcile the claim and lifecycle contracts. The current manifest forbids numerical recall claims, retains five materialization blockers, and leaves F2/A4 without a reserved rank. Update `permitted_claims`, limitations, blockers, `specs/ma-discovery-integration/`, and `specs/status.md` to record the failed cut and the chosen closure path.

hollomancer:
## Evidence audit: **BLOCK**. `confirmatory` cannot be true, and the reason is not the one this plan assumed.

Ran `evidence-auditor` on the single `confirmatory` question. It returned BLOCK, and the decisive fact was not in my framing at all.

### The pinned design postdates the run by ten hours

```
pinned design sha:  e7090020ef22b870   (study.yaml, held-out-1501.md)
first exists at:    9bcb6dc9  2026-09-09 07:30:41 -0400
replay as_of_utc:   2026-09-09T01:25:00Z = 2026-09-08 21:25 EDT
```

I verified this independently. `confirmatory` asserts the design bytes were fixed in git **before** the evaluated run. They were not — by ~10 hours, and ~2 days after the capture. That is false on the schema's own wording with no interpretation required, and the schema check passes it anyway because it only compares `design_sha256` against `frozen_artifacts` — precisely the check the docstring tells the auditor not to rely on.

No weighing of defect 3 is needed. The plan's central question was the wrong question.

### Three things I got wrong, corrected

1. **I classified defects 1 and 2 as "instrument" rather than "design" defects.** Both are design-compliance defects. The `unverified_empty` retry class was introduced at `f34266c5` (09-08 19:00) and applied to **476 of 1000 rows after those rows existed and after their emptiness was observable**. Retry eligibility must be pre-specified; here it was not, and re-querying changes the snippets freeze that determines the number.

2. **I wrote that the frozen bytes "still carry the original claim."** They do not. The pinned `e7090020` contains the full correction paragraph and the superseded hash. What I quoted was `provenance_defects[2]`, written in the present tense against an earlier version — so a pinned frozen artifact now misdescribes another pinned frozen artifact. That is a separate defect needing a superseded entry, not the one I described.

3. **`denominator: 307` is not preregistered.** `strict_eligible_pairs_in_cut` is a replay output. The preregistered gate is a count floor of 10 over the 1000-pair cut. "4 of 307, Wilson 95%" must be labelled post-hoc, not presented as the preregistered metric. It does not change `threshold_met` — 4 < 10 on any denominator — but I presented it as the confirmatory measurement and it is not.

### What the frozen design already says

> "This cut does not promote on any reading."

The note's own text, added while 948/1000 pairs were on disk. I was asking the auditor to overrule the design's explicit self-disqualification.

### Also: restoring the squashed ordering would not help

Pinning the pre-capture `b983466f` hash instead would make it worse, not better — that version authorizes no retry class, and the run re-queried 476 rows under one. That would assert confirmation against a design the execution violated. **The squash is a real provenance problem, but it is not what blocks this.**

### Correct end state: option B

`exploratory`, `confirmatory: false`, numbers recorded. Per the auditor, a full `validation_design` and `validation_result` validate fine at `exploratory` — the early return means no recorded detail is lost. Plus four `post_hoc_analyses` entries, a narrowed permitted claim, and an amendments entry for the squash lesson.

I will update the plan to option B rather than leaving it presented as a choice.

PR #730, opened 2026-09-13
Plan the transition-scoring promotion to reproducible

hollomancer:
**Study 2 of 3.** Plan only — no manifest change in this PR.

`transition-scoring` has the *implementation* half of `reproducible` already: five registered entry points, three frozen artifacts, materialization allowed, zero blockers, and a reproduced refit on record (828 rows, 138 positives, 101 firms).

It does not have the *inputs* half.

### 1. The scoring weights are deliberately mutable

The study's own limitations say `config/transition/detection.yaml` holds hand-set weights and **that live configuration is deliberately mutable**. Its permitted claim is telling — scoring is deterministic *"given a fixed detection configuration"* — and the configuration is not fixed. Same code, same inputs, different scores after any edit, with nothing to detect it.

**This is a decision, not a task.** Someone chose mutability, presumably for operational tuning. Three options in the plan:

- freeze the file outright — cheapest to verify, costliest operationally;
- **freeze a named profile** with live override, the pattern `sbir_etl.identity` already uses — recommended;
- emit the config digest with every score, making a divergent rerun detectable rather than prevented.

### 2. The fusion corpus is local-only and gitignored

Not SHA-enforced. It *is* regenerable from committed scripts and pinned manifests, so the fix is to record the command and expected digest and prove a clean-checkout rebuild matches. Until then "can be rerun" holds only on a machine that already has the file.

### 3. No stored benchmark report

The `>=85%` precision target isn't tied to a declared cut or label set. `CLAUDE.md` confirms it's measured only by a manual run against the S3 corpus and is not a CI gate.

This does **not** block `reproducible` — but it is the largest obstacle to any later `validated`, and gap 2 must close before it can even be attempted.

## A correction

My earlier assessment ranked this second-easiest and called it *"verification, not research"*, completable in days. **That underestimated it.** The mutable-config gap is a deliberate design decision with a real operational cost, and it has to be settled by a person before any manifest edit means anything.

## Not in scope

No attempt at `validated`. The natural design — the `>=85%` benchmark — can't run as a preregistered test until the corpus is pinned and the label set declared. Gap 3 is named for sequencing, not closed here.


hollomancer:
Review findings:

1. The plan conflates the composite `TransitionScorer`, the notice-fusion ranker, and the ≥85% retrospective backtest. They use different configurations and corpora. The operated and benchmark scoring paths do not consistently consume `config/transition/detection.yaml`, so freezing that file would not reproduce the outputs named by the study. Select one canonical scorer/result, then align the estimand, implementation references, artifacts, and actual runtime configuration with it.

2. The 828-row fusion corpus is not clean-checkout reproducible from the committed manifest. Its upstream local inputs and their hashes are missing, and an output digest cannot reconstruct absent bytes. Require recoverable source snapshots, hashes for every raw input, and the exact generation command and flags.

3. The notice-fusion corpus is not the corpus used by the ≥85% backtest. The active `transition-precision-benchmark` spec already owns the contracts/Phase-II benchmark corpus and its manifest. Keep that work there instead of coupling the two datasets.

4. Strengthen completion checks to verify the frame hash, refit metrics, coefficient hash, selected configuration provenance, and transition-polarity canary. Also correct the existing permitted claim that calls ≥85% an enforced change gate; only the fixture canary runs in PRs, while the full benchmark remains manual.

Lifecycle verdict: `GATED`. Trim this plan to one scoring system before promotion.

hollomancer:
## Data check: the corpus is regenerable locally, and gap 1 needs no data at all

Checked what actually blocks this promotion after the #731 investigation raised the same question.

### Present

| Input | State |
|---|---|
| `config/transition/detection.yaml` | **present** |
| `fusion_coefficients.json` | **present, committed** |
| `corpus.manifest.json`, `refit_ladder.json` | **present, frozen** |

### Absent but regenerable

The fusion corpus parquet is not on disk. It is genuinely regenerable, and the producers are committed:

- `scripts/phase3_benchmark/pull_gsa_archive.py` — streams per-fiscal-year archive CSVs from the public `falextracts` bucket over plain HTTPS. Its own docstring notes ListObjects is denied but GetObject is public, so **no AWS credentials are involved**. A ~1 GB yearly file reduces to a small firm-tagged parquet.
- `scripts/phase3_benchmark/build_notice_corpus.py` — assembles the frozen corpus from those pulls. Explicitly *"fully offline given the filtered pulls and the firm seed"*.
- `scripts/phase3_benchmark/refit_fusion.py` — the refit.

So gap 2 is bandwidth and runtime, not access. The verification target is already pinned: a clean rebuild must reproduce **828 rows / 138 positives / 101 firms**.

### Gap 1 is unblocked right now

The mutable-configuration decision needs **no data whatsoever**. `config/transition/detection.yaml` is present, and the choice — freeze the file, freeze a named profile with live override, or emit the config digest with every score — is a design call that can be settled and implemented today.

It will not complete the promotion on its own, but it is the gating decision, and unlike gap 2 it costs nothing to resolve.

### Suggested order

1. Settle gap 1 and implement it. No downloads.
2. Re-pull the archive and rebuild the corpus; confirm 828/138/101.
3. Then set `evidence_status: reproducible`.

Step 1 does not depend on step 2, so it should not wait for it.

hollomancer:
## Scope settled: this PR lands gap 1; the promotion moves to a branch off #732

Gap 1 is done and stands on its own. `Config.scoring_digest` hashes the `to_detector_config()` surface, the committed `detection.yaml` pins to `779cf5bc` (identical to the code `DEFAULTS` digest — they had not drifted), and 11 tests enforce it.

**`evidence_status` stays `exploratory` in this PR.** The promotion is deliberately not here.

### Why gap 2 changed character

It is no longer "download the corpus." The rebuild ran:

| Quantity | Frozen | Rebuild |
|---|---:|---:|
| rows | 828 | 822 |
| positives | 138 | 137 |
| firms | 101 | 100 |

The blocker is now that **there is no honest way to state whether that is a reproduction**. Promoting today would mean either declaring a tolerance that does not exist in the schema, or asserting a bit-exactness that is false against a live archive.

That gap is specced in #732. The promotion itself moves to a branch off it, so the two claims stay separate: the scoring config is identifiable now, and reproducibility waits on a definition of what reproduction means.

### What merging this gives you

- A divergent rerun becomes detectable rather than silent — the defect that kept the study below `reproducible` regardless of the corpus.
- A recorded decision, including that the named-profile approach I first recommended was wrong for this codebase: config reaches the scorer as a plain dict, so a profile registry would bind nothing a caller could not bypass.
- The finding that the digest has no structured home in `StudyManifest` — `frozen_artifacts` means exact bytes and `extra=\"forbid\"` rules out an ad hoc field — which is now open question 1 in #732.

PR #731, opened 2026-09-13
Plan the phase-iii-census promotion to validated

hollomancer:
**Study 3 of 3.** Plan only — no manifest change in this PR.

The cleanest study in the repository and **the only one with a realistic path to a *met* threshold**: frozen fixture-verified design, two frozen artifacts, four registered implementations, materialization allowed, zero blockers. Nothing is in its way. It has simply never been given a validation design.

It already has most of one.

## The half-built design

Its permitted claims record that under the frozen fixed-seed cross-firm completion-date placebo, **the actual frame exceeds the placebo frame on every complete-filter metric and in every cell.** That was preregistered, frozen, ran, and separated.

What it lacks is uncertainty. Its own first limitation:

> The materialized placebo is one non-uniform cyclic derangement, not an inferential permutation distribution; labeled validation remains unresolved.

One draw gives a direction, not a magnitude. `ValidationResult` requires an interval and an interval method, and a single derangement can supply neither.

## Two designs; one is reachable

**A — Permutation separation (recommended).** Run N fixed-seed derangements instead of one and record where the observed statistic falls in the empirical null. `threshold_basis: proportion`, numerator/denominator = draws exceeded over draws run, Wilson or exact binomial interval.

**No new data, no labels, no external corpus.** The placebo machinery is already frozen and already ran once; the change is running it N times.

**B — Labeled precision of the proxy.** The question a reader most wants answered, and unreachable: there is no adjudicated ground-truth label set, which is exactly what *"labeled validation remains unresolved"* records. Building one is a separate study.

## What `validated` would and would not mean

Under A it means **the observed separation from the frozen placebo is larger than chance, by a preregistered margin, with a stated interval.**

It does **not** mean the proxy identifies genuine statutory Phase III awards. The study's own limitation already says so, and promotion must not be read as retiring that sentence.

This matters more than usual: a met threshold here would be **the repository's first public `Validated` rank**, and that word will be read as "this measures Phase III correctly" unless the manifest says otherwise in its own voice. The plan puts that distinction into `permitted_claims` at promotion time rather than leaving it to be inferred.

## Sequencing note

Step 2 is *freeze the design before running it*, so `design_path`/`design_sha256` point at pinned bytes and `confirmatory: true` is defensible from the PR's commit order — which, per #729, is where that ordering has to be read once a PR is squash-merged.

## Why this is the one worth doing

The other two terminate in records rather than findings: #729 records a missed threshold, #730 reaches `reproducible` without measuring anything. **This is the only one where a preregistered test can plausibly pass.** It is also the largest of the three, and it is research work rather than paperwork.


hollomancer:
Review findings:

1. The proposed repeated-permutation result cannot be `confirmatory` on the existing February population. The favorable one-placebo outcome was recorded on August 3, and this plan selects its statistic and threshold afterward. A later design commit cannot erase that outcome visibility. Record the repeated analysis as post-hoc, or use a prospectively frozen design with genuinely unseen data or labels.

2. Current R15 cannot produce N distinct draws: it fixes `PLACEBO_SEED = 20260801`, recreates that generator on every call, and explicitly defines one non-uniform descriptive run. A new versioned protocol must justify the null distribution, accept a frozen seed schedule, persist each draw, and verify N unique mapping digests. This does not cure the confirmatory issue above on the existing data.

3. One temporal-placebo test does not validate the full B2/B3/E1 study. Inventory authorization is section-wide, so a met result could mechanically back claims beyond the tested estimand. Use a separate, narrowly scoped validation study and retain the current reproducible census study for its other claims.

4. Freeze the complete decision rule before any eligible run: scalar statistic, null, seed scheme, tail/tie handling, confidence method and level, multiplicity policy, and whether the threshold applies to the point estimate or interval bound. Add hashed run/distribution artifacts and a blocking integrity check.

5. Freeze the exact permitted wording for both met and missed outcomes before execution. Repository semantics allow `validated` with `threshold_met: false`; only a met threshold authorizes the public `Validated` rank.

Evidence-audit verdict: `BLOCK`. The current valid route remains an unseen prospective validation design, including the gated hand-label path.

hollomancer:
## Steps 1–2 done; step 3 cannot run here

Pushed `94b708cc`. The design is frozen and N draws are now possible. **The test is not run**, and that is not a shortcut — two independent preconditions block it:

- `data/processed/phase_ii_awards.parquet` is **absent from a clean checkout** (this one has 4.7M of reference data only).
- `build_phase_iii_placebo.py` states its first production invocation is gated on a separate repository-owner approval. 1000 draws is a production invocation.

Both are recorded in the design as preconditions rather than glossed.

### The blocker that made N draws impossible

`PLACEBO_SEED` was a module constant read directly inside `_cross_firm_assignment`, so exactly one draw was reachable. It is now a parameter threaded through the four public entry points, defaulting to the frozen value — **the single materialized placebo is byte-identical**, asserted by test.

### What the invariant test found

My first version asserted the pair-level date multiset is preserved. It isn't, and the failure was informative:

> Preservation is at **unique-award grain**, not pair grain: an award fans out to several pairs, so moving a date between awards of unequal fanout legitimately changes the pair-level multiset.

That fixes what the test actually rejects. The null is a null over *award-grain completion dates* — the placebo preserves the unique-award date multiset and reassigns across firms. Rejecting it says the census output depends on the real date-to-award correspondence rather than on the date distribution alone. The frozen design now says this explicitly.

### The design

`studies/phase-iii-census/validation-design.md`, pinned at `f3a15692`:

| | |
|---|---|
| Draws | N = 1000, seeds `PLACEBO_SEED+1..1000` |
| Primary | `surviving_pairs`, complete-filter cell |
| Basis / value | `proportion` / `0.95` |
| Interval | Wilson 95% — at N=1000 that is `[0.935, 0.962]`, verified |

`threshold_met` is keyed to the interval's **lower bound**, not the point estimate. Recording an interval whose lower bound sits below the threshold and calling it met is the failure this design exists to prevent.

The frozen draw (`i = 0`) is reported but is **not** a member of the null sample.

### What a pass would not license

Nothing about whether an uncoded lineage proxy identifies a genuine statutory Phase III award. That limitation survives promotion unchanged, and the design requires `permitted_claims` to say so in its own voice — because a reader seeing the inventory word `Validated` will otherwise read it as "this measures Phase III correctly."

### Remaining

Steps 3–7: run the draws, record `validation_result`, rewrite `permitted_claims`, promote, audit. All need the inputs and the approval.

6,927 tests pass; ruff, manifests, tiers, and boundary guards clean.

hollomancer:
## Owner approval received; the run is still blocked, on data not permission

Approval clears the process gate. The inputs are not on this machine.

| Input | State |
|---|---|
| `data/raw/sbir/award_data.csv` | **present** (376M) |
| `data/processed/enriched_sbir_awards.parquet` | missing — buildable from the CSV |
| `data/transition/contracts_ingestion.parquet` | **missing — not buildable here** |
| `data/processed/phase_ii_awards.parquet` | missing — needs both above |

The contract half is the blocker. `raw_contracts` produces `contracts_ingestion.parquet`, and all four of its inputs are absent:

```
transition_award_archive_dir: data/raw/usaspending/award_archive        exists=False
transition_dump_dir:          data/transition/pruned_data_store_api_dump  exists=False
transition_vendor_filters:    data/transition/sbir_vendor_filters.json    exists=False
```

The asset extracts USAspending transactions **"from removable storage"**. `data/raw/usaspending/` holds only `recipient_universe` directories — 116M of a 9.8G data tree. The FY2009–FY2025 archives the census covers are not here and are not regenerable locally.

`_load_contracts` also requires a `.checks.json` provenance manifest beside the parquet, so a reconstructed contracts file would fail verification. That refusal is correct.

### Not doing

Building a partial or synthetic `phase_ii_awards.parquet` to make the run execute. The design preregisters N=1000 draws against provenance-verified inputs; running against a reconstructed frame yields a number with a Wilson interval that looks exactly like a real result and is not one. That is the failure the `validated` reframing exists to prevent.

### To unblock

Either the USAspending award archive mounted at `data/raw/usaspending/award_archive`, or an existing `contracts_ingestion.parquet` plus its `.checks.json` from wherever it was last materialized — most likely the live server, since `docs/deployment/server-status.local.md` tracks materialization state.

From there the run is mechanical: the seed is parameterized and tested at every draw, and the design is frozen at `f3a15692`.

### Recommendation

Merge this PR as the frozen design and the enabling machinery. Run the draws as a follow-up when the corpus is reachable — which is the order the design requires anyway, since freezing before the run is what makes `confirmatory: true` defensible.

hollomancer:
Design half of path A is up as #745: Revision 16, the pinned `validation-design.md`, the seed-parameterised placebo, the per-draw path with fixture equivalence against `build_census_tables`, and the owner-gated resumable runner. Study stays `reproducible`; merging approves the design, not the run.

Three things #731 didn't specify that #745 decides — worth a look before merging: (1) the statistic is defined at the **final clause only**, because the R15 record shows the placebo ahead on distinct contracts at three intermediate stages; (2) the **R15 seed is excluded** from the 500 confirmatory seeds since its result was visible, and runs instead as an equivalence precondition that must reproduce the recorded values exactly; (3) the threshold is on the Wilson **lower bound** (≥0.95), not the point estimate.

hollomancer:
## Closing in favour of #745

#745 freezes the same validation design, written against the actual placebo machinery, and it is the better document. Its seed list is also a strict superset relationship with this one, so keeping both would be a preregistration problem rather than two experiments:

```
#731 seeds: 20260802..20261801  (n=1000)
#745 seeds: 20260802..20261301  (n=500)
#745 is a subset of #731: True     overlap: 500 of 500
```

Same placebo family, same statistic, same `threshold_basis: proportion`, same `threshold_value: 0.95`, same Wilson-lower-bound rule. Running both would be two readings of nested data against one threshold — the post-hoc selection that blocked `ma-discovery-recall` twice.

### Two things #745 has that this did not

**It restricts the statistic to the final cumulative clause, with a reason from the data.** The R15 record shows the placebo frame carrying *more* distinct contracts than the actual at three intermediate stages (0.896×, 0.886×, 0.879×). A statistic defined over every stage was **known in advance to fail** on a metric the criteria do not target. This design would have run that.

**It states the power concretely.** N = 500 tolerates at most 15 non-exceedances: 485/500 gives a lower bound of 0.9511 and passes; 480/500 has a point estimate of 0.96 and fails at 0.9390. This document gestured at an interval without pinning what would clear it.

Its section 1 also covers everything this branch's prose did — that the design does not test whether the proxy identifies statutory Phase III awards, that no label set exists, and that a `validated` rank means separation from the frozen placebo family and the permitted claim must say exactly that.

### What was carried over

One test, in `5d56f09d`: `test_each_seed_rebuilds_to_the_same_assignment`. #745's suite establishes that distinct seeds diverge and that invariants hold, but not that one seed returns the same assignment twice — and the design records 500 draws by per-seed digest, so a nondeterministic draw would pass every existing test and still leave the distribution unreconstructable. Mutation-checked.

The seed parameterisation in `placebo.py` was done independently on both branches, so nothing was lost there.

PR #732, opened 2026-09-13
Spec how reproduction works against a drifting upstream

hollomancer:
Requirements only. **No schema change, no study promoted.**

## Motivated by a real rebuild, not a hypothetical

`transition-scoring` claims its fusion corpus is *"regenerable from committed scripts and pinned manifests."* I regenerated it today to test that claim:

| Quantity | Frozen (2026-08-01) | Rebuild (2026-09-13) |
|---|---:|---:|
| corpus rows | 828 | 822 |
| positives | 138 | 137 |
| firms | 101 | 100 |

One positive short, and the six missing rows are that positive plus its negatives. **By any ordinary reading this is a successful reproduction. The repository cannot say so, and that is the defect.**

## The diagnosis is not "archives drift"

It is that the evidence needed to *classify* the difference was discarded.

`recover_award_grain.py` already writes a retrieval manifest recording exactly what's needed:

```json
{"fiscal_year": 2018,
 "url": "https://falextracts.s3.amazonaws.com/.../FY2018_archived_opportunities.csv",
 "fetched_at": "2026-09-13T22:05:14Z",
 "rows_scanned": 428147, "sbir_notices": 184, "rows_kept": 17}
```

`rows_scanned` measures the upstream **independently of any filtering this repository does**. With the original pull's manifest, 828-to-822 resolves in one comparison:

| `rows_scanned` | `rows_kept` | Meaning |
|---|---|---|
| same | same | exact reproduction |
| moved | moved | upstream drift — a finding, not a defect |
| same | moved | the pipeline changed — a regression |
| moved | same | drift that did not reach the kept set |

`study.yaml` recorded its source as the string `"/tmp/gsa_award_grain"` — a path, on a temp filesystem, now gone. **So the repository measured the drift and then threw the measurement away.** Today all four cells look identical.

This recurs for any study touching USAspending bulk archives, EDGAR full-text search, SAM.gov, or the GSA archive.

## Four requirements

1. **Pin the retrieval manifest, not a path.** The producers already write them; the change is that the study pins one. Near-zero cost.
2. **Declare a reproduction tolerance, with a derivation** — mirroring `ValidationDesign.threshold_derivation`. A band wide enough to admit any rebuild is a defect, and the derivation is what makes that visible.
3. **Report which cell a rebuild landed in.** Drift within tolerance passes and is recorded; changed kept-rows at constant upstream fails regardless of tolerance, because that is this repo's behaviour changing, not the world's.
4. **Weaken nothing.** `frozen_artifacts` still means exact bytes; studies without a live upstream are checked exactly as today.

## Explicitly out of scope

**Caching upstream snapshots to force bit-exactness.** ~10 GB per rebuild for the GSA archive alone, and it converts a live-data problem into a storage problem without making the claim any truer — the study still measured a moment in time.

## Four open questions, recorded rather than answered

Including where tolerance lives given `extra="forbid"` — the same gap that just pushed the `transition-scoring` config digest into `limitations` prose instead of a typed field. One block may serve both.

Registered in `specs/status.md`; `make docs-check` and `check_epistemic_tiers` pass.


specs/upstream-drift-reproduction/requirements.md
```diff
@@ -0,0 +1,147 @@
...
+## Explicitly out of scope
+
+- **Caching upstream snapshots to force bit-exactness.** For the GSA archive
+  alone this is roughly 10 GB per rebuild, and it converts a live-data problem
+  into a storage problem without making the claim any truer: the study still
+  measured a moment in time. If a snapshot is kept for other reasons, R1 still
+  applies to it.
+- **Promoting any study.** This spec changes what a study must record, not any
+  study's rank.
+- **Retrofitting every existing study at once.** R1 through R3 bind a study when
```
hollomancer: **[bug]** R1 is written as a universal obligation (“A study whose inputs include a source outside this repository must list the retrieval manifest…”), but the retrofit clause says “R1 through R3 bind a study when it next claims `reproducible` or above.” Those two sentences are not the same contract. The motivating rebuild is `studies/transition-scoring/study.yaml` at `evidence_status: exploratory`; under the retrofit sentence this spec would not have required a pinned retrieval manifest and would not have made 822-versus-828 classifiable. The same clause also never fires for studies that already hold `reproducible` and already use live public inputs — `studies/form-d-fundraising/study.yaml` (SEC Form D; limitations already admit input bytes are unpinned) and `studies/phase-iii-census/study.yaml` (USAspending / SBIR.gov). “When it next claims `reproducible`” is not “when it is `reproducible` and lists a live source.” As written, the rank this spec exists to make checkable stays unchecked for every current holder of it, and for the case in the PR body.

**Suggestion:** Make the binding rule one sentence and use it in both R1 and the out-of-scope bullet. If the intent is forward-only, say so in R1, and replace “when it next claims” with a rule that actually reaches existing `reproducible` live-source studies (for example: any live-upstream study at `reproducible` or above is bound; this spec does not itself edit those `study.yaml` files, but enforcement applies on the next manifest change or as named follow-up tasks). If exploratory studies may still pin a path and remain undiagnosable, say that explicitly so the 2026-09-13 rebuild is not implied to be in scope.

specs/upstream-drift-reproduction/requirements.md
```diff
@@ -0,0 +1,147 @@
...
+The derivation requirement mirrors `threshold_derivation` in
+`ValidationDesign`: a number with no stated basis is not a contract. A band
+wide enough to admit any rebuild is a defect, and the derivation is what makes
+that visible.
+
+### R3 — The check distinguishes drift from regression
+
+A rebuild comparison must report which of the four cells above it landed in.
+Upstream drift within the declared tolerance is recorded and passes. A change
+in kept rows at constant upstream size fails regardless of tolerance, because
```
hollomancer: **[suggestion]** R3 treats a single 2×2 of `rows_scanned` versus `rows_kept` as a complete drift-versus-regression test: “moved|moved” is “upstream drift — a finding, not a defect” and passes inside the R2 band; “same|moved” “fails regardless of tolerance, because that is this repository's behaviour changing, not the world's.” That is stronger than the table can support. Grain is unspecified: R1 is per retrieved object, but the motivating table and R3 read as one aggregate cell, so a pipeline change in one fiscal year can hide behind drift in another. The motivating counts (828/138/101) are corpus grain after `build_notice_corpus.py --award-grain` expands each positive by five negatives, not `recover_award_grain.py` `rows_kept`. `rows_kept` also moves when the citation index (`award_data.csv`, default `data/raw/sbir/award_data.csv`) changes at constant GSA `rows_scanned`; that input is not in `corpus.manifest.json` `sources` and has no retrieval manifest, so the “same scanned, moved kept ⇒ this repository changed” cell is not uniquely a pipeline regression. Concurrent pipeline change plus upstream drift lands in “moved|moved” and passes if the R2 band is wide enough.

**Suggestion:** State the comparison grain as per retrieved object, matching R1. Describe the 2×2 as a classification of that source’s size versus yield, not a proof of cause. Name the other live inputs that can move yield at constant `rows_scanned` (for this corpus, the award index), and require those under R1 or carve them out. Treat “moved|moved” as consistent with drift, recorded, and still subject to R2 on derived quantities — not as “not a defect.”

specs/upstream-drift-reproduction/requirements.md
```diff
@@ -0,0 +1,147 @@
...
+**Target epistemic tier:** `primitives`
+
+The deliverable is a `StudyManifest` contract change and its enforcement,
+serving studies at `reproducible` and above.
+
+**Research question anchor:** none directly. This is repository machinery: it
+changes what a study must record so that `reproducible` remains a checkable
+claim when an input is a live source.
+
+**Status:** proposed. No schema change is implemented in this spec.
```
hollomancer: **[suggestion]** The spec-workflow scope gate requires every new spec to state the verification that proves the work complete. Peer machinery specs (`epistemic-tier-enforcement`, `transition-precision-benchmark`, `modular-analysis-platform`) put that in a header bullet or a “Done when” block with an observable outcome. This file has no such statement. Open question 2 leaves whether R3 runs in CI or only at rebuild time unanswered, so R2/R3 are not independently testable as written: there is no named check, artifact, or command that would show the four requirements are met. That is the missing field; estimand is not required at `primitives`.

**Suggestion:** Add the compact header used by the other operational specs, including “Verification that proves completion.” A requirements-only freeze can still say what later implementation must demonstrate (for example: CI rejects a `reproducible` live-upstream manifest that lacks a retrieval-manifest pin and a derived-quantity band with a derivation; a rebuild path emits the four-cell report from two manifests without a snapshot cache). If Q1–Q2 must be answered before that sentence can be written, the spec is not yet Active (see the registry comment).

specs/status.md
```diff
@@ -200,6 +200,14 @@ bypassing lifecycle review; the status and rationale still require human judgmen
   into weekly reporting modules. Remaining work is injection, coverage, and
   alias cleanup.
 
+- **`upstream-drift-reproduction` — Active.** Proposed contract change so
```
hollomancer: **[suggestion]** The registry marks `upstream-drift-reproduction` as **Active** (“current implementation or cleanup work can proceed from the spec”), while `requirements.md` says **Status:** proposed, states that no schema change is implemented, and leaves Q1 (where tolerance lives under `extra="forbid"`) and Q2 (CI versus audit) open. Implementation cannot proceed from this text until those are decided; a spec-implementer reading only the registry would start a `StudyManifest` change the spec itself refuses. Closest house analogue is `phase-iii-hand-label-validation` (Gated backlog: do not implement until the design is frozen), not an Active cleanup spec. Placement is also off the rest of the list: the new bullet is appended after `weekly-awards-report-refactor` rather than with the other `transition-*` / `u*` entries.

**Suggestion:** Register as Gated backlog until Q1–Q2 are decided, or freeze those two answers in this spec so Active is true. Align `requirements.md` Status with the registry category. If it stays in the top-level list, insert it with the surrounding names rather than after weekly-awards.

specs/upstream-drift-reproduction/requirements.md
```diff
@@ -0,0 +1,147 @@
...
+Cost is near zero: the manifests are small JSON files and are already written.
+
+### R2 — A study declares what reproduction means for it
+
+`reproducible` currently reads as bit-exact, which is unachievable against a
+live source and is therefore quietly ignored rather than enforced. A study with
+a live upstream must state its reproduction tolerance: which derived quantities
+are checked, the agreement band for each, and why that band is the right one.
+
+The derivation requirement mirrors `threshold_derivation` in
```
hollomancer: **[suggestion]** R2 says the derivation requirement “mirrors `threshold_derivation` in `ValidationDesign`.” On this PR’s base, `sbir_etl/quality/study_manifest.py` has no `ValidationDesign` type and no `threshold_derivation` field; `StudyManifest` is `schema_version: Literal[1]` with `extra="forbid"`. That name exists only as unmerged work (`feat/validation-design-block` / the verification-practices SDD), so a reader of main cannot look the parallel up, and Q1’s “one block may serve both” points at a type this branch does not depend on.

**Suggestion:** Drop the type name, or declare an explicit dependency on the ValidationDesign change landing first. The rule itself (“a number with no stated basis is not a contract”) can stand without citing an unshipped model.

specs/upstream-drift-reproduction/requirements.md
```diff
@@ -0,0 +1,147 @@
...
+
+### R1 — A live-upstream study pins a retrieval manifest, not a path
+
+A study whose inputs include a source outside this repository must list the
+retrieval manifest for that source in `frozen_artifacts`. The manifest must
+record, per retrieved object: the URL, the fetch timestamp, and at least one
+upstream-size measure taken before this repository's own filtering
+(`rows_scanned` or equivalent).
+
+The producers already emit these. The change is that the study pins one instead
```
hollomancer: **[suggestion]** “The producers already emit these” is true for the GSA scripts (`recover_award_grain.py`, `pull_gsa_archive.py`, `extract_phase3_selflabeled.py` write `url` / `fetched_at` / `rows_scanned` / `rows_kept`). It is not true of the other live sources the problem statement names (USAspending bulk, EDGAR full-text search, SAM.gov). Combined with R1’s “must list the retrieval manifest,” a study on those sources cannot satisfy R1 until a producer exists. Q3 admits the size measure varies by source kind; the R1 prose still reads as if pinning is the only remaining change and the cost is near zero for every live-upstream study.

**Suggestion:** Narrow the sentence to the GSA award-grain producer that motivated the spec, and state that other source kinds need a retrieval-manifest producer before a study that uses them can claim `reproducible` under R1.

specs/upstream-drift-reproduction/requirements.md
```diff
@@ -0,0 +1,147 @@
...
+resolves in one comparison:
+
+| `rows_scanned` | `rows_kept` | Meaning |
+|---|---|---|
+| same | same | exact reproduction |
+| moved | moved | upstream drift — a finding, not a defect |
+| same | moved | the pipeline changed — a regression |
+| moved | same | drift that did not reach the kept set |
+
+`studies/transition-scoring/study.yaml` recorded its source as the string
```
hollomancer: **[nit]** The path string `"/tmp/gsa_award_grain"` is not in `studies/transition-scoring/study.yaml`. It is in `specs/phase3-notice-corpus-fusion/corpus.manifest.json` `sources`, which that study pins as a frozen artifact. A reader who opens `study.yaml` will not find the quoted path.

**Suggestion:** Cite `corpus.manifest.json` (and that the study freezes it) instead of saying `study.yaml` recorded the path.

hollomancer:
## Review addressed in `719fd894` — five accepted, one rejected with evidence

### Accepted: the provenance attribution was wrong

`/tmp/gsa_award_grain` is **not** in `studies/transition-scoring/study.yaml`. I verified: it is the first entry of `sources` in `specs/phase3-notice-corpus-fusion/corpus.manifest.json`, which the study pins as a frozen artifact.

The point survives — the provenance chain terminates in a dead temp path rather than in the retrieval manifest that recorded `rows_scanned` — but naming the wrong file would have pointed implementation at the wrong artifact. Corrected.

### Accepted: R1 said two different things

It was a universal obligation, then scoped by a retrofit clause that disagreed. **The obligation now attaches with the claim, not with the data source**: a study claiming `reproducible` or above must pin a retrieval manifest; a study at `exploratory` is encouraged to. Written once.

### Accepted: "the producers already emit these" was overclaimed

True of the GSA scripts only — `recover_award_grain.py`, `pull_gsa_archive.py`, `extract_phase3_selflabeled.py`. **Not** true of the USAspending bulk extractor, EDGAR full-text search, or SAM.gov. R1 therefore implies producer work for those sources, and the spec now says so rather than implying the cost is already paid.

### Accepted, and it made the spec better: R3 claimed more than the 2×2 supports

Counts can hold steady while the *identity* of the kept rows changes, and an upstream can revise a record in place without moving any count. R3 now requires a row-identity check at a declared grain — notice id for GSA, accession for EDGAR, award key for USAspending — and **a counts-agree-identity-disagrees result fails**. That is precisely the case counts alone would have hidden.

### Accepted: no completion criteria for a `primitives` spec

Added a **Done when** section with five observable checks. The one that matters is item 5: `transition-scoring`'s 2026-09-13 rebuild must return a determinate verdict. *If the mechanism cannot classify the rebuild that motivated it, it has not solved the problem.*

### Accepted: lifecycle mismatch

`Status: proposed` against a registry entry of `Active`. Both now read active, with the registry noting that Q1 and Q2 must be settled before implementation starts.

---

### Not accepted

> On this PR's base, `sbir_etl/quality/study_manifest.py` has no `ValidationDesign` type and no `threshold_derivation` field… That name exists only as unmerged work.

This is incorrect. #726 merged as `6704d329` and is an ancestor of this branch:

```
git merge-base --is-ancestor 6704d329 origin/spec/upstream-drift-reproduction  → in base
study_manifest.py:84   class ValidationDesign(BaseModel)
study_manifest.py:102  threshold_derivation: str = Field(min_length=1)
```

The R2 reference to `threshold_derivation` points at merged code and stands as written.

PR #735, opened 2026-09-14
Keep acquirer-side Form D rows out of the exit artifact

hollomancer:
Extracts the three defensible changes from #705 onto current `main`.

#705 is `CONFLICTING`, failing four tests, and carrying nine live threads against a base that moved twelve commits — including a **Blocking** finding that `docs/research-questions.md`, the north star and scope gate, now describes behaviour a rebase dropped. This ports what holds and leaves that behind.

## Three changes

**1. Tier filter.** `extract_form_d_signals` drops records whose identity match did not reach `KEEP_MATCH_TIER`. The scorer already graded those joins; carrying them forward re-admitted matches it had rejected. The drop count prints, so the filter is visible rather than silent.

**2. Form D alone is no longer `high`.** Item 10 marks a Rule 145 transaction — a deemed offer and sale of securities *by the issuer* — so the filer is the acquirer and the flag is evidence in the wrong direction. Grading it `high` inflated the exit population with rows recording **purchases by the SBIR firm**.

**3. A sibling artifact.** Rows whose only evidence is that flag go to `data/sbir_ma_non_exit.jsonl` with `non_exit_reason='acquirer_side'`, not to the exit file. Demoting them was not enough — several consumers treat row presence as an exit and never read confidence.

## One decision made differently from #705, deliberately

#705's `EFTS_SIGNAL_KEYS` included `efts_ma_proxy` and `efts_ownership_active`, so **a single SC 13D or merger-proxy mention kept an otherwise acquirer-side row in the exit artifact**. The signal memo grades both Low and neither is target-side evidence.

`TARGET_SIDE_EFTS_SIGNALS` here is `subsidiary`, `ma_definitive`, `acquisition_text` only. A review thread on #705 raised this; it is settled here rather than inherited.

## Not ported

Moving the producer out of `scripts/archive/`. That is a separate question the archive README has its own view on, and it does not belong in a correctness fix.

## Tests

`test_assign_confidence_form_d_is_high` **asserted the defect** and is inverted, with its docstring recording what it used to claim. Five added: the sibling routing, each target-side signal keeping a row, both low-grade mentions failing to rescue one, and a row with no Form D flag never being acquirer-side-only.

6,939 unit tests pass; ruff, format, and `docs-check` clean.

## Suggested disposition for #705

Close it. Its remaining threads are documentation describing dropped code, and the substantive questions are decided here.


scripts/archive/data/detect_sbir_ma_events.py
```diff
@@ -32,14 +64,22 @@ def extract_form_d_signals(
...
     """
     events = []
+    dropped_on_tier = 0
     for r in records:
         confidence = require_form_d_tier_rule(
             r.get("match_confidence"),
             expected_rule_version=expected_rule_version,
             context=f"Form D record {r.get('company_name') or '<unnamed>'!r}",
         )
+        if confidence.get("tier") != KEEP_MATCH_TIER:
```
hollomancer: **[suggestion]** Change 1 (drop identity matches below `KEEP_MATCH_TIER`) has no test that a medium or low record *with* a business-combination offering is omitted. Existing coverage still passes for other reasons: `test_extract_form_d_signals_finds_business_combination` is already high, and `test_extract_form_d_signals_skips_non_combo` uses `tier: "medium"` with no combo, so after this patch it is skipped by the new gate rather than by the combo check. A regression that kept medium/low combo rows would not fail the suite.

**Suggestion:** Add a high-tier combo control plus a medium (and/or low) combo record and assert only the high-tier row is emitted. Point `test_extract_form_d_signals_skips_non_combo` at a high-tier non-combo record so it still tests the combo filter.

scripts/archive/data/detect_sbir_ma_events.py
```diff
@@ -32,14 +64,22 @@ def extract_form_d_signals(
...
+            dropped_on_tier += 1
```
hollomancer: **[suggestion]** `dropped_on_tier` increments for every non-high Form D *record* before the combo check. `extract_form_d_signals` is called on the full `form_d_details.jsonl` (thousands of fundraising-only rows). The log line will mostly count records that never would have become M&A events, so it does not show how many combo joins the filter actually removed.

**Suggestion:** Increment the counter only when `tier != KEEP_MATCH_TIER` and the record has at least one `is_business_combination` offering — or print both “records below tier” and “combo events dropped.”

tests/unit/scripts/archive/test_detect_ma_events.py
```diff
@@ -231,11 +232,53 @@ def test_merge_events_separate_companies():
...
+def test_form_d_with_target_side_efts_still_reaches_high():
+    """A Form D flag alongside target-side evidence is not demoted."""
+    event = {
+        "form_d_detail": {"filing_date": "2020-01-01"},
+        "efts_detail": {"mention_types": ["subsidiary"]},
+    }
     assert assign_confidence(event) == "high"
 
 
+def test_acquirer_side_only_routes_out_of_the_exit_artifact():
```
hollomancer: **[suggestion]** `test_acquirer_side_only_routes_out_of_the_exit_artifact` is named and documented as writer behavior (“written to the sibling file, not the exit file”) but only asserts `is_acquirer_side_only(...)`. A bug that wrote the row to `--output`, omitted `non_exit_reason`, or skipped the sibling file would still pass.

**Suggestion:** Drive `main()` (or an extracted writer) with temp paths and assert a Form D-only row appears solely in `--non-exit-output` with `non_exit_reason == "acquirer_side"`, and that a target-side EFTS row appears only in `--output`.

scripts/archive/data/detect_sbir_ma_events.py
```diff
@@ -22,6 +22,38 @@
...
+# evidence, so they do not rescue an otherwise acquirer-side row.
+TARGET_SIDE_EFTS_SIGNALS = (
+    "efts_subsidiary",
+    "efts_ma_definitive",
+    "efts_acquisition_text",
+)
+
+
+def is_acquirer_side_only(signals: dict[str, bool]) -> bool:
+    """True when the only evidence is a Form D business-combination flag.
```
hollomancer: **[suggestion]** The docstring says the predicate is true when “the only evidence is a Form D business-combination flag,” but the body — and `test_low_grade_efts_mentions_do_not_rescue_an_acquirer_side_row` — return true when that flag is joined by `efts_ma_proxy` or `efts_ownership_active`. The second paragraph also narrates the sibling-file design; that belongs on the writer, not on this boolean.

**Suggestion:** State that it is true when Form D Item 10 fired and none of `TARGET_SIDE_EFTS_SIGNALS` did. Drop the consumer/sibling rationale from this docstring.

hollomancer:
## Measured impact, and a thread-status pass

I re-checked each open thread against the current code rather than the thread list, and measured what this PR moves. Two things worth adding to the record.

### Eight of the nine open threads are already fixed

GitHub has not marked them outdated, but the code answers them:

| Thread | Where it is answered |
|---|---|
| `--output x --non-exit-output x` truncates one file twice | `SystemExit` guard at `detect_sbir_ma_events.py:346-351`, plus `test_writer_refuses_to_write_both_artifacts_to_one_path` |
| Tier gate untested; would pass if inverted | `test_tier_gate_drops_a_combination_record_below_high` and `test_tier_gate_drops_low_as_well_as_medium` |
| Routing not tested end-to-end | `test_writer_routes_acquirer_side_rows_to_the_sibling_file` calls `main()` and asserts both files and `non_exit_reason` |
| Test named as writer behaviour but only asserts the predicate | Same test — now exercises the writer |
| Import ordering, `dropped_on_tier` population, predicate docstring | Already outdated |

24 tests pass on the branch. The remaining substantive thread is the downstream one, below.

### What the PR moves, measured

The description argues the correction qualitatively but gives no counts. Against the committed `data/sbir_ma_events.jsonl` (4,004 rows):

**Rows routed to the sibling artifact: 117 (2.9%).** Distinct firms 4,001 → 3,884. All 117 are currently graded `low`.

That last point looks at first like it undercuts the PR — if the acquirer-side rows are already `low`, what is being inflated? The opposite turns out to be true. Recomputing every row's tier under each rule:

| | `high` |
|---|---|
| committed `sbir_ma_events.jsonl` | **676** |
| recomputed under `main`'s `assign_confidence` | **811** |
| recomputed under this branch's `assign_confidence` | **676** |

247 rows disagree with `main`'s rule; 113 with this branch's, and those are medium/low shuffles rather than `high`.

**The committed artifact already matches this branch's rule, not `main`'s.** Whatever produced it had already dropped `has_form_d` from the `high` condition. So regenerating with `main` today would raise the high tier from 676 to 811 — **+135 rows, +20%** — of rows whose only evidence is an Item 10 flag filed by the acquirer.

That reframes the PR usefully: it is not proposing a new position, it is restoring the code to the semantics the committed data already reflects, and it is what prevents a 20% inflation of the high tier the next time anyone regenerates.

### The downstream thread is real, and larger than it reads

Copilot's point about rows leaving the refinement chain is the one thread I would still treat as open. The blast radius is wider than the two scripts named: **ten consumers read `enriched_sbir_ma_events.jsonl`**, including `sbir_etl/utils/transition_signals.py`, `sbir_etl/capital_events/sources/ma_events.py`, `sbir_etl/reporting/tech_area_cohort.py`, `build_capital_events.py`, `build_supplier_share_census.py`, `build_nano_cohort.py` and `nano_prime_acquisitions.py`.

Given the numbers above, the honest framing is that those consumers are *already* living on this branch's semantics for the `high` tier, and this PR stops the next regeneration from diverging. But the 117 rows leaving `sbir_ma_events.jsonl` do change what reaches the enriched file, so the firm counts every one of those ten reads will move. Worth stating the 4,001 → 3,884 delta in the PR body so nobody rediscovers it downstream.

Numbers produced with the branch's own `is_acquirer_side_only` and both `assign_confidence` implementations against the committed artifact; happy to share the script.

hollomancer:
Took another pass at this against current `main` (`5e227a20`). CI green, mergeable, no conflicts.

**The five open threads are closed in code at `594909e`.** I checked by mutation rather than by reading — broke each fix deliberately to confirm a test catches it:

| Thread | Mutation check |
|---|---|
| same file for both outputs | disabling the guard fails `test_writer_refuses_to_write_both_artifacts_to_one_path` |
| tier gate untested (raised twice) | neutering the gate fails both `test_tier_gate_*` tests |
| routing not tested end-to-end | disabling the routing fails `test_writer_routes_acquirer_side_rows_to_the_sibling_file`, which drives the CLI and asserts both files |

24 tests pass. The remaining thread — the cross-enrichment contract — is a decision, not a defect, and is disclosed in the doc note rather than resolved.

**That decision now has a number on it.** Applying `is_acquirer_side_only` to the real artifacts:

`data/sbir_ma_events.jsonl` (4,004 rows): **117 acquirer-side-only (2.9%), all already graded `low`.** Rerouting them costs the cross-enrichment measure very little, because nothing treated them as strong evidence.

`data/enriched_sbir_ma_events.jsonl` (4,306 rows — what consumers actually read): **434 acquirer-side-only (10.1%), and all 434 are graded `high`.**

| Signals present | Rows |
|---|---|
| `form_d_business_combination` only | 407 |
| `+ efts_ownership_active` | 14 |
| `+ efts_ma_proxy` | 12 |
| `+ both` | 1 |

So the enriched artifact currently grades **407 rows as high-confidence exits on a single Form D Item 10 flag** — filed by the acquirer. That is change 2's argument, realized at 9.4% of the artifact rather than hypothetical. The refinement chain is what promotes them: the raw exit file carries 141 `form_d_business_combination` rows, the enriched file 552.

The 27 rows with `efts_ma_proxy` / `efts_ownership_active` alongside are precisely the population `TARGET_SIDE_EFTS_SIGNALS` excludes where #705 would have kept them. That deliberate divergence is worth 27 rows — worth knowing when judging whether it was the right call.

**Reading:** the evidence runs in this PR's favour. The undercount risk behind the open thread is real but small at the exit artifact (117 low-confidence rows); the larger number points the other way — 434 rows consumers read as high-confidence exits are acquirer-side. It also reaches the transition signals, since `sig_ma_detected` is computed from the enriched file, so roughly a tenth of firms currently counted as having an M&A signal carry acquirer-side evidence.

One caveat on those figures: they measure the artifacts as they stand, not a prediction of a rerun — the two files disagree on Form D flag counts (141 vs 552) because refinement adds evidence, so the post-merge enriched artifact has to be regenerated to be counted. And whether each of the 434 is genuinely an acquisition *by* the SBIR firm was not adjudicated here; that rests on Item 10's filer-side semantics, which is the PR's argument.

PR #745, opened 2026-09-14
Freeze the R16 permutation-separation validation design for phase-iii-census

hollomancer:
## Description

Adds a preregistered validation design to `phase-iii-census` as **Revision 16**, without changing
any criterion, cell, population, or the recorded R15 result. The study stays `reproducible` in this
PR; no result is recorded. This is the design half of #731's recommended path A.

**Why a revision rather than a study-file edit.** The R15 placebo contract says no "similarity
threshold, pass/fail rule, preferred cell, headline number, inferential test, or criterion change"
follows from the comparison. A permutation test is an inferential test, so it enters as a new
numbered revision with re-pinned design and amendment-log digests, exactly as R15 did.

### The design (`studies/phase-iii-census/validation-design.md`, pinned)

- **Family:** the frozen R15 randomized cyclic group derangement, unchanged. Not uniform over all
  derangements; the null is over this family only, and the document says so.
- **Seeds:** `20260802`–`20261301`, N = 500, in order. **The R15 seed is excluded** — its result was
  visible when this was written. It runs first as an equivalence precondition instead.
- **Primary statistic:** share of draws in which the actual frame *strictly* exceeds the placebo on
  `surviving_pairs` at the final cumulative clause. Ties are not exceedances. Wilson 95% interval.
- **Threshold:** met iff the Wilson **lower bound** ≥ 0.95 (`threshold_basis: proportion`,
  `threshold_value: 0.95`). Tolerates ≤15 non-exceedances of 500; 480/500 has a 0.96 point
  estimate and fails. The interval rule follows #726: the floor is cleared by the interval, not
  the estimate.
- **Secondary, no threshold:** firms and contracts at the final clause, and every metric in each of
  the six cells. Dollars descriptive (signed).
- **Final clause only, deliberately.** The R15 record shows the placebo carrying *more* distinct
  contracts at three intermediate stages (0.896×, 0.886×, 0.879×). An "all stages, all metrics"
  statistic would be known in advance to fail on a metric the criteria do not target.

### Implementation

- `build_placebo_assignment` and friends gain `seed=`, defaulting to the R15 seed. A test asserts
  the default reproduces the R15 mapping digest byte for byte.
- `permutation.py` evaluates the three date-independent clauses once and the two date-dependent
  clauses per draw, then applies the frozen final-stage summary and six-cell builder via two new
  public aliases in `criteria.py`. Every core clause is row-wise, so this is the same intersection
  `build_census_tables` computes; **fixture tests assert equality with the shared builder for four
  seeds**, following the Revision 8 execution-equivalence precedent.
- `build_phase_iii_placebo_permutation.py` is owner-gated like R15. Sequence: freeze check →
  provenance-verified inputs → **precondition: seed 20260801 must reproduce the recorded R15
  values** (546,242 / 1,985 / 21,357 / $46,386,904,542.06; actual 727,292 / 2,369 / 28,665 /
  $55,080,851,466.46; digest `c1c97a9c…`) or the run stops with no confirmatory draw → draws in
  fixed order into a resumable store → exceedance table and manifest **only when all 500 seeds are
  present**. It refuses any draw count other than 500 and any store containing a foreign seed.
- Wilson implementation checked against `scipy.stats.binomtest(...).proportion_ci("wilson")` to
  five decimals on six cases; the test carries those values.

### Manifest

`frozen_artifacts` re-pinned (`design.md`, `amendments.md`) and the validation design added;
`identity_policy.version: phase-0-r16`; new `implementation` entry; a `validation_design` block
with `threshold_basis: proportion`, `threshold_value: 0.95`; a new limitation stating the design
is frozen but unrun and what a met threshold will and will not mean.

### What merging does and does not approve

Merging **approves Revision 16** — the amendment records the merge as the approval act and its
commit as the anchor, and asks you to read that as deliberate. Merging **does not** approve the
production run; that stays behind the separate owner-approval gate as with R15. Runtime is
unmeasured (two date predicates over 53.9M rows plus six cells over ~7M survivors, per draw);
expect hours, and use `--batch-size` to run it in resumable pieces.

**Merge-commit, not squash.** #729 found that a squash destroyed the freeze ordering that decides
`confirmatory`. This PR is one commit so the anchor survives either way, but the later run PR will
depend on this commit's position in history.

Fixes # (no issue). Related: #731.

## Type of Change

- [x] Documentation update

## Versioning Impact

## Testing

Clean `git archive` of this branch, repository venv:

- Census, placebo, permutation, both placebo scripts, manifest suites: **219 passed**
- `tests/unit/scripts` + `tests/unit/quality`: 913 passed; the 12 failures are identical on an
  untouched archive of `origin/main` (git-dependent hygiene scans and network-mocked literature
  tests that do not run in a plain archive). `check_config_boundaries.py` passes against the
  committed tree.
- `ruff check .`, `ruff format --check .`, UP042 preview: clean
- `validate_study_manifests.py` (5 manifests) and `check_research_question_status.py`: pass
- Every pinned digest in `study.yaml` matches the committed bytes.

- [x] All tests pass locally

## Checklist


packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
```diff
@@ -0,0 +1,327 @@
...
+    """Run the frozen placebo family over ``seeds`` and record final-stage metrics.
+
+    The actual frame is summarised once with the same code path. Every seed
+    produces exactly one assignment, one final-stage row, one six-cell table and
+    one mapping digest; a seed that cannot be deranged raises, it does not skip.
+    """
+
+    if len(set(seeds)) != len(seeds):
+        raise CensusInputError("permutation seeds must be distinct")
+    fixed = date_independent_mask(pairs, data_cut_date)
```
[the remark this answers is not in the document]
hollomancer: Confirmed and fixed in 4397360c. `_iter_core_clause_survivors` does call `validate_pair_frame` before yielding the first stage, so the fast path was skipping a fail-closed check the frozen path performs — which also weakened the equivalence claim the whole design rests on. `run_permutation_draws` now validates before computing the fixed mask.

One validation covers all 500 draws rather than one per draw, and that is provable rather than a shortcut: `validate_pair_frame` reads key and identifier columns only (`target_transaction_id`, `target_contract_key`, `prior_award_id`/`prior_award_key`, both UEIs), and the placebo changes nothing but `prior_period_of_performance_end`. Two tests: a malformed frame raises from both `build_census_tables` and `run_permutation_draws`, and a permuted frame still passes validation.

scripts/data/build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,387 @@
...
+            "the repository owner separately approves the permutation run"
+        )
+    if draws != perm.R16_DRAWS:
+        raise CensusInputError(
+            f"the preregistered design fixes {perm.R16_DRAWS} draws; got {draws}. A different "
+            "count is a different design and needs a new amendment."
+        )
+
+    freeze = verify_frozen_spec()
+    pairs, inputs, data_cut = _load_pairs()
```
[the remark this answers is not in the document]
hollomancer: Fixed in 4397360c — this was the most consequential of the five. `verify_frozen_spec` covers `design.md` and `amendments.md` only, so nothing tied a run to the protocol it was executing, and the emitted manifest could not say which bytes it followed. That matters directly for `confirmatory`: the later `validation_result.design_sha256` has to name bytes the run actually ran under.

`verify_validation_design()` reads the pin from `study.yaml` (single source of truth rather than a second constant), hashes the file, and refuses on mismatch; the digest is recorded in the run manifest and is part of the run fingerprint. Verification happens before any input read. Tested against an edited protocol in a temporary repository root.

scripts/data/build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,387 @@
...
+    store_dir = output_dir / "draw_store"
+
+    precondition_path = output_dir / PRECONDITION_NAME
+    if precondition_path.exists():
+        precondition = json.loads(precondition_path.read_text(encoding="utf-8"))
+        if not precondition.get("matched_recorded_r15"):
+            raise CensusInputError("recorded precondition did not match R15; refusing to resume")
+    else:
+        precondition = run_equivalence_precondition(pairs, data_cut)
+        _write_json_atomic(precondition, precondition_path)
```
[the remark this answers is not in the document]
hollomancer: Fixed in 4397360c. A `run_fingerprint` over the freeze record, the pinned design digest, the input digests and row counts, and the data cut is written with the precondition; resuming against a different fingerprint is refused with the recorded and current values in the message. A draw store with no recorded precondition is now refused outright rather than resumed. Both are tested, with `run_permutation_draws` patched to fail the test if any draw is taken after a guard should have stopped the run.

scripts/data/build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,387 @@
...
+
+    seeds = perm.preregistered_seeds(draws)
+    existing = _load_store(store_dir)
+    done = set(existing["placebo_final"]["seed"].astype(int)) if existing else set()
+    if not done <= set(seeds):
+        raise CensusInputError("draw store contains seeds outside the preregistered list")
```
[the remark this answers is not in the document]
hollomancer: Fixed in 4397360c. `complete_seeds()` intersects the seeds that appear with their full expected row count in all three frames (1 / 6 / 1), and `require_complete_store()` asserts exact seed-set equality and per-seed counts before summarising — so secondary intervals can no longer be computed over a different number of draws than the primary statistic.

One note: with the reconciliation in the next comment, `require_complete_store` is unreachable through `run()`. I kept it as a checkable invariant rather than deleting it, and tested it directly instead of writing a test that pretends to reach it through the public path.

scripts/data/build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,387 @@
...
+        ),
+        "placebo_cells": pd.concat(
+            [existing.get("placebo_cells"), draws.placebo_cells], ignore_index=True
+        ),
+        "mapping_digests": pd.concat(
+            [existing.get("mapping_digests"), draws.mapping_digests], ignore_index=True
+        ),
+    }
+    for label, frame in merged.items():
+        _write_parquet_atomic(frame, store_dir / STORE_NAMES[label])
```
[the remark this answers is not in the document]
hollomancer: Fixed in 4397360c, though not the way you suggested — worth saying why. A versioned batch directory plus a completion marker makes a torn batch *detectable*; the store then refuses to load, which strands a run that may be hours in. Reconciliation makes it *recoverable*: `_load_store` truncates all three frames to the seeds complete in every one of them and rewrites, so an interruption between the per-frame writes costs the affected seeds a rerun and nothing else. Each individual write was already atomic via `os.replace`, so the only exposure was the set, and truncation is idempotent.

A complete foreign seed is still refused — that is a different design, not a resumable draw. Tested both ways: a torn batch truncates to the complete seed, resumes, and completes the full seed list; a foreign seed present in all three frames raises.

hollomancer:
All five review findings fixed in `4397360c`, pushed as a second commit rather than a force-push so the review history stays readable.

Four were straightforwardly right and I've replied on each thread. The `validate_pair_frame` bypass was the one I most needed pointing out: it isn't only a fail-closed gap, it weakened the equivalence claim the design rests on, since the frozen path validates and the fast path did not.

I took a different route on one. The suggestion for batch atomicity was a versioned batch directory plus a completion marker; that makes a torn batch *detectable*, and the store then refuses to load, stranding a run that may be hours in. Reconciliation makes it *recoverable* — `_load_store` truncates to the seeds complete in all three frames and rewrites, so the affected seeds are simply rerun. Individual writes were already atomic via `os.replace`; only the set of three was exposed.

One thing the review changed in the design document: §7 now records the design-digest binding, the pair-frame validation, the run fingerprint, and the store reconciliation. That moves the document's SHA-256, so it is re-pinned in `study.yaml` — legitimate here only because the design is not frozen until this PR merges. After merge any change to it is a new numbered revision.

Re-verified in a clean archive: 373 passed, 3 skipped; ruff check, format, and UP042 clean; both CI manifest checkers pass; all three pinned digests match their files; and the runner resolves and verifies the pinned design end to end from the repository root.

scripts/data/build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,387 @@
...
+        return status
+
+    final = existing["placebo_final"].sort_values("seed", kind="stable").reset_index(drop=True)
+    cells = (
+        existing["placebo_cells"]
+        .sort_values(["seed", "cell_id"], kind="stable")
+        .reset_index(drop=True)
+    )
+    digests = existing["mapping_digests"].sort_values("seed", kind="stable").reset_index(drop=True)
+    if digests["mapping_sha256"].nunique() != draws:
```
hollomancer: **[suggestion]** The “same assignment” refusal hashes `mapping_sha256` values that already include the per-row `seed` (`placebo.py` `_mapping_digest`). Distinct confirmatory seeds therefore produce distinct digests even when the donor mapping is identical, so `nunique() != draws` cannot detect a collapsed null. The confirmatory sample would still be summarised at N = 500.

**Suggestion:** Compare a seed-independent assignment identity (donor award/firm sequence, or a digest that omits `seed`) and refuse when that identity repeats. Keep the per-seed digest for audit.

scripts/data/build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,387 @@
...
+        precondition = json.loads(precondition_path.read_text(encoding="utf-8"))
+        if not precondition.get("matched_recorded_r15"):
+            raise CensusInputError("recorded precondition did not match R15; refusing to resume")
+    else:
+        precondition = run_equivalence_precondition(pairs, data_cut)
+        _write_json_atomic(precondition, precondition_path)
+
+    seeds = perm.preregistered_seeds(draws)
+    existing = _load_store(store_dir)
+    done = set(existing["placebo_final"]["seed"].astype(int)) if existing else set()
```
hollomancer: **[suggestion]** Completion and the foreign-seed rule are taken only from `placebo_final`. `validation-design.md` §7 says a store containing a seed outside the list is refused, but a foreign or missing seed in `placebo_cells` or `mapping_digests` is accepted. The runner then writes a confirmatory exceedance table whose secondary intervals are not over the preregistered 500 draws. `test_store_with_foreign_seed_is_refused` only mutates `draws_final.parquet`, so this hole is untested. An empty store still cannot emit a result; a planted 500-row `placebo_final` can.

**Suggestion:** Require the three store frames to carry exactly the preregistered seed set, with one digest row and six cell rows per seed, before setting `complete` or calling `exceedance_table`. Extend the foreign-seed test to cells and digests.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
hollomancer: **[suggestion]** `run_permutation_draws` never calls `validate_pair_frame`, which `build_census_tables` runs before any count. `build_uei_pairs` is a UEI merge and can retain duplicate prior-award × target-transaction rows; the census contract fails closed on that grain. The public aliases themselves are identity wrappers around `_summarize` / `_build_sensitivity_grid_from_full`, and the fixture tests lock numeric equality on already-valid frames, so the survivor-set identity holds when validation would pass. On a malformed frame the R16 path summarises instead of raising, which is not execution-equivalent to the shared builder.

**Suggestion:** Call `validate_pair_frame(pairs)` once before computing `date_independent_mask`, and add a regression that a duplicate prior × transaction frame raises in `run_permutation_draws` the same way it does in `build_census_tables`.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
```diff
@@ -0,0 +1,327 @@
...
+        raise CensusInputError("the permutation design requires at least one draw")
+    if first_seed <= PLACEBO_SEED:
+        raise CensusInputError(
+            "confirmatory seeds must all exceed the R15 seed; that draw was seen before "
+            "the design was written and is excluded from the confirmatory sample"
+        )
+    return [first_seed + i for i in range(draws)]
+
+
+def _clause_masks(pairs: pd.DataFrame, data_cut_date: date) -> dict[str, pd.Series]:
```
hollomancer: **[suggestion]** `_clause_masks` is defined and unused. `date_independent_mask` walks `CORE_CLAUSES` itself. Dead helpers do not belong in a freeze whose per-draw path is part of the study contract.

**Suggestion:** Remove `_clause_masks`.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
```diff
@@ -0,0 +1,327 @@
+"""Permutation separation for the Phase III census placebo (R16).
+
+R15 ran one fixed-seed cross-firm derangement and recorded direction only. This
```
hollomancer: **[suggestion]** The module docstring narrates R15 versus R16 (“R15 ran one fixed-seed… recorded direction only”) and restates the amendment history. The useful constraint — date-independent clauses evaluated once, row-wise intersection equals `build_census_tables` — is buried in that history. Design rationale belongs in `validation-design.md` / `amendments.md`.

**Suggestion:** Keep a short WHY for the two-clause fast path and the builder identity. Drop the R15 recap.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/placebo.py
```diff
@@ -153,8 +153,14 @@ def _mapping_digest(audit: pd.DataFrame) -> str:
...
 
 
-def _cross_firm_assignment(unique_awards: pd.DataFrame) -> pd.DataFrame:
-    """Build the frozen randomized cyclic assignment (not a uniform derangement)."""
+def _cross_firm_assignment(
+    unique_awards: pd.DataFrame, *, seed: int = PLACEBO_SEED
+) -> pd.DataFrame:
+    """Build the frozen randomized cyclic assignment (not a uniform derangement).
+
+    ``seed`` defaults to the R15 seed so the recorded single-draw assignment is
```
hollomancer: **[suggestion]** The new `seed` docstring narrates the revision (“defaults to the R15 seed… R16 passes the members of its preregistered seed list here”) instead of the non-obvious constraint: omitting `seed` must keep the recorded single-draw assignment unchanged.

**Suggestion:** State that the default is `PLACEBO_SEED` so callers that omit `seed` reproduce the frozen R15 mapping. Leave R16’s seed list to the permutation runner.

tests/unit/scripts/test_build_phase_iii_placebo_permutation.py
```diff
@@ -0,0 +1,167 @@
...
+    store = tmp_path / "draw_store" / "draws_final.parquet"
+    frame = pd.read_parquet(store)
+    frame.loc[0, "seed"] = 1
+    frame.to_parquet(store, index=False)
+
+    with pytest.raises(CensusInputError, match="outside the preregistered list"):
+        MODULE.run(tmp_path, owner_approved=True, draws=2)
+
+
+def test_recorded_r15_constants_are_the_published_final_stage_values() -> None:
```
hollomancer: **[nit]** `test_recorded_r15_constants_are_the_published_final_stage_values` is the transcription guard for `placebo-results-2026-08-03.md`, but it never asserts the dollar totals (`46_386_904_542.06` / `55_080_851_466.46`) that `run_equivalence_precondition` compares to the cent. A mistyped dollar constant would not fail CI.

**Suggestion:** Assert both `total_obligated_dollars` values (and the full 64-character assignment digest) next to the pair/firm/contract counts.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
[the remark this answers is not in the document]
hollomancer: Addressed. `run_permutation_draws` calls `validate_pair_frame(pairs)` at permutation.py:136, before the fixed mask is computed, with a comment noting it inspects only key and identifier columns so one call covers every draw.

hollomancer: Same point as the thread above; addressed at permutation.py:136.

scripts/data/build_phase_iii_placebo_permutation.py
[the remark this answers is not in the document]
hollomancer: Addressed in `4397360c`. `run_fingerprint(freeze, design, inputs, data_cut)` is persisted in the precondition and compared on resume, refusing with 'this run does not match the one the draw store was started under' when it differs. The resume path no longer trusts `matched_recorded_r15` alone.

scripts/data/build_phase_iii_placebo_permutation.py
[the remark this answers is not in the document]
hollomancer: Addressed. `require_complete_store(frames, seeds)` requires every store frame to hold exactly its expected per-seed row count for exactly the preregistered seed set, and raises before summarising otherwise. Its docstring states the reason directly: secondary intervals must never be drawn from a different number of rows than the primary statistic.

tests/unit/scripts/test_build_phase_iii_placebo_permutation.py
hollomancer: Already asserted. `test_recorded_r15_constants_are_the_published_final_stage_values` checks both dollar totals to the cent (`46_386_904_542.06`, `55_080_851_466.46`) alongside the 64-character assignment digest and the pair/firm/contract counts.

scripts/data/build_phase_iii_placebo_permutation.py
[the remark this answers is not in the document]
hollomancer: Addressed in `4397360c`. `verify_validation_design()` verifies the pinned `studies/phase-iii-census/validation-design.md` bytes and returns the evaluated digest, which feeds `run_fingerprint` and lands in the manifest. Its docstring names the gap exactly: `verify_frozen_spec` covers design.md and amendments.md only, and without this a run could proceed against an edited protocol with the manifest never saying which bytes it followed.

Worth recording why this one mattered most. It is the same failure `ma-discovery-recall` died of: its protocol was editable after its own run, and the design ended up pinned about ten hours after the replay it was supposed to have preregistered, with every run-time check passing. PR #742 hardened the discovery CLI against that a day ago, and this producer needed it independently — which suggests the check belongs somewhere shared rather than re-implemented per producer. Not a change for this PR; noting it so the pattern is visible.

scripts/data/build_phase_iii_placebo_permutation.py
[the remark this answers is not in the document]
hollomancer: Answered by design rather than changed, and I want to show the reasoning so it can be overruled if you disagree.

The comment offers two outcomes for a torn batch — `_load_store` either refuses the resumable run or accepts inconsistent tables. It does neither. It truncates to the seeds that are complete in **all three** frames, and `complete_seeds` requires each seed to carry its exact expected row count per frame, not merely to be present:

```python
per_frame.append({int(seed) for seed, n in counts.items() if int(n) == expected})
return set.intersection(*per_frame)
```

A seed written to two frames and interrupted before the third is dropped from all three and simply redrawn. The store that loads is consistent by construction, and `require_complete_store` then re-checks the invariant before any summarising, so the reasoning is verified rather than assumed.

`_load_store` states the tradeoff outright: truncating makes a torn batch recoverable where refusing to load would strand a multi-hour run. A batch directory plus completion marker would also be correct, and it would add a second consistency mechanism alongside one that already holds. I would rather not carry both.

Resolving on that basis — reopen if you want the marker anyway.

hollomancer (commented):
## Summary

Revision 16 freezes a preregistered permutation-separation design without promoting the study or recording a result: `evidence_status` stays `reproducible`, `validation_result` is absent, and the three pinned SHA-256 digests match the committed bytes. The executable path matches the frozen contract on the honest run — same R15 cyclic family, confirmatory seeds `20260802`–`20261301`, R15 seed excluded, strict exceedance, Wilson lower bound ≥ 0.95, final clause only, and the recorded R15 precondition — and `seed=` still defaults to `20260801`. The owner-gated runner cannot emit a confirmatory result from a partial store, and it refuses a foreign seed in `placebo_final`. Residual risk is in the resume/integrity guards (digest uniqueness is vacuous; cells/digests are not seed-checked), a fail-open skip of `validate_pair_frame`, unused code, and comments that narrate R15/R16 history.

## Issue counts by severity

- bugs: 0
- suggestions: 6
- nits: 1

scripts/data/build_phase_iii_placebo_permutation.py
hollomancer: Correct, and the check was vacuous as written: `_mapping_digest` folds `seed` into every record, so distinct seeds always yield distinct `mapping_sha256` values and the uniqueness assertion could never fire. Addressed on the branch before merge in `326a5827` — `assignment_identity()` hashes the donor mapping without the seed, the runner persists it per seed in the digest store, and the collapse check now runs on identity rather than on the seed-salted digest. A test plants a collapsed identity across two seeds and asserts the refusal.

scripts/data/build_phase_iii_placebo_permutation.py
hollomancer: Fixed in `4397360c`, and you were right that §7's wording was ahead of the code. `complete_seeds()` intersects the seeds carrying their full expected row count in all three store frames (1 final / 6 cells / 1 digest), `_load_store` truncates to that set, and `require_complete_store()` asserts exact seed-set equality and per-seed counts before summarising. The foreign-seed rule is applied to that reconciled set, so a seed outside the preregistered list is refused rather than inferred from `placebo_final` alone.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
hollomancer: Removed in `326a5827`. It was a leftover from an earlier shape of the per-draw path; `date_independent_mask` walks `CORE_CLAUSES` directly and nothing else called it. Agreed on the principle — an unused helper inside a frozen execution path invites someone to wire it in later and change what the design means.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/permutation.py
hollomancer: Trimmed in `326a5827`. The docstring now states only the load-bearing constraint — frozen predicates, three date-independent clauses evaluated once, row-wise clauses so the per-draw final stage equals the last row of `build_census_tables` — and the R15-versus-R16 history lives in `amendments.md` where it belongs.

packages/sbir-analytics/sbir_analytics/assets/phase_iii_negative_controls/placebo.py
hollomancer: Trimmed in `326a5827` to the non-obvious fact a caller needs: omitting `seed` reproduces the frozen R15 mapping. The revision narrative is in the amendment log.

hollomancer:
Submitting the review that was sitting in `PENDING` — its 14 comments are now visible on the merged PR, and every thread has a reply.

Seven were fresh findings and all seven were already fixed on the branch before merge; I verified each against `origin/main` rather than trusting the commit messages, and posted the specific fix note on each thread:

| Finding | Fixed in | On `main` |
|---|---|---|
| `mapping_sha256` uniqueness check vacuous — digest is seed-salted | `326a5827` | `assignment_identity()` + persisted per seed; collapse check runs on identity |
| Completion and foreign-seed rule read `placebo_final` only | `4397360c` | `complete_seeds()` across all three frames; `require_complete_store()` |
| `_clause_masks` dead in a frozen path | `326a5827` | removed |
| Module docstring narrates R15 vs R16 | `326a5827` | trimmed to the load-bearing constraint |
| `seed` docstring narrates the revision | `326a5827` | trimmed |
| `validate_pair_frame` bypassed | `4397360c` | called before the fixed mask |
| Dollar totals missing from the transcription guard | on branch | both asserted to the cent |

Re-verified in a clean archive of the merged tip: **378 passed, 3 skipped** across the census, placebo, permutation, runner and manifest suites.

One note for the record, since it bears on `confirmatory` later. This merged as a squash, so the four branch commits are one commit on `main`. That does not affect the freeze ordering the R16 design depends on — what matters is that the frozen design bytes are on `main` (`f7cba483…`, pinned in `study.yaml`) before any draw is taken, and no draw has been taken. The #729 hazard was a squash that destroyed the ordering *between* a freeze and a run; here the run is still ahead of us.

